PR #74, opened 2011-04-01
Namespaced bindings

hunterloftis:
After using KO extensively in our latest web app (clickdummy.com) we loved it - it's brilliant! ...but we had some real beef with its complexity limits in these scenarios:
- undefined VMs (single-page app, some data upfront, the rest ajaxed in)
- loose coupling (applyBindings(vm, domElement) means you have to explicitly bind/tightly couple)
- dynamic binding
- forced bad architecture (global variables, monolithic VM objects)

The clear solution to this is namespaced binding:
- bind after VM is defined, only for that VM
- no more need to scope to a particular DOM container / can loosely couple
- can bind dynamically
- good architecture (related VMs in closure, bind from closure, no data leakage, good modularity)

So here it is.

There's a complex demo with namespaces (designed for the latest browsers):
https://developer.mozilla.org/en-US/demos/detail/ghostwriter-art-studio/launch

It's backwards-compatible, so it won't break any existing KO implementations. New implementations can deprecate DOM-containers for namespaces (or stick with the old ways if they prefer).

Example:

span data-bind="text: userName"
a data-bind="click: newImage"
a data-bind-eraser="click: activate"
div data-bind-marker="visible: visible"

ko.applyBindings(GlobalViewModel);
ko.applyBindings(marker, 'marker');
ko.applyBindings(eraser, 'eraser');

cheers,
Hunter Loftis (hunterloftis) via Skookum (skookum)
Chris Gomez (evlapix)


brunohstein:
Is there a way to namespace bindings like this in the current version?

We have a large application with several components that auto-initialize. The components are rendered on the server side and we don't know the context in which they will be used. A component can be rendered in a page that don't even use Knockout at all. We are looking for a simple tool to bind data and event handlers to the view and Knockout still looks like a solid tool for that, but as we can have components inside other components, we have problems regarding the context of the binding instructions.

If we could namespace the `data-bind` (e.g. `data-image-field-bind`) we would be able to keep the components completely decoupled from each other. I'm not familiar with the Knockout code. Is there a way to write a plugin to the current version to achieve that? Maybe with some directions I would be able to do it.

Issue #93, opened 2011-05-05
Options binding does not support OPTGROUPs


tiberiuana:
There is no support for groups in the _options_ binding.

Group support needs two things:
- a way to specify the group of each option
- optionally, a way to specify the order of displayed groups (and thus the groups to be displayed)

May I suggest an _optionsGroup_ option to the _options_ binding and a new _optionsGroupNames_ binding on SELECTs.

PR #94, opened 2011-05-05
Groups support in the options binding

tiberiuana:
Support for OPTGROUP tags as _optionsGroup_ option of _options_ binding (for issue #93).

I'm trying my hand at adding some functionality to KO, the optgroups seemed like a nice place to start. Any comments/guidance much appreciated.


ReyPena:
I know this issue is close but there is nothing in the documentation, and it doesn't look that it has been add it to the core functionality of knockout.
(I'm trying to build a select box with options and optgroup/options)

DenisPitcher:
If you're using boostrap take a look at bootstrapselect, I posted some code a while back that I used for optgroup support with the bootstrap selectpicker
https://github.com/silviomoreto/bootstrap-select/issues/200

ReyPena:
@DenisPitcher Thank you, but I'm not using any css lib, and this seen to be a recurrent issue for the users of knockout, I decide to post here after a lot of research, the closes, easy and dirty way to do this without creating a custom binding is doing something like:

`
<select class="form-control">
	<option data-bind="text: fieldCaption, value: ''">&nbsp;</option>
	
		
			<optgroup data-bind="attr:{label: display}">
				
					<option data-bind="text: display,	value: name"></option>
				
			</optgroup>
		
		
			<option data-bind="text: display,	value: name"></option>
		
	
</select>
`

And this solution slow the rendering of the DOM, I think knockout should give default support for this kind of cases, maybe the solution is add this to the core functionality.

PR #339, opened 2012-02-23
afterAllRender callback for after a foreach binding has add all elements to the DOM

ckniffen:
Added a optional callback to foreach bindings that is executing after the binding has been fully iterated over.

The benefit of having a callback after the observable array has been iterated over is for dom mapipulations like those used in jQuery mobile to progressively enhance the just added page elements. This way it is happening all at one.

In the actual project I am using knockout on the difference between initializing the DOM elements with jQuery mobile in afterAllRender was around 3x faster.

I haven't yet decided what parameters should be passed through and was hoping for some input. I was thinking of passing in **targetNode** and possibly **filteredArray** found in **ko.renderTemplateForEach**

Demo: http://jsfiddle.net/ckniffen/n2ZTU/2/


zachpainter77:
I believe pull requests https://github.com/knockout/knockout/pull/1803 and https://github.com/knockout/knockout/pull/1802 solve this issue and doesn't appear to have merge conflicts.

zachpainter77:
@ckniffen --> Here is an update to your original fiddle that uses my version of knockout.debug.js in external resources and executes your function...

[Updated @ckniffen  JSFiddle](http://jsfiddle.net/zachpainter77/n2ZTU/55/)

ckniffen:
Closed because I no longer use knockout and this is a super old PR.

PR #359, opened 2012-03-05
allow long subscriber chains

haberman:
This allows subscriber chains to be very long
without stack overflow.  This is only done for
the default event; could probably be done for
"beforeChange" also, but would be a more
intrusive code change.

This change passes all tests in my browser
(Chrome).

See: https://github.com/SteveSanderson/knockout/issues/358

### DIFF
--- a/spec/dependentObservableBehaviors.js
+++ b/spec/dependentObservableBehaviors.js
@@ -13,7 +13,7 @@ describe('Dependent Observable', {
     'Should advertise that instances are computed': function () {
         var instance = new ko.dependentObservable(function () { });
         value_of(ko.isComputed(instance)).should_be(true);
-    },    
+    },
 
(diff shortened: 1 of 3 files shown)


brianmhunt:
@haberman – If you have a look at the comments linked to this issue, you'll note that the plan is, at the moment, to have this problem persist on legacy Firefox, for two reasons:

1. One can work around this with deferred updates, which is available as a plugin for KO 3.3 and before, and integrated as of (forthcoming) 3.4
2. The problem only exhibits on old versions of Firefox (versions 10 to 23)

The patch you proposed certainly looks to work, but it adds some complexity and does not extend to other event types that may also be affected.

On that basis I'll close this issue, but please comment if you think we're missing something important here and we can certainly discuss/re-open.

Sorry it took so long to get to this @haberman . I hope you found a workaround in the mean time! 

Cheers

PR #387, opened 2012-03-15
bug in version 2.1.0pre Feb 24, 2012

mbest:
I tried upgrading to latest 2.1.0pre from 2.0.0 in my application and on a page with a lot of complex functionality, nested templates, bindings, etc, in certain cases depending on the data I get error:

Uncaught TypeError: Cannot call method 'dispose' of undefined

2.0.0 works reliably as expected.

My specific case is very complicated to reproduce so I am hoping that it is still helpful just to let you know about this, in case you already have an idea what this might be?

Thanks!

### DIFF
--- a/src/subscribables/dependentObservable.js
+++ b/src/subscribables/dependentObservable.js
@@ -59,8 +59,13 @@ ko.dependentObservable = function (evaluatorFunctionOrOptions, evaluatorFunction
     }
 
     function evaluateImmediate() {
-        if (_isBeingEvaluated)
+        if (_isBeingEvaluated) {
+            // If the evaluation of a ko.computed causes side effects, it's possible that it will trigger its own re-evaluation.
+            // This is not desirable (it's hard for a developer to realise a chain of dependencies might cause this, and they almost
+            // certainly didn't intend infinite re-evaluations). So, for predictability, we simply prevent ko.computeds from causing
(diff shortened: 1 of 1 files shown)


fastfasterfastest:
Sorry to dig up an old issue.

> "A computed re-evaluates when any of its dependencies change. If dependencies change further still during evaluation, this does not restart evaluation. The next evaluation happens only after the previous evaluation has finished, in response to a subsequent dependency change."

Since then, knockout has added `ko.options.deferUpdates`.

In knockout 3.4.2, if a dependency of a computed change during the evaluation of the computed, then an evaluation of the computed is not restarted - that is true, an evaluation is not _immediately_ restarted.  But if `deferUpdates==true` then a re-evaluation of the computed _is_ scheduled and _will_ happen, and in effect the evaluation of the computed is restarted.  I don't know if that is the intent, or whether it is desirable or possible to prevent, that such a schedule of re-evaluation to happen in the case of `deferUpdates`, or not.
It does cause some difference in behavior whether `deferUpdates` is `true` or `false`.

http://knockoutjs.com/documentation/computed-dependency-tracking.html says:
> Knockout will not restart evaluation of a computed while it is already evaluating

True - but knockout does "restart" the evaluation of the computed "later" if `ko.options.deferUpdates==true`

equisgroup:
At least for prevent you can use inside the computed ko.ignoreDependencies(function(args){})

PR #582, opened 2012-07-25
Add ko.asyncComputed shorthand for a minimally-throttled computed

SteveSanderson:
I've been making lots of use of `.extend({ throttle: 1 })` recently to debounce recomputations during batch operations. Speaking to others at the "Throne of JS" conference last weekend, this seems to be a more common requirement than I realised before. One or two people knew how to use `throttle`, but most said they had never heard of it.

It would be nice to offer a really straightforward way of discovering this functionality we already have. By adding `ko.asyncComputed`, and by documenting it, hopefully people will be more confident in making use of this ability.

### DIFF
--- a/spec/asyncBehaviors.js
+++ b/spec/asyncBehaviors.js
@@ -108,10 +108,10 @@ asyncTest("Should run evaluator only once when dependencies stop updating for th
 	}, 10);
 });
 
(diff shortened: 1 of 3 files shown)


SteveSanderson:
This is redundant now we have deferred in core. Deferred is much better than this ever would have been :)

PR #690, opened 2012-10-31
Add global before/after callbacks to foreach/template bindings

seanami:
This is a feature request branch to add two additional callbacks to the ones that the `foreach`/`template` bindings already support (`beforeRemove`, `afterAdd`, `beforeMove`, `afterMove`). These callbacks are a bit different in that they are "global": called once each for each set of changes to the items within the binding instead of once for each different type of change (removes, adds, moves).

I've named the two new callbacks `before` and `after`, but I'm open to naming suggestions. The `before` callback fires once for each set of removes/adds/moves before any changes to the DOM are made and before any other callbacks have fired. The `after` callback is symmetrical and fires once for each set of removes/adds/moves after all changes to the DOM are made and after all other callbacks have fired. Neither callback receives any arguments, although it might be useful if they received arrays of added, removed, and moved nodes. I wanted to start simple and see what people thought.

These global callbacks give you a single opportunity to execute code before and after the DOM is modified during a mapping of an array to DOM node children. This allows you to do things like animate the transition between old and new states in a way that requires knowing the old and new location of child elements. With the current callbacks, the only callback that fires before the DOM is updated is beforeMove, so if there aren't any moves there's no way to execute code to check on old positions before the DOM is changed. Additionally, it's inefficient to execute some code after all changes to the DOM are finished today, as you have to set up some sort of async timeout that is continually cancelled and renewed during each callback for an individual element and which finally triggers once all have fired.

These callbacks are important to efficiently support the animations that we use with Knockout at Typekit, and any animations where you need to know things about position or other aspects of the DOM before and after a change. If you [check out our font browsing page](https://typekit.com/fonts), which is implemented in Knockout, the font cards are drawn with a `template` binding. When you click on a classification or other filter on the right, we animate the position of the cards to visually communicate the idea that you're narrowing down a set of items. In order to pull this off, we look at the status (removed/added/moved) and position of elements before and after the change and then set up animations between the two states.

The changes in the last few versions to the Knockout internals of foreach/template have invalidated our current approach. They're _close_ to making a much simpler and more robust approach possible, but I need callbacks that will reliably allow me to inspect the DOM before and after changes.

### DIFF
--- a/spec/defaultBindings/foreachBehaviors.js
+++ b/spec/defaultBindings/foreachBehaviors.js
@@ -175,6 +175,54 @@ describe('Binding: Foreach', {
         value_of(testNode.childNodes[0]).should_contain_html('<span data-bind="text: childprop">first child</span><span data-bind="text: childprop">added child</span>');
     },
 
+    'Should be able to supply before and after global callbacks': function() {
+        testNode.innerHTML = "<div data-bind='foreach: { data: someItems, before: myBefore, after: myAfter }'><span data-bind='text: childprop'></span></div>";
+        var someItems = ko.observableArray([{ childprop: 'first child' }]);
+        var beforeCallbacks = 0, afterCallbacks = 0;
+        ko.applyBindings({
+            someItems: someItems,
+            myBefore: function() { beforeCallbacks++ },
+            myAfter: function() { afterCallbacks++ }
+        }, testNode);
+
+        value_of(testNode.childNodes[0]).should_contain_html('<span data-bind="text: childprop">first child</span>');
+
+        // Try adding
+        someItems.push({ childprop: 'added child'});
+        value_of(testNode.childNodes[0]).should_contain_html('<span data-bind="text: childprop">first child</span><span data-bind="text: childprop">added child</span>');
+        value_of(beforeCallbacks).should_be(1);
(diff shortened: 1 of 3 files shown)


zachpainter77:
Is there any reason why you do not execute your callback on the first execution?  I created a very similar pull request without seeing this first that is almost the same except I did not test for editScript.length or isFirstExecution.  I use these callbacks with datatables.  In order to get datatables to play nice with knockout you have to destroy the jquery datatable before knockout renders/removes/adds/moves items.  then let knockout render the elements in the original way and then last initialize jquery datatables.  I tested this pull request and it doesn't initialize my jquery knockout bound datatable until I add/remove/move an item in the array.

mbest:
This pull request is obviously quite old (three years) and the linked site (https://typekit.com/fonts) is still using Knockout 2.3.0. It appears from this code and looking at the site that there was a need for different code to run initially vs. on subsequent updates. Since they were also using a wrapper binding, it was easy to have code that ran initially. From looking at the code on the site, it appears that they did not use the method in this pull request to deal with subsequent updates, but instead had their wrapper binding also "unwrap" the array so that it could update the binding directly (and bypass the internal computed that `foreach` uses).

PR #694, opened 2012-11-02
Fix javaScriptAssignmentTarget regular expression

anru:
old re fail to parse expressions like this: VG[foo['param']]


SteveSanderson:
I'll close this since the code in question has changed various times in the intervening years :) If anyone still thinks this remains applicable, please reopen!

PR #701, opened 2012-11-05
Issue with asynchronous load of options and a caption

IanYates83:
I have an issue in knockout where the "value" binding for a select is ready before the "options", so knockout sets the value to the caption's _undefined_.
In the production system I'm working on, we've created a custom "nonNullableObservable" that prevents writing _null_ or _undefined_ to the observable, in order to prevent this from happening, but the ideal would be to use an official solution as part of the framework.
The issue is easily reproduced: http://jsfiddle.net/IanYates83/QQNG8/
(Due to work limitations, I have only currently tested this in IE8)

I tried to keep the fix to as few changes as possible, however there may be some unintended consequences that I'm not aware of. We use KO pretty extensively, and everything else is working for us.
Since the issue is the framework taking the caption value as it is the only correct option at the time of binding, I changed "readValue" in selectExtensions to return _undefined_ if the caption is selected and the only possible value - making the assumption that if the model has a value, there must have been some legitimate options.

By changing the caption value to _null_ in order to be able to single out _undefined_ as an unprocessed value, potentially this could be a breaking change, although _undefined_ shouldn't come from any UI element as a legitimate user input.

The same example using my modified code: http://jsfiddle.net/IanYates83/QQNG8/2/

Apologies for not doing a complete build, but I can't access appspot and thus closure-compiler from work.

### DIFF
--- a/build/output/knockout-latest.debug.js
+++ b/build/output/knockout-latest.debug.js
@@ -1450,7 +1450,10 @@ ko.exportSymbol('toJSON', ko.toJSON);
                         ? (element.getAttributeNode('value').specified ? element.value : element.text)
                         : element.value;
                 case 'select':
-                    return element.selectedIndex >= 0 ? ko.selectExtensions.readValue(element.options[element.selectedIndex]) : undefined;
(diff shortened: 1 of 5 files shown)


SteveSanderson:
Hopefully this is not still an active issue for you! Closing as it appears the behaviour is as expected.

PR #720, opened 2012-11-16
Allow 'options' binding on <datalist> element

LaurentVB:
HTML5 _datalist_ contains _option_ elements. Like _select_, it should accept the _options_ binding. This would allow to do:

``` html
<datalist data-bind="options: list"></datalist>
```

instead of:

``` html
<datalist data-bind="foreach: list">
    <option data-bind="text: $data"></option>
</datalist>
```


castaway:
Just discovered I'd like this too.. any movement on it?

bilal-infotrack:
bump, would be great to have this feature

PR #758, opened 2012-12-13
Deferred templates from a URL source and text interpolation for text binding

brianwigfield:


SteveSanderson:
Wow, this was a very prescient pull request :)

We did end up focusing on both of these bits of functionality. The async external templates thing is now covered by components (KO 3.2.0+), and string interpolation is an ongoing possibility, though we've held back so far because of determination to keep perf very high.

I'll close this since it's not quite applicable any more, but thanks again for suggesting it.

Issue #765, opened 2012-12-14
Expose registerDependency method as observable's member


AbraaoAlves:
As valueHasMutated and valueWillMutate the registerDependency could be exposed as a member of the observable. Providing a simple and customized way to treat the dependancy of record.

eg

``` javascript
var obj = {
    get name() {
        this.__name.registerDependency();
        return this._name;
    },
    set name(value) {
        this.__name.valueWillMutate();
        this._name = value;
        this.__name.valueHasMutated();
    },
    __name    : new ko.observable()
};
```

Look testing and implementation on my [fork](https://github.com/AbraaoAlves/knockout)

What do you think of putting it in the knockout ?
I am a beginner, how to insert it in knockout ?

PR #768, opened 2012-12-19
Expose registerDependency method as observable's member

AbraaoAlves:
Issue #765


SteveSanderson:
Thanks for posting, and sorry I never responded.

After all this time I think it's unlikely we'd include such a feature in KO as it's somewhat outside the pattern of dependency detection that KO uses, plus other people haven't asked for it in the meantime.

Please reopen if you do still feel this is something we should do!

PR #800, opened 2013-01-16
Small performance mainly to degrease statements causing IE8 slow script problem

delixfe:
We have a project for a customer where IE8 will still run for some month. We have a rather complex view model and run into issues with IE8's slow script problem. Analysing the issue with @LuisCantero, we identified that ko.utils.arrayIndexOf is called very often and therefore optimized it a bit.
Working on that I also optimized for loops to cache the length of the array if that made sense.


SteveSanderson:
Since #824 was merged, can we close this? Thanks for the original suggestion!

PR #806, opened 2013-01-21
Exported expressionRewriting.writeValueToProperty so that it can be used...

AndersMalmgren:
... for custom bindings


YipYipX4:
+1

PR #810, opened 2013-01-23
Issue when binding 'type' on IE8

dgaviola:
When binding the 'type' attribute on IE8, I get the error "This command is not supported" when executing code from the 'attr' binding:

    element.setAttribute(attrName, attrValue.toString());

Applying this change fix that problem and it keeps working in other browsers (tested Chrome and Firefox).

Let me know if this makes sense. Thanks!


SteveSanderson:
Based on @mbest's response, I'll close this now, but if anyone still thinks there's an outstanding KO issue with IE8 here, please reopen and I'll look at it with higher priority.

Issue #849, opened 2013-02-15
Disallow Binding/Writing to file inputs


micahasmith:
If you bind a value to a input[type="file"] it throws INVALID_STATE_ERR: DOM Exception 11 as explained here:

http://stackoverflow.com/questions/3488698/invalid-state-err-dom-exception-11-webkit/14901729#14901729

PR #851, opened 2013-02-16
Fix For Input[type="file"] Value Binding

micahasmith:
Based upon https://github.com/SteveSanderson/knockout/issues/849

I ran the build.sh on my windows comp and everything succeeded. 

This is my first pull request-- let me know if there is anything else i need to do.


mbest:
Fixed. See #849.

PR #889, opened 2013-03-16
Fixed #888: Problems when object has own 'hasOwnProperty' property

### LINKED ISSUE #888: Problems with objects containing "hasOwnProperty" property
In the contrived situation where an object actually contains a property named `hasOwnProperty`, some parts of Knockout would fail, namely the `ko.utils.objectForEach`.

This example would fail if it were running against the latest knockout code from master:
http://jsfiddle.net/cspotcode/ePZkk/

The fix is to rewrite `obj.hasOwnProperty(name)` into `Object.prototype.hasOwnProperty.call(obj, name)`.

cspotcode:
Fixes #888.


mbest:
This has already been fixed. See 1523efbde42330

PR #957, opened 2013-05-07
observableArray.remove and destroy wipes the whole array if called with ...

kicsiede:
...observable value


mbest:
This was already fixed with #482.

PR #968, opened 2013-05-13
hidden binding

adoconnection:
as with enable disable


cervengoc:
@mbest @brianmhunt What about this old PR? I like the idea, and KO already has the `enable` `disable` pair, so it would fit well to have `visible` and `hidden`. Shall I create a new PR?

mbest:
@cervengoc Sure that's a good idea.

mbest:
Completed with #2103

Issue #938, opened 2013-04-26
KO has "special AMD handling, and I'm not sure that's playing well"...


thinkloop:
Hello, I asked a question on stack regarding loading the KnockoutFire plugin using RequireJS and it was mentioned that KO has "special AMD handling, and I'm not sure that's playing well with the things I've tried (using shim and exports)": http://stackoverflow.com/questions/16227151/how-to-load-a-js-lib-that-defines-a-global-and-depends-on-another-3rd-party-lib

What is the cleanest way to load KO and KnockoutFire using RequireJS? Are there simple modifications KnockoutFire could do to make it easier for everyone?

AlexGodofsky:
Just spent a couple hours figuring out that this was my problem, in exactly the same situation as @slaneyrw.

PR #1005, opened 2013-06-16
Fix for #938: always export ko as global

dotnetwise:
As per #938 and many other reported bugs on 3rd party libraries i.e. https://github.com/Knockout-Contrib/Knockout-Validation/issues/259

Because some of the third party libraries such as knockout validation are not aware of `AMD` we need to always export `ko` to the global namespace.
Otherwise when optimizing multiple modules with AMD including knockout itself, the other 3rd party modules would simply fail because they would always rely on the global `ko` variable.
This pull request easily fixes that case, and is also making sure the define is not anonymous, so that the shim configuration can be used side-by-side with the default ko module.

### DIFF
--- a/build/fragments/amd-pre.js
+++ b/build/fragments/amd-pre.js
@@ -6,9 +6,10 @@
         factory(target);
     } else if (typeof define === 'function' && define['amd']) {
         // [2] AMD anonymous module
-        define(['exports'], factory);
-    } else {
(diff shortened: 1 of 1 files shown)


SteveSanderson:
Does anyone still have an opinion on this, or should we close it? I'm open to doing this if people still feel there's a need.

PR #1033, opened 2013-07-12
New toJS mapping handler to generate JS along with a model prototype

imrefazekas:
Dear Knockout team,

I added a new function to the mapping handler: toJSByPrototype

Using this function, one can generate the JS by a prototype. So walking through recursively the prototype, every matching observables function will be taken into account.

Helps me a lot in my projects where models are coming from the server and the viewmodel is created according to that and when sending back, the JS has to follow and must be restricted to that original model.

I hope you consider it useful enough to merge to.

### DIFF
--- a/src/subscribables/mappingHelpers.js
+++ b/src/subscribables/mappingHelpers.js
@@ -2,6 +2,54 @@
 (function() {
     var maxNestedObservableDepth = 10; // Escape the (unlikely) pathalogical case where an observable's current value is itself (or similar reference cycle)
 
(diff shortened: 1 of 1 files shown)


SteveSanderson:
Thanks very much for suggesting this!

I hope you don't mind, but I'll close this issue because it was so long ago that I doubt you're still waiting for this to be merged, and nobody else has asked for similar functionality in the meantime so it's probably not a common enough requirement to be baked into the core library. Please reopen if you disagree!

PR #1066, opened 2013-08-07
Support namespaced attributes with default attr binding

rampion:
When using knockout to create manage SVG or XML nodes, it is sometimes desirable to have it manage attributes from specific namespaces.  For example, links in SVG documents use the `xlink` namespace. Currently, the `attr` binding doesn't support attribute namespaces, as can be seen in [this fiddle](http://jsfiddle.net/ZghP7/).  The hardcoded `xlink:href` works as a link, while the `xlink:href` attribute created by knockout doesn't, because it doesn't have the right namespace.

This could certainly be compensated for by creating a `attr-ns` binding just to handle namespaced attributes, however not only would this have a high degree of code repetition, but this breaks the similarity between hardcoded attribute and attributes defined by `attr`. The HTML parser is smart enough to figure out which attributes have namespaces, why shouldn't knockout do the same? The DOM provides the `Node.lookupNamespaceURI()`, `Node.setAttributeNS()` and `Node.removeAttributeNS()` methods just for this purpose.

This pull request uses these API functions to add namespace support to the default `attr` binding. It also adds a test to the spec to check that namespace support is working properly.

ref: http://stackoverflow.com/questions/18083723/how-can-i-get-knockout-js-to-set-the-namespaceuri-for-attributes

### DIFF
--- a/spec/defaultBindings/attrBehaviors.js
+++ b/spec/defaultBindings/attrBehaviors.js
@@ -9,6 +9,24 @@ describe('Binding: Attr', function() {
         expect(testNode.childNodes[0].getAttribute("second-attribute")).toEqual("true");
     });
 
+		it('Should be able to set namespaced attribute values', function() {
+				var model = { myValue: "first value" };
+				testNode.innerHTML = [
+					'<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">',
+						'<g>',
+							'<a data-bind="attr: { \'xlink:href\': myValue }">',
+								'<text>foo</text>',
+							'</a>',
+						'</g>',
+					'</svg>'
+				].join('');
+
+        ko.applyBindings(model, testNode);
+				var anchor = testNode.childNodes[0]/*svg*/.childNodes[0]/*g*/.childNodes[0]/*a*/;
+				expect( anchor.getAttributeNode('xlink:href').value ).toEqual( 'first value' );
+				expect( anchor.getAttributeNode('xlink:href').namespaceURI ).toEqual( 'http://www.w3.org/1999/xlink' );
+		});
+
     it('Should be able to set \"name\" attribute, even on IE6-7', function() {
         var myValue = ko.observable("myName");
         testNode.innerHTML = "<input data-bind='attr: { name: myValue }' />";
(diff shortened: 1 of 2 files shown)


troyschneringer:
+1

krnlde:
It's been two, almost three years, since that feature was requested. What is happening? Why is this PR not merged yet? Big bump!

brianmhunt:
Submit a PR and I'll merge this against tko.binding.core/src/attr.js

rampion:
@brianmhunt I would love to, but I can't seem to run tko.binding.core's tests right now:

```
root@5b0632d0c3f6:~/tko.binding.core# gulp test --once
[00:53:48] Using gulpfile ~/tko.binding.core/gulpfile.js
[00:53:48] Starting 'test'...
08 05 2017 00:53:48.215:WARN [watcher]: Pattern "/root/tko.binding.core/spec/helpers/setup.js" does not match any file
.
08 05 2017 00:54:18.191:INFO [karma]: Karma v1.7.0 server started at http://0.0.0.0:7777/
08 05 2017 00:54:18.194:INFO [launcher]: Launching browser PhantomJS with unlimited concurrency
08 05 2017 00:54:18.227:INFO [launcher]: Starting browser PhantomJS
08 05 2017 00:54:18.690:INFO [PhantomJS 2.1.1 (Linux 0.0.0)]: Connected on socket IY4nCyPIs2VTHKBrAAAA with id 7442831
4
PhantomJS 2.1.1 (Linux 0.0.0)  🚨  Error: SyntaxError: Unexpected token 'const'
at http://localhost:7777/base/spec/attrBehaviors.js?48d2f00b2ac970646673804383217f610e500667:3919
PhantomJS 2.1.1 (Linux 0.0.0) ERROR
  SyntaxError: Unexpected token 'const'
  at spec/attrBehaviors.js:3919


 🏁  Run complete.
...
```

I'm not sure what your in the configuration chain is causing `const` to be emitted in the code, but AFAIK, PhantomJS doesn't support ES6, so this is breaking my ability to run tests.

Tell me how I can test my PR, and I'll gladly submit it.

brianmhunt:
Thanks @rampion – I just broke the testing – we're moving to a monorepo (at knockou/tko) so there are a few hitches, like this, as I nail tooling down.

You can either checkout from a few days ago (where testing should work) or give me a couple days and I'll have it working – it'll be in the monorepo at github.com:knockout/tko's packages/tko.binding.core directory.

Please ping me if I haven't gotten back by next week!  Thanks @rampion

krnlde:
Can I help somehow?

mbest:
Turns out some browsers (Chrome) don't like it when you use `setAttributeNS(null,..` instead of `setAttribute`.

PR #1073, opened 2013-08-15
Performance boost for templatized foreach

evanworley:
I've been running into a number of performance issues with using native templates and foreach, with a large number of items (1000s).

With this change, parsed templates are cached for 5s and the nodes are cloned instead of re-parsed. With my testbed, this gave a 17% speedup (5200ms instead of 6300) ms to render 5000 items.

I'm hoping to uncover even bigger wins, but this is some low hanging fruit for now. Please let me know if there is anything else you'd like me to do with this commit.


mbest:
Superseded by #1201.

PR #695, opened 2013-08-21
Add support of filters/projections for observableArrays

SteveSanderson:
A common scenario where experience from usage of KnockoutJS can be improved is having an observableArray of elements that is not binded directly to HTML, but first projected to a set different items, or filtered by a condition.

Would be great to have something similar to:

var array = ko.observableArray([]);
var filtered = array.filter(function(item) { /\* condition _/ });
var projection = array.select(function(item) { /_ selector */ });

Right now you can achieve this by using a computed observable, but this will mean that every item of resulting array will be tested against condition/selector if any item of initial collection changes, and thus whole array will be rebound, that is not effective in terms of performance.
The fact of recreation is itself crucial as it may result in the loss of state for items that were recreated, in the case it was a projection.

### DIFF
--- a/spec/observableArrayChangeTrackingBehaviors.js
+++ b/spec/observableArrayChangeTrackingBehaviors.js
@@ -207,6 +207,22 @@ describe('Observable Array change tracking', function() {
             { status : 'deleted', value : 'Gamma', index : 2 },
             { status : 'deleted', value : 'Delta', index : 3 }
         ]);
+
+        // Check that extending the observable again doesn't break anything an only one diff is generated
+        var changelist2, callCount = 0;
+        myArray = myArray.extend({trackArrayChanges:true});
+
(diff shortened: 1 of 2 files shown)


MrAndMrsK:
@coryandrew1988 - you mentioned a `mapBind(mapFn, inverseMapFn)` that you were able to build off this.  Is this code available anywhere?  Seems like a useful addition to knockout-projections.

(Overall, I am hoping that this will be useful for `<select multiple>` binding scenarios against many-to-many data models.)

coryandrew1988:
@MrAndMrsK Sorry for the late response. [The code (written years ago, so it might need some fixing) can be found here.](https://gist.github.com/coryandrew1988/738343529502a5a11d8b)

JasonKleban:
@coryandrew1988 Thanks!  (from MrAndMrsK)

PR #1054, opened 2013-09-13
Initialization of observableArray with Null in 2.3.0/3.0 is not backward compatible with 2.2.1 and lower

mbest:
I think this change probably falls under the release note "Observable arrays have better handling of unexpected parameters (for example, when initialized with non-array values)", but here's the issue

The old code:
```javascript
ko.observableArray = function (initialValues) {
    if (arguments.length == 0) {
        // Zero-parameter constructor initializes to empty array
        initialValues = [];
    }
 //...code removed
    var result = ko.observable(initialValues);
//...code removed
```
Will allow the observableArray to initialize to Null whereas the new code:

```javascript
ko.observableArray = function (initialValues) {
    initialValues = initialValues || [];
//...code removed
    var result = ko.observable(initialValues);
//...code removed
```

The problem as I see it is this:
```javascript
var x = ko.observableArray(null);
//x == [] 
x(null);
//x == null  
```
Since null is falsy, when x is initialized, it is set to empty array in 2.3/3.0.
If x is set to null after initialization, it returns null until it reset since it is just an observable under the hood.

We implemented a workaround in our application, but we had valid circumstances where the observable array would be initialized with or reset to null and then tested for such. 

I would consider the inconsistency in behavior between the init and setter to be a bug in 2.3/3.0.

### DIFF
--- a/spec/observableArrayBehaviors.js
+++ b/spec/observableArrayBehaviors.js
@@ -1,5 +1,7 @@
 
 describe('Observable Array', function() {
+    var testObservableArray, notifiedValues, beforeNotifiedValues;
+
     beforeEach(function () {
         testObservableArray = new ko.observableArray([1, 2, 3]);
         notifiedValues = [];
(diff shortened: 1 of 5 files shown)


SteveSanderson:
This conversation is quite dated now, and I doubt that many people are still in the position of needing to upgrade pre-3.x apps. Can we close it, or does anyone still feel there's something we should change?

rcarek:
I'm ok with the close. Ultimately, we took a different route to resolution within our app.

brianmhunt:
Should the branch `1054-observableArray-filter-writes` be deleted now?

mbest:
@brianmhunt I'd like to keep it for now.

PR #1152, opened 2013-10-13
Throw errors if virtual tree traversal finds an unmatched ko closing comment

pimterry:
Currently Knockout will throw an error if you use a opening virtual element with no closing element, but will otherwise completely ignore extra closing comments. It's extremely easy to accidentally use too many closing elements or incorrect nesting and end up with all sorts of problems, particularly with 'with' bindings which will then leave you with slightly wrong contexts applied in the wrong places.

This patch fixes that. It hooks into the virtual tree traversal and loudly fails if at any point you attempt to traverse to a tree node that hasn't previously been matched by a starting node. End comments that are matched are tracked by tagging with ko.domData. This gets done when starting nodes are about to be included in the traversal route; they're found via virtualElements.nextSibling, which calls getMatchingEndComment, which calls getVirtualChildren, which _now_ adds a marker to correctly matching end comments as they're found.

One contentious bit that's worth thinking about is the error messages: I did this with outerHTML initially, so it gave you the relevant chunk of the page as part of the error message. That was useful, but a potentially bit long-winded, and doesn't work in FF < 4. I've pulled that out, and it now gives errors along the lines of 'Found end comment without opening comment, as first child of [object HTMLDivElement]'. Gives you the parent element type, and makes the situation fairly clear if you debug into it, but otherwise not too great.

### DIFF
--- a/src/virtualElements.js
+++ b/src/virtualElements.js
@@ -141,8 +141,8 @@
 
         firstChild: function(node) {
             if (!isStartComment(node)) {
-                if (node.firstChild && isUnmatchedEndComment(node.firstChild)) {
-                    throw new Error("Found end comment without opening comment, as first child of " + node);
+                if (node.firstChild && isEndComment(node.firstChild)) {
+                    throw new Error("Found invalid end comment, as the first child of " + node);
                 }
                 return node.firstChild;
             } else if (!node.nextSibling || isEndComment(node.nextSibling)) {
--- a/src/virtualElements.js
+++ b/src/virtualElements.js
@@ -145,6 +145,8 @@
                     throw new Error("Found end comment without opening comment, as first child of " + node);
                 }
                 return node.firstChild;
+            } else if (!node.nextSibling || isEndComment(node.nextSibling)) {
+                return null;
             } else {
                 return node.nextSibling;
             }
--- a/src/virtualElements.js
+++ b/src/virtualElements.js
@@ -24,12 +24,19 @@
         return (node.nodeType == 8) && endCommentRegex.test(commentNodesHaveTextProperty ? node.text : node.nodeValue);
     }
 
+    function isUnmatchedEndComment(node) {
+        return isEndComment(node) && !(ko.utils.domData.get(node, matchedEndCommentDataKey));
+    }
+
+    var matchedEndCommentDataKey = "__ko_matchedEndComment__"
+
     function getVirtualChildren(startComment, allowUnbalanced) {
         var currentNode = startComment;
         var depth = 1;
         var children = [];
         while (currentNode = currentNode.nextSibling) {
             if (isEndComment(currentNode)) {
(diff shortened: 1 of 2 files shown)


pimterry:
Any updates or thoughts on this? I hit this problem recently and remembered this fix, which is (amazingly) still conflict-free, and could be merged any time. If not, any thoughts on my last comment above?

pimterry:
@mbest any enthusiasm for merging this? It's my oldest still-open PR by a long way, but after many years it's still mergeable, and still a nice way to help people avoid an annoying gotcha.

brianmhunt:
This is a good idea.  If a PR were set up, I'd merge this onto tko.

src/virtualElements.js
```diff
@@ -133,19 +140,30 @@
         },
 
         firstChild: function(node) {
-            if (!isStartComment(node))
+            if (!isStartComment(node)) {
+                if (node.firstChild && isUnmatchedEndComment(node.firstChild)) {
+                    throw new Error("Found end comment without opening comment, as first child of " + node);
+                }
                 return node.firstChild;
-            if (!node.nextSibling || isEndComment(node.nextSibling))
```
mbest: I don't think this should be removed. If you have `<!-- ko text: abc --><!-- /ko -->`, and call `firstChild`, it should return `null`, but the new code will return the closing comment.

src/virtualElements.js
```diff
@@ -133,19 +140,30 @@
         },
 
         firstChild: function(node) {
-            if (!isStartComment(node))
+            if (!isStartComment(node)) {
+                if (node.firstChild && isUnmatchedEndComment(node.firstChild)) {
```
mbest: This code assumes that we've already called getVirtualChildren in these node, but that isn't necessarily true.

pimterry: Interesting, that's certainly true. Is it ever actually valid to have a closing comment as the first child of any element though? Getting the first child of the div in `<!-- ko text abc --><div><!-- /ko --></div>` should really throw an error, right? We know at that point, with no wider context, that your structure is definitely invalid (in a normal applyBinding the full example would throw an error anyway, complaining about the start comment being unmatched, but the specific firstChild call case is interesting standalone to).

If we can take as given that _any_ end comment here is invalid, then this could just throw an error like 'Found closing virtual element as first child of [node]', and not worry about the matching here either way.

src/virtualElements.js
pimterry: True. Updated.

src/virtualElements.js
mbest: Right. This could just check `isEndComment(node.firstChild)`

pimterry: Done, and I've added an extra test to try and cover these cases a little better en route too.

mbest:
@pimterry Thanks for being willing to revisit this after so long! It looks good now.

PR #1158, opened 2013-10-16
For value binding, when model rejects change, update view to match

mbest:
The documentation includes an example of a computed observable that will "reject" certain updates: http://knockoutjs.com/documentation/extenders.html#live_example_1_forcing_input_to_be_numeric

This example includes code to update the input when the value is rejected: 

``` javascript
//if the rounded value is the same, but a different value was written, force a notification for the current field
if (newValue !== current) {
    target.notifySubscribers(valueToWrite);
}
```

In Knockout 3.0.0, even this code won't be sufficient because the computed observable will also ignore the change. See 78442f90df5a.

A small change to the value binding that checks if the model property has rejected a change can solve this problem without having to complicate the issue for the user.


mbest:
Related: #1054

PR #1184, opened 2013-10-30
Normalize build filenames

heruan:
Unminified (debug) output file would be more compatible with package managers like RequireJS or Bower if renamed to `knockout.js` without `-debug` or `-latest` suffixes.
Similar for the minified version, would be better to use the de facto standard `.min` suffix.


mbest:
Tagged releases (since 3.1.0) have a `bower.json` and a `dist` folder with normalized names.

PR #1198, opened 2013-11-11
simplify isWritableObservable, hasPrototype

timtucker:
Simplify isWritableObservable by using existing methods for isObservable and isComputed.

Since the only uses of ko.hasPrototype are in ko.isObservable and ko.isComputed and doesn't seem to be used, I went ahead and simplified the null / undefined checks with a check to see if the instance has a typeof "function".  (Arguably the function might be more clearly named "ko.hasFunctionPrototype" with the new behavior)


mbest:
This code is now (since cdd2741d0f1960) pretty simple. I don't think this is necessary.

PR #1201, opened 2013-11-13
Performance improvements for named templates

evanworley:
I originally tried to get this performance boost in https://github.com/knockout/knockout/pull/1073, but it was not accepted. I've now re-done something similar for Knockout 3.0, and I hope it will be accepted. It provides a significant speed up when using named templates.

Extending the native template engine to cache the parsed HTML from named
templates. The cached nodes are stored for 10s, and then purged. This
greatly speeds up cases like templated foreach.

Performance comparison - On my machine there is a 400ms speed up
  Knockout 3.0.0 - http://jsfiddle.net/u6CvU/
  Knockout 3.0.0 with my changes - http://jsfiddle.net/Rmf3z/

### DIFF
--- a/src/templating/native/nativeTemplateEngine.js
+++ b/src/templating/native/nativeTemplateEngine.js
@@ -1,5 +1,8 @@
 ko.nativeTemplateEngine = function () {
     this['allowTemplateRewriting'] = false;
+
+    // Maps a element id to the parsed HTML for that template
+    this.namedTemplateCache = {}
 }
 
 ko.nativeTemplateEngine.prototype = new ko.templateEngine();
@@ -10,7 +13,16 @@ ko.nativeTemplateEngine.prototype['renderTemplateSource'] = function (templateSo
         templateNodes = templateNodesFunc ? templateSource['nodes']() : null;
 
     if (templateNodes) {
-        return ko.utils.makeArray(templateNodes.cloneNode(true).childNodes);
+        var nodes;
+        if (templateNodes instanceof Array) {
+            nodes = [];
+            for (var i = 0; i < templateNodes.length; ++i) {
+                nodes.push(templateNodes[i].cloneNode(true));
+            }
+        } else {
+            nodes = ko.utils.makeArray(templateNodes.cloneNode(true).childNodes);
+        }
+        return nodes;
     } else {
         var templateText = templateSource['text']();
         return ko.utils.parseHtmlFragment(templateText);
--- a/src/templating/templateEngine.js
+++ b/src/templating/templateEngine.js
@@ -37,11 +37,27 @@ ko.templateEngine.prototype['createJavaScriptEvaluatorBlock'] = function (script
 ko.templateEngine.prototype['makeTemplateSource'] = function(template, templateDocument) {
     // Named template
     if (typeof template == "string") {
+        var namedTemplateCache = this.namedTemplateCache || {};
+
+        // See if we've already build this template
+        if (namedTemplateCache[template]) {
+            return namedTemplateCache[template];
+        }
+
         templateDocument = templateDocument || document;
         var elem = templateDocument.getElementById(template);
         if (!elem)
             throw new Error("Cannot find template with ID " + template);
-        return new ko.templateSources.domElement(elem);
+        var domElementSource = new ko.templateSources.domElement(elem);
+
+        namedTemplateCache[template] = domElementSource;
+
+        // Garbage collect this template if not used within 10s (helpful for batch templating, like in a foreach
+        setTimeout(function() {
+            delete namedTemplateCache[template];
+        }, 10000);
+
+        return domElementSource;
     } else if ((template.nodeType == 1) || (template.nodeType == 8)) {
         // Anonymous template
         return new ko.templateSources.anonymousTemplate(template);
--- a/src/templating/templateEngine.js
+++ b/src/templating/templateEngine.js
@@ -39,9 +39,45 @@ ko.templateEngine.prototype['makeTemplateSource'] = function(template, templateD
     if (typeof template == "string") {
         var namedTemplateCache = this.namedTemplateCache || {};
 
+        // Cache entries last for 10s after the last access
+        var cacheLifetime = 10000;
+
+        // Registers a timer to clean up stale entries from the template cache.
+        function cleanupCache() {
+            // If we do not yet have a cleanup scheduled, schedule one
+            if (!this.templateCacheCleanupTimeoutID) {
+                this.templateCacheCleanupTimeoutID = setTimeout(function() {
+                    this.templateCacheCleanupTimeoutID = null;
+
+                    var cacheEmpty = true;
+                    var currentTime = Date.now()
+
+                    // Walk through the templates, and delete all that have expired
+                    for (var templateName in namedTemplateCache) {
+                        var template = namedTemplateCache[templateName];
+                        if (currentTime - template.lastAccessed > cacheLifetime) {
+                            delete namedTemplateCache[templateName];
+                        } else {
+                            cacheEmpty = false;
+                        }
+                    }
+
+                    // If there are remaining cache entries, register a clean up timer
+                    if (!cacheEmpty) {
+                        cleanupCache();
+                    }
+              }, cacheLifetime);
+          }
+        }
+
+        // Garbage collect this template if not used within cacheLifetime (helpful for batch templating, like in a foreach)
+        cleanupCache();
+
         // See if we've already build this template
         if (namedTemplateCache[template]) {
-            return namedTemplateCache[template];
+            var cacheEntry = namedTemplateCache[template];
+            cacheEntry.lastAccessed = Date.now()
+            return cacheEntry.domElementSource;
         }
 
         templateDocument = templateDocument || document;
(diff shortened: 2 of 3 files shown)


src/templating/native/nativeTemplateEngine.js
```diff
@@ -10,7 +13,16 @@ ko.nativeTemplateEngine.prototype['renderTemplateSource'] = function (templateSo
         templateNodes = templateNodesFunc ? templateSource['nodes']() : null;
 
     if (templateNodes) {
-        return ko.utils.makeArray(templateNodes.cloneNode(true).childNodes);
+        var nodes;
+        if (templateNodes instanceof Array) {
+            nodes = [];
+            for (var i = 0; i < templateNodes.length; ++i) {
```
brianmhunt: Why not `ko.utils.arrayForEach`?  If doing the loop here this should cache `templateNodes.length`

src/templating/templateEngine.js
```diff
@@ -37,11 +37,63 @@ ko.templateEngine.prototype['createJavaScriptEvaluatorBlock'] = function (script
...
     if (typeof template == "string") {
+        var namedTemplateCache = this.namedTemplateCache || {};
+
+        // Cache entries last for 10s after the last access
+        var cacheLifetime = 10000;
+
+        // Registers a timer to clean up stale entries from the template cache.
+        function cleanupCache() {
+            // If we do not yet have a cleanup scheduled, schedule one
+            if (!this.templateCacheCleanupTimeoutID) {
```
brianmhunt: `this` == `window`?

src/templating/templateEngine.js
```diff
@@ -37,11 +37,63 @@ ko.templateEngine.prototype['createJavaScriptEvaluatorBlock'] = function (script
...
+                    // If there are remaining cache entries, register a clean up timer
+                    if (!cacheEmpty) {
+                        cleanupCache();
+                    }
+              }, cacheLifetime);
+          }
+        }
+
+        // Garbage collect this template if not used within cacheLifetime (helpful for batch templating, like in a foreach)
+        cleanupCache();
```
brianmhunt: I'd consider naming this to something more descriptive like `startCleanupCacheTimer`

src/templating/templateEngine.js
```diff
@@ -37,11 +37,63 @@ ko.templateEngine.prototype['createJavaScriptEvaluatorBlock'] = function (script
 ko.templateEngine.prototype['makeTemplateSource'] = function(template, templateDocument) {
     // Named template
     if (typeof template == "string") {
+        var namedTemplateCache = this.namedTemplateCache || {};
+
+        // Cache entries last for 10s after the last access
+        var cacheLifetime = 10000;
+
+        // Registers a timer to clean up stale entries from the template cache.
+        function cleanupCache() {
```
brianmhunt: I may have missed it, but there doesn't appear to be a reason this is a closure - could it go outside the `makeTemplateSource` as a standalone function?

src/templating/native/nativeTemplateEngine.js
```diff
@@ -10,7 +13,16 @@ ko.nativeTemplateEngine.prototype['renderTemplateSource'] = function (templateSo
         templateNodes = templateNodesFunc ? templateSource['nodes']() : null;
 
     if (templateNodes) {
```
brianmhunt: Could this be `if (templateNodes && templateNodes instanceof Array)` or just `if (templateNodes instanceof Array)`?

brianmhunt:
I think we'd want to have some tests that the cache adds and removes as expected (probably in [spec/nativeTemplateEngineBehaviors.js](https://github.com/knockout/knockout/blob/master/spec/nativeTemplateEngineBehaviors.js)).

src/templating/native/nativeTemplateEngine.js
evanworley: Not unless we refactored some other code. Notice the Array check inside and handling if it's not an array. The only way that change would make sense would be if we inserted an `else if (templateNodes)` before the `else`.

src/templating/native/nativeTemplateEngine.js
evanworley: No reason why I used a native for loop vs arrayForEach. Also I am not sure caching the array length really makes much of a difference these days. Even in older browsers it doesn't make much difference with the amount of items we have here. Try http://jsperf.com/cached-array-length and see what you get. My Chrome/Mac was actually marginally faster without caching the array length.

src/templating/templateEngine.js
evanworley: `this` is the templateEngine that extended this prototype. So each template engine will have it's own cache cleanup timeout. I'm open to ideas, how should I make this more clear?

src/templating/templateEngine.js
evanworley: No good reason, I guess we could have it outside and then pass the `namedTemplateCache` as an argument?

src/templating/templateEngine.js
evanworley: Good idea, that would be more clear

evanworley:
I agree, thanks for the suggestion. I'll look into this.

src/templating/templateEngine.js
stalniy: No, `this` there is `window`. `cleanupCache` function is called without context, so by default it's `window`. Also it looks like you can move this function out of `mateTemplateSource` method

src/templating/templateEngine.js
stalniy: or maybe `scheduleCacheCleanUp`?

src/templating/native/nativeTemplateEngine.js
stalniy: There should be used `ko.utils.arrayMap` instead of regular `forEach`

src/templating/templateEngine.js
```diff
@@ -37,11 +37,63 @@ ko.templateEngine.prototype['createJavaScriptEvaluatorBlock'] = function (script
...
+                this.templateCacheCleanupTimeoutID = setTimeout(function() {
+                    this.templateCacheCleanupTimeoutID = null;
+
+                    var cacheEmpty = true;
+                    var currentTime = new Date().getTime();
+
+                    // Walk through the templates, and delete all that have expired
+                    for (var templateName in namedTemplateCache) {
+                        if (namedTemplateCache.hasOwnProperty(templateName)) {
+                            var template = namedTemplateCache[templateName];
```
stalniy: No need it `template` variable as you use it only once

stalniy:
Looks like all this optimization is applicable only to `domElement` templateSource, can all this code be moved under its responsibility?

src/templating/templateEngine.js
```diff
@@ -37,11 +37,63 @@ ko.templateEngine.prototype['createJavaScriptEvaluatorBlock'] = function (script
...
+                            if (currentTime - template.lastAccessed > cacheLifetime) {
+                                delete namedTemplateCache[templateName];
+                            } else {
+                                cacheEmpty = false;
+                            }
+                        }
+                    }
+
+                    // If there are remaining cache entries, register a clean up timer
+                    if (!cacheEmpty) {
```
stalniy: As I understood every time `mateTemplateSource` is called cache cleanUp process is scheduled. Why do you schedule once more cleanUp process for entries which were checked for validity just a moment ago? (in this case you have 100500 cleanUp processes)

brianmhunt:
I have started an experiment at [brianmhunt/knockout-fast-foreach](/brianmhunt/knockout-fast-foreach) that uses `trackArrayChanges` and `requestAnimationFrame` to create an alternative but faster *foreach* binding.

Preliminary tests indicate an improvement of around an order of magnitude for adding to lists with 500 items, two orders of magnitude for deleting from similarly sized lists, with exponential gains for larger lists. If you check it out, you can tinker with it yourself using `npm install; gulp`.

I would be grateful for any thoughts, feedback and assistance to make this a viable drop-in replacement for the standard `foreach` binding.

It needs some re-assurance in the design department, unit and off-the-cuff testing, documentation, and to be set up for distribution to the package repos.

4ver:
Sounds great @brianmhunt ! Will try it out next week cc @adam-lynch @irlTopper

mbest:
I haven't looked at the code here yet, but just looking at the discussion, I'm wondering why we shouldn't just cache the template nodes indefinitely since that's what is done for inline templates. Any thoughts?

Cheema1:
It should be cache the nodes.

evanworley:
@mbest I was hoping to clean up the template nodes eventually to reduce memory utilization. We have a large single page app which uses many different named templates.

mbest:
After thinking about this some more, I decided to work on a similar idea that could provide even better benefits. See #1839.

evanworley:
Very cool! +1

mbest:
Once people are using 3.4.0 and can switch their templates over to using `<template>`, perhaps we won't have much of a need for optimizing `<script>` templates.

evanworley:
Hi Michael,

Can you elaborate on the template element vs script templates, and how they
impact performance? I want to make sure I take advantage of the performance
boost in 3.4.0 for named templates.

Thanks,
Evan

mbest:
When using `<template>`, the template contents are already parsed by the browser and Knockout can use the parsed nodes directly as the template without the need to do additional parsing. Of course `<template>` doesn't have 100% support, but Knockout will still be able to use `<template>` even if the browser treats it as a regular element.

mbest:
Replaced by #2170

simeyla:
There's no mention of <template> on the main docs page. Would be great to get it added along with any caveats.

PR #1231, opened 2013-12-01
Detect and reject (probably accidental) empty anonymous templates

pimterry:
It's very easy to accidentally define your template or foreach binding incorrectly and end up building an anonymous template from your empty element, rather than using the named template you intended to (I personally have done `data-bind="template: { templateName: 'xyz' }"` for example quite a few times). It can be non-trivial to debug this (particularly if the template isn't supposed to end up showing anything, to start with), and it'd be much better if Knockout caught it for you.

This patch solves that. As far as I know there's no good reason whatsoever to define an empty anonymous template otherwise; it's a binding that can never do anything, and it seems like a clear place where Knockout could usefully point out that you've probably made a mistake.


pimterry:
Awesome, thanks! 🙌

PR #1405, opened 2014-05-06
Memory leak in ko.computed

FallSe7en:
Hi,

I think ko.computed is holding onto a reference to readFunction even after it gets disposed.

This pastebin demonstrates the leak: http://pastebin.com/Ub0Ef6LY

Let me know if there are any issues!

Thanks,
Justin


froodian:
@mbest I noticed you setting up https://github.com/knockout/knockout/milestones/Post%203.4.0 (congrats on the release btw and thank you, very excited to upgrade to it today) - our knockout has been forked with this patch - maybe this is something that could get pulled into an upcoming release?

froodian:
Or actually, has this been totally obsoleted in the 3.4 release?

mbest:
`ko.computed` has been completely reorganized in 3.4.0. I don't know if this problem is fixed or not.

PR #1415, opened 2014-05-15
Fix $rawData within 'with' bindings and rendered templates

yjo:
Make $rawData correctly refer to the local view model observable inside a 'with' binding.


mbest:
I've merged this with some minor modifications. (6dbbbe2)

PR #1334, opened 2014-06-03
Conflict between afterkeydown and rateLimit

mbest:
If you bind an observable extended with `rateLimit` using `value` with the `afterkeydown` event, you can end up with characters missing from your value. It is quite hard to reproduce because it depends on the timing being exactly right. But if you could slow down the steps and inject the `keydown` event at the right time, you'll reproduce it every time. Here are the steps:
1. Observable is changed to `'abc'`, which schedules notification in _x_ ms.
2. Just before the notification happens, a `keydown` event is triggered, for `'d'`, which schedules an update to the observable in 1 ms.
3. The `'abc'` notification happens and updates the value of the `<input>` to `'abc'`.
4. The observable is updated from the `keydown` event using the current value of the `<input>`, which is `'abc'`. The `'d'` is lost.

This also occurs when using the `throttle` extender instead of `rateLimit`.

### DIFF
--- a/src/binding/defaultBindings/value.js
+++ b/src/binding/defaultBindings/value.js
@@ -43,6 +43,13 @@ ko.bindingHandlers['value'] = {
             var handler = valueUpdateHandler;
             if (ko.utils.stringStartsWith(eventName, "after")) {
                 handler = function() {
+                    // The elementValueBeforeEvent variable is non-null *only* during the brief gap between
+                    // a keyX event firing and the valueUpdateHandler running, which is scheduled to happen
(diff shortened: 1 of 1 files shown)


jonathancounihan:
First off great library, thanks. And I know I am like years late to the party :), but I think a little console log message would have been super helpful....

Something like: 

`'update': function() { 
   if (console && console.warn) { 
   console.warn('This method is no longer supported.') 
}
}`

We just upgraded from 3.1.0 this week (because reasons) and this caught me a bit by surprise....

Thanks for the hard work!

PR #1432, opened 2014-06-11
ApplyBindings now accepts DocumentFragments.

soruban:
Allows situation like the following to work:

```
    var div = document.createElement("div");
    div.innerHTML = "<div data-bind='text: fooText'></div><div data-bind='text: barText'></div>"

    var frag = document.createDocumentFragment();
    frag.appendChild(div);

    var model = {
        fooText: ko.observable("foo"),
        barText: ko.observable("bar")
    }

    ko.applyBindings(model, frag);
    document.body.appendChild(frag);
```

### DIFF
--- a/src/binding/bindingAttributeSyntax.js
+++ b/src/binding/bindingAttributeSyntax.js
@@ -429,7 +429,7 @@
             jQueryInstance = window['jQuery'];
         }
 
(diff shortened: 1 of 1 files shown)


soruban:
outdated

PR #1160, opened 2014-06-13
Add 'input' binding for real-time updates from text fields

mbest:
The `value` binding documentation includes examples of `valueUpdate` options (`keyup`, `keypress`, and `afterkeydown`) and recommends `afterkeydown` as the best choice to keep the view model updated in real-time. But in modern browsers, `input` is actually the best choice since it fires for changes from mouse events (such as from the context menu or drag-and-drop).

I propose that we present the `input` event as the preferred option for real-time updates. In browsers that don't support `input`, we can simulate it using `afterkeydown` and/or `propertychange`.

In the meantime, we should add `input` as a suggested option in the documentation.

### DIFF
--- a/src/binding/defaultBindings/textInput.js
+++ b/src/binding/defaultBindings/textInput.js
@@ -106,10 +106,11 @@ ko.bindingHandlers['textInput'] = {
                 }
             });
         } else {
-            if (ko.utils.ieVersion < 9) {
+            if (ko.utils.ieVersion < 10) {
                 // Internet Explorer <= 8 doesn't support the 'input' event, but does include 'propertychange' that fires whenever
                 // any property of an element changes. Unlike 'input', it also fires if a property is changed from JavaScript code,
-                // but that's an acceptable compromise for this binding.
+                // but that's an acceptable compromise for this binding. IE 9 does support 'input', but since it doesn't fire it
+                // when using autocomplete, we'll use 'propertychange' for it also.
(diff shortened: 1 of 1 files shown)


MaleDong:
Hello:

I tested "input" on IE8,7,6. And I wanna know what process we are now? And you said earlier that we can consider using "onpropertyChanged" event (ONLY FOR IE). So is that OK?

PS：Of course I know that it won't be an easy task to support all kinds of browsers……Hmmmmm……If we can support more, that'd be better;)

mbest:
@MaleDong, Knockout includes the `textInput` binding that does what you're looking for.

MaleDong:
Well……Is that the latest version that has the property?
Any examples on your official website? I didn't find anything about that?

mbest:
http://knockoutjs.com/documentation/textinput-binding.html

alundiak:
There are 2 items I'm wondering:
1) According best practices we better use camelCase whenever we can. So I would say to have only `textInput`
2) From consistency and standard prospective, I would say we have to have standard following.
So following this behavior, we have to have aliases for rest bindingsHandlers `selectedOptions`, `checkedValue`, `uniqueName` ?. 

IMHO I don't need aliases, I would reduce amount code and got rid of aliases features. For me camelCase rule has bigger priority.

@mbest what do you think? Is this aliasing is the final approach that KO will use and follow?

PR #1448, opened 2014-07-03
Update the array utilities to use the native methods

ismell:
By using the native methods this opens up an interesting usecase. I was playing around with the component flow and was having a hard time because my View Models are created by the template engine. What I needed was a way for my Backbone collections and models to get passed in as parameters into my View Models. 

viewmodel.js

```
var viewModel = {collection: ko.observable(new Backbone.Collection())}
```

template.html

```
<!-- ko foreach: collection -->  
  <custom-component params="model: $data"></repository-card>
<!-- /ko -->
```

After some digging around I found that if I just use the _native_ methods in the array helpers everything else just works.

The last step was to hook onto the collections events so I wrote a quick extender that will do just that. https://gist.github.com/ismell/d4f977f2d7a4a86b9cb1

So now if I just extend my observable with trackCollectionChanges my views will correctly update whenever my backbone collection changes.

viewmodel.js

```
var viewModel = {
    collection: ko.observable(new Backbone.Collection()).extend({'trackCollectionChanges': true})
}
```

Sorry for the long story...

### DIFF
--- a/spec/observableArrayChangeTrackingBehaviors.js
+++ b/spec/observableArrayChangeTrackingBehaviors.js
@@ -74,6 +74,32 @@ describe('Observable Array change tracking', function() {
         });
     });
 
+    it('Converts a non array to an array', function() {
+        captureCompareArraysCalls(function(callLog) {
+            var myArray = ko.observable('hello').extend({'trackArrayChanges':true}),
+                changelist1;
+
+            myArray.subscribe(function(changes) { changelist1 = changes; }, null, 'arrayChange');
+            myArray(1);
+
+            // See that, not only did it invoke compareArrays only once, but the
+            // return values from getChanges are the same array instances
+            expect(callLog.length).toBe(1);
+            expect(changelist1).toEqual([
+                { status: 'deleted', value: 'hello', index: 0 },
+                { status: 'added', value: 1, index: 0 }
+            ]);
+
+            // Objects get converted to Array
+            myArray({a: 1});
+            expect(callLog.length).toBe(2);
+            expect(changelist1).toEqual([
+                { status: 'deleted', value: 1, index: 0 },
+                { status: 'added', value: {a: 1}, index: 0 }
+            ]);
+        });
+    });
+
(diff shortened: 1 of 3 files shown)


fastfasterfastest:
This changes the behavior of `ko.utils.arrayFirst` if an element is not found in the array.  Previously, `ko.utils.arrayFirst` returned `null` if an element was not found.  With this change,  `ko.utils.arrayFirst` returns `undefined` if an element is not found.

Is this an intentional change of behavior of `ko.utils.arrayFirst`?

I am running into this because I am doing strict equality check against between the return value and `undefined`

mbest:
I wonder why that wasn't caught by the tests. Probably we should change it back.

fastfasterfastest:
Yes, it was a surprise to me that `ko.utils.arrayFirst` changed behavior.  And maybe it will be to others as well.  That said, it is easy enough for me to adjust my code to use another check as long as it is well defined what the return value will be.

It might be tricky to implement `ko.utils.arrayFirst` using `Array.prototype.find` (assuming `ko.utils.arrayFirst` should return `null` when an element is not found).  In particular if the element that fulfills the predicate has the value `undefined`...

mbest:
I think having `arrayFirst` return `undefined` is preferable to make it match the official `find` function. We'll have to note this change in the release notes, of course.

fastfasterfastest:
There are no bug reports or other issues reported with regards to `ko.utils.arrayFirst`, right?  Yet, the behavior of it is changed and a breaking change is introduced for seemingly no good reason.

An alternative course of action would be to leave `ko.utils.arrayFirst` as-is/was (returning `null` if element is not found) and add `ko.utils.arrayFind`, if deemed needed and necessary, that behaves like `Array.find` (returning `undefined` if element is not found.)

ManuelHaas:
@mbest Can you please list changing the return value of ko.utils.arrayFirst as breaking change in the release history! Thx

mbest:
@ManuelHaas updated https://github.com/knockout/knockout/releases/tag/v3.5.0-rc2

Issue #1307, opened 2014-02-10
Instead of `instanceof` use `toString` to solve cross-frame/context issues


faridnsh:
So I use jsdom to run my binding handlers unit tests. And I realized checked binding is failing to recognize that the value of the binding is an array but it was treating it as a boolean, so checked binding was checking all my checkboxes instead of one. I checked knockout source and found the problem was coming from this line: [`ko.utils.unwrapObservable(valueAccessor()) instanceof Array`](https://github.com/knockout/knockout/blob/31b9df3a05326de1c46e8ac326f02016e6ce2d29/src/binding/defaultBindings/checked.js#L76), 

The problem is that my unit tests is in one context(the main nodejs script context) and knockoutjs source script is evaluated in another context(done by jsdom). And the `Array` constructor is different in both contexts. Same issue happens when if the array came from another frame in the browser.

This problem has been described here: http://perfectionkills.com/instanceof-considered-harmful-or-how-to-write-a-robust-isarray/

To replicate the problem on browser you can try this:

```
var iframe = document.createElement('iframe');
document.body.appendChild(iframe);
xArray = window.frames[window.frames.length-1].Array;
var arr = new xArray(1,2,3); // [1,2,3]

// Boom!
arr instanceof Array; // false

// Boom!
arr.constructor === Array; // false
```

As a fix, we can add a util function called isArray:

```
knokcout.utils.isArray = Array.isArray || function isArray(o) {
  return Object.prototype.toString.call(o) === '[object Array]';
};
```

But arrays are not the only things that are checked with instanceof. I found this other checking if object is primitive or not:
https://github.com/knockout/knockout/blob/c59104cde3eaeb2120203a17a19d2fd620df1055/src/subscribables/mappingHelpers.js#L27

and also somebody asked for this before: https://github.com/knockout/knockout/pull/184#issuecomment-24156563

Note that most of popular utilities libraries(underscore, lodash, sugar, jquery) use this method to check the type.

aufula:
please fix this

PR #1465, opened 2014-07-22
Remove use of instanceof

JamesThorpe:
instanceof does not work correctly for cross frame code.  Removed it's use in mapJsObjectGraph to fix ko.toJS on cross-frame objects.

See here for example of problem caused:
http://stackoverflow.com/questions/24890168/ko-tojs-converting-array-to-object/

And here for a bug report for using instanceof in general:
https://github.com/knockout/knockout/issues/1307

### DIFF
--- a/src/subscribables/mappingHelpers.js
+++ b/src/subscribables/mappingHelpers.js
@@ -28,7 +28,7 @@
         if (!canHaveProperties)
             return rootObject;
 
-        var outputProperties = rootObject instanceof Array ? [] : {};
+        var outputProperties = toString.call(rootObject) === "[object Array]" ? [] : {};
         visitedObjects.save(rootObject, outputProperties);
 
         visitPropertiesOrArrayEntries(rootObject, function(indexer) {


aufula:
+1 to this MR. but it is still not be  merged to master

brianmhunt:
The idea here is right.  The canonical way, following MDN, would probably be a polyfill.

```javascript
if (!Array.isArray) {
  Array.isArray = function(arg) {
    return Object.prototype.toString.call(arg) === '[object Array]';
  };
}
```

JamesThorpe:
I think the reason it didn't get beyond my initial couple of stabs at it is that at the time I didn't have time to get the build process working to ensure any tests I wrote to go with the patch worked correctly.  Interestingly the same issue has cropped up for us again a month or two ago, and after debugging for a while we found that we'd hit the same issue, 2 years later... I think maybe I should be able to justify spending some time on it at some point so that our internal patch can be accepted upstream :)

PR #1511, opened 2014-09-03
Spectator-mode subscriptions OR "side-effect-free" subscriptions

lcorneliussen:
In #855 knockout "fixed" the flaw that a subscribe on a deferredComputed wouldn't lead to it's computation.

However, in some cases it is still desired to subscribe to a deferred or pure computed without beeing counted as an active subscriber.

This could be solved by adding an option to subscriptions, that could be called `spectatorMode` (`true` /`false`). Subscriptions in spectatorMode would behave differently in a couple of ways:

*For deferred computed observables:*
(same as before the #855 fix)

* they would not lead to computation
* when the computed is evaluated after subscribe, the subscription would be informed about the initial computed value

*For pure computed observables:*

* They would not awaken a sleeping pure computed
* The pure computed would fall asleep, if all "normal" subscriptions are disposed
* On awake, they would get triggered with the computed value.
* On falling asleep, they would get triggered with an "undefined" value.

It would also make sence to introduce a mode for pureComputed, stating, if the computed value should be cached, although the computed falls asleep. Then on awake it can compare the values and only notify the spectatorMode-subscriptions, if the value has changed since when it was awake last time.

We have already coded this in a patch, since we need it. But we'd like to discuss it before integrating it into knockout, submitting a pull request and implementing (more) tests.

This is a very important feature for us, since we have designed our api around it. And we would be very happy, if we didn't need to maintain our own fork, but find a solution together instead.

### DIFF
--- a/src/subscribables/dependentObservable.js
+++ b/src/subscribables/dependentObservable.js
@@ -100,6 +100,9 @@ ko.computed = ko.dependentObservable = function (evaluatorFunctionOrOptions, eva
                 });
                 _dependenciesCount = 0;
                 _latestValue = readFunction.call(evaluatorFunctionTarget);
+
+                dependentObservable['notifySubscribers'](_latestValue, null, true /* Only subscribers in suspectorMode */);
+
             } finally {
                 ko.dependencyDetection.end();
                 _isBeingEvaluated = false;
@@ -153,9 +156,7 @@ ko.computed = ko.dependentObservable = function (evaluatorFunctionOrOptions, eva
                     _latestValue = newValue;
                     if (DEBUG) dependentObservable._latestValue = _latestValue;
 
-                    if (suppressChangeNotification !== true) {  // Check for strict true value since setTimeout in Firefox passes a numeric value to the function
-                        dependentObservable["notifySubscribers"](_latestValue);
-                    }
+                    dependentObservable['notifySubscribers'](_latestValue, null, suppressChangeNotification == true /* Only subscribers in suspectorMode */);
                 }
             } finally {
                 _isBeingEvaluated = false;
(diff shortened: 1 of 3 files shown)


lcorneliussen:
That worked fine! Thank you very much for you help.

brianmhunt:
On the basis that this is a dupe of #1576 or otherwise @lcorneliussen thinks the workaround is good, I'll close this. Please let me know though if I'm jumping the gun here. :)

lcorneliussen:
Actually this is not yet solved. Tested a bit more...

When I subscribe to a computed that doesn*t register dependencies, it won't awake and hence i won't get any notification about the new value.

What about an "evaluated" event that just publishes all changes done to _latestValue?

mbest:
@lcorneliussen - I don't know that we'll add an "evaluated" event, but you might be able to generate it outside of Knockout. Maybe something like this:

```javascript
function myPureComputed (evaluator) {
    var wrappedEvaluator = function () {
        var value = evaluator();
        computed.notifySubscribers(value, 'evaluated');
        return value;
    }; 
    var computed = ko.pureComputed(wrappedEvaluator);
    return computed;  
}
```

lcorneliussen:
Hm. right. or just live with the fork. again. 

"evaluate" is also misleading - shouldn't fire if it didn't change.

"change" is perfect. Just one that actually only fires when _latestValue is changed and does nothing else :-)

mbest:
> "change" is perfect. Just one that actually only fires when _latestValue is changed and does nothing else :-)

That might be possible. Can you open a new issue for this?

lcorneliussen:
"cache" or "change"? I think "cache" is a good name. The value in the cache was updatede - for whatever reason.

lcorneliussen:
Added "cache" in an edit - maybe you didn't see it.

mbest:
> Added "cache" in an edit - maybe you didn't see it.

Which issue?

lcorneliussen:
This one :-) #1511

lcorneliussen:
I'll open a new issue

Issue #1531, opened 2014-09-10
Performance problem with peek on pureComputed (and foreach-binding)


lcorneliussen:
When a pureComputed is peeked while it is sleeping, it will always evaluate (like a "normal" function), but keep sleeping.

A subsequent subscribe, will lead to another evaluation.

In #652 a peek was introduced to the `makeTemplateValueAccessor` of foreach.
https://github.com/knockout/knockout/blob/master/src/binding/defaultBindings/foreach.js#L7

If the peekedValue is an array, the underlying observable is passed to the template binding, which will unwrap it to collect dependenies. If it is not an array, the item is unwrapped again just a couple of lines below (now in order to collect dependencies).

Hence, when using the foreach-binding together with a pureComputed, the pureComputed will allwas evaluate twice.

(I might have missed something - but that is what it looks like for me...)

lcorneliussen:
#1542 helps for the performance problem

PR #1543, opened 2014-09-18
Add caching to sleeping pure computed observables

mbest:
This is mainly in response to #1531, although I had the idea to do this while originally working on the pure computed feature.

### DIFF
--- a/src/subscribables/dependentObservable.js
+++ b/src/subscribables/dependentObservable.js
@@ -34,9 +34,11 @@ ko.computed = ko.dependentObservable = function (evaluatorFunctionOrOptions, eva
     function haveDependenciesChanged() {
         var id, dependency;
         for (id in dependencyTracking) {
-            dependency = dependencyTracking[id];
-            if (dependency._target && dependency._target.hasChanged(dependency._version)) {
-                return true;
+            if (dependencyTracking.hasOwnProperty(id)) {
+                dependency = dependencyTracking[id];
+                if (dependency._target.hasChanged(dependency._version)) {
+                    return true;
+                }
             }
         }
     }
--- a/src/subscribables/dependentObservable.js
+++ b/src/subscribables/dependentObservable.js
@@ -21,44 +21,45 @@ ko.computed = ko.dependentObservable = function (evaluatorFunctionOrOptions, eva
     if (typeof readFunction != "function")
         throw new Error("Pass a function that returns the value of the ko.computed");
 
-    function addSubscriptionToDependency(subscribable, id) {
-        if (!_subscriptionsToDependencies[id]) {
-            _subscriptionsToDependencies[id] = subscribable.subscribe(evaluatePossiblyAsync);
-            ++_dependenciesCount;
+    function addDependencyTracking(id, target, trackingObj) {
+        if (pure && target === dependentObservable) {
+            throw Error("A 'pure' computed must not be called recursively");
         }
-    }
 
-    function disposeAllSubscriptionsToDependencies() {
-        ko.utils.objectForEach(_subscriptionsToDependencies, function (id, subscription) {
-            subscription.dispose();
-        });
-        _subscriptionsToDependencies = {};
+        dependencyTracking[id] = trackingObj;
+        ++_dependenciesCount;
     }
 
     function disposeComputed() {
-        disposeAllSubscriptionsToDependencies();
+        if (!isSleeping && dependencyTracking) {
+            ko.utils.objectForEach(dependencyTracking, function (id, dependency) {
+                if (dependency.dispose)
+                    dependency.dispose();
+            });
+        }
+        dependencyTracking = null;
         _dependenciesCount = 0;
         _isDisposed = true;
         _needsEvaluation = false;
+        isSleeping = false;
     }
 
     function evaluatePossiblyAsync() {
         var throttleEvaluationTimeout = dependentObservable['throttleEvaluation'];
         if (throttleEvaluationTimeout && throttleEvaluationTimeout >= 0) {
             clearTimeout(evaluationTimeoutInstance);
-            evaluationTimeoutInstance = setTimeout(evaluateImmediate, throttleEvaluationTimeout);
+            evaluationTimeoutInstance = setTimeout(function () {
+                evaluateImmediate(true /*notifyChange*/);
+            }, throttleEvaluationTimeout);
         } else if (dependentObservable._evalRateLimited) {
             dependentObservable._evalRateLimited();
         } else {
-            evaluateImmediate();
+            evaluateImmediate(true /*notifyChange*/);
         }
     }
 
-    function evaluateImmediate(suppressChangeNotification) {
+    function evaluateImmediate(notifyChange) {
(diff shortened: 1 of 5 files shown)


mbest:
Since this change is all about performance, it's useful to check that there are no negative affects on performance. To check this, I set up a series of benchmark tests using a simple computed observable that just returns the value of a single observable. The following actions showed no significant change in performance:

1. Creating a normal computed
2. Creating a pure computed
3. Accessing a normal (or awake) computed
4. Updating a normal (or awake) computed

The following actions showed a significant change in performance:

1. Accessing a sleeping computed whose dependencies haven't changed: about 3 times faster
2. Accessing a sleeping computed whose dependencies have changed: about 40% slower (but about par with updating a normal computed)

As the computed observable itself becomes more complex, those two differences will improve as well.

@brianmhunt, @rniemeyer, @SteveSanderson, I'd like to get this merged in soon. So please let me know if you have any concerns.

brianmhunt:
Cool. This is all a bit beyond me, but I like the idea. Would every `ko.subscribable` will now have an associated version, which changes whenever the observable does?

mbest:
> Would every `ko.subscribable` will now have an associated version, which changes whenever the observable does?

Yes. Putting it in `ko.subscribable` allows the code to be shared between computed and plain observables.

mbest:
Thanks, Ryan.

src/subscribables/observable.js
```diff
@@ -25,8 +25,13 @@ ko.observable = function (initialValue) {
 
     if (DEBUG) observable._latestValue = _latestValue;
     observable.peek = function() { return _latestValue };
-    observable.valueHasMutated = function () { observable["notifySubscribers"](_latestValue); }
-    observable.valueWillMutate = function () { observable["notifySubscribers"](_latestValue, "beforeChange"); }
+    observable.valueHasMutated = function () {
+        observable.updateVersion();
```
SteveSanderson: Do you think the above line should perhaps be moved inside `notifySubscribers`? The version number would then mean "notification number" rather than "mutation number".

The reason I suggest this is that I think there's a lot of code out in the wild that does stuff like this:

    myObservable().subProperty = "New value";
    myObservable.notifySubscribers();

I know that's not the recommended pattern, but people may do it (and in practice, there are sure to be plenty of cases where it is done).

If the version continues to refer to mutations rather than notifications, pure computeds won't update correctly in this case, will they?

src/subscribables/dependentObservable.js
```diff
@@ -21,44 +21,56 @@ ko.computed = ko.dependentObservable = function (evaluatorFunctionOrOptions, eva
...
     }
 
-    function disposeAllSubscriptionsToDependencies() {
-        ko.utils.objectForEach(_subscriptionsToDependencies, function (id, subscription) {
-            subscription.dispose();
-        });
-        _subscriptionsToDependencies = {};
+    function haveDependenciesChanged() {
+        var id, dependency;
+        for (id in dependencyTracking) {
```
SteveSanderson: Normally we use `ko.utils.objectForEach` (or at least do a `hasOwnProperty` check) to avoid things getting messed up if people have JS environments with extra properties on `Object.prototype`.

src/subscribables/dependentObservable.js
```diff
@@ -250,23 +242,62 @@ ko.computed = ko.dependentObservable = function (evaluatorFunctionOrOptions, eva
...
-                _needsEvaluation = true;
-                evaluateImmediate(true /* suppressChangeNotification */);
-                if (!_isDisposed) {     // test since evaluateImmediate could trigger disposal
+                if (_needsEvaluation || haveDependenciesChanged()) {
+                    dependencyTracking = null;
+                    _dependenciesCount = 0;
+                    _needsEvaluation = true;
+                    evaluateImmediate();
+                } else {
+                    // First put the dependencies in order
```
SteveSanderson: I'm sure this manual order management is necessary, but I'm struggling a bit to understand why so (or what's different from before when we didn't need this). Could you clarify?

Also I don't see anything obvious in the added test that shows anything to do with dependency order manangement. In general the amount of test changes are quite small compared with the amount of change to the production code. Is this because our existing tests already demonstrate all the functionality added here, or would it be worth ensuring more of the cases accounted for in the code here are shown in the tests?

SteveSanderson:
In principle this sounds great. The way you've described the feature totally makes sense, and thanks for quantifying the perf impact.

I won't pretend to have made sense of every line in the diff (in fact, it's complex enough that I probably would have to reimplement it from scratch to understand it) but I'm fine with trusting that you know what you are doing with this :)

Some line comments added.

src/subscribables/observable.js
mbest: I also thought about using the notification, but if the observable is rate-limited, that would change when a sleeping computed would see the new value. Also how would it know when a sleeping computed has changed?

src/subscribables/dependentObservable.js
mbest: I can add the `hasOwnProperty` check. I'm not using `objectForEach` because it doesn't allow the loop to end early.

src/subscribables/observable.js
mbest: I suppose we could update the version in both cases (mutation and notification), at least for rate-limited and sleeping observables.

src/subscribables/dependentObservable.js
mbest: In Chrome, at least, when the object keys are numeric, it re-orders them in numeric order. In an earlier iteration of this feature, I had a test that failed without this code but it was mainly testing another feature that I didn't include. I'll see if I can modify the test to show the need for this.

mbest:
I split the code changes into two commits to help make the actual changes more clear. The first commit mainly re-organizes the code and doesn't change functionality

mbest:
I've added a test regarding the order of subscriptions. I hope it's clear enough.

mbest:
@SteveSanderson, I thought some more about the notification/mutation issue and now agree with you that the version should be changed on change notification. This makes a sleeping computed act more like a normal computed when dependent on a rate-limited observable. I've added some tests to show this.

rniemeyer:
I think this looks good to merge. @SteveSanderson - do you feel that your concerns were addressed?

SteveSanderson:
@mbest's comments are comforting :) I'll check the code and merge it in the morning hopefully!

SteveSanderson:
Thanks again @mbest! This looks like it was tricky to get right, so thanks for being so thorough with the testing. It's great that `ko.pureComputed` will now re-evaluate just as infrequently as a regular `ko.computed` while also retaining its memory-safe properties. This is cool stuff!

We have so much functionality around computeds by now that the logic in `dependentObservable.js` is *really* intricate... I think we've done a great job of keeping it robust by having such extensive test coverage, but it's absolutely at the limit of what I think is practical to maintain. I can't say I really understand it without mentally stepping through dozens of if/else cases and trying to contemplate all the scenarios. In the future, we'll have to focus on layering additional functionality on from outside (rather than expanding the core dependent observable logic) if at all possible.

Anyway, great to get this in :)

PR #1504, opened 2014-09-26
Components should render synchronously when possible

SteveSanderson:
While I love the syntax and design of the new components feature, the affect it has on rendering makes it a poor choice for many projects.  
- Since components are always rendered asynchronously, leveraging them causes lots of "janky" loading 

_If you've ever been scrolling down a webpage on your mobile device and been frustrated by the page constantly jumping and changing scroll position due to things loading up above, then you should understand the meaning here_
- Similarly, it causes a lot of [layout thrashing](http://wilsonpage.co.uk/preventing-layout-thrashing/).

Here is a clear illustration of what I'm talking about.  Let's say a page has two boxes of content which have shared code at the top.

---

CONTENT A
CONTENT B
CONTENT C

---

CONTENT A
CONTENT D
CONTENT E

---

Then you decide that since CONTENT A is shared code and decide to extract it out to a knockout component.   You might think you've simply re-factored your code, but by bringing in components you've actually changed the rendering order.   Whereby it used to render top-to-bottom, it will now render B,C,D,E first, and then after that the CONTENT A section will pop in after that (and it will seem jerky).

Given that projects using components are probably large enough that they are almost certainly bundling their javascript anyways, I really think components need to the capability to render syncronously, or at least somehow give the appearance that it is.

### DIFF
--- a/spec/components/loaderRegistryBehaviors.js
+++ b/spec/components/loaderRegistryBehaviors.js
@@ -212,6 +212,28 @@ describe('Components: Loader registry', function() {
         });
         expect(cachedLoadCompletedSynchronously).toBe(true);
         expect(getConfigCallCount).toBe(1); // Was cached, so no extra loads
+
+        // See that, if you use ko.components.get synchronously from inside a computed,
+        // it ignores dependencies read inside the callback. That is, the callback only
+        // fires once even if something it accesses changes.
+        // This represents @lavimc's comment on https://github.com/knockout/knockout/commit/ee6df1398e08e9cc85a7a90497b6d043562d0ed0
+        // This behavior might be debatable, since conceivably you might want your computed to
+        // react to observables accessed within the synchronous callback. However, developers
+        // are at risk of bugs if they do that, because the callback might always fire async
+        // if the component isn't yet loaded, then their computed would die early. The argument for
+        // this behavior, then, is that it prevents a really obscure and hard-to-repro race condition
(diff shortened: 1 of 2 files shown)


IPWright83:
I've a similar need to use a synchronous mode or at least have some sort of notification once complete which may help @laurentbourgaultroy with tests.

In my case I want to hook up some advanced events (being built in D3) but whichever way I seem to register to bindings I always get my function firing before the component has loaded properly. This means that my selectors always return 0 items, rather than the number of items I expect in the DOM.

reflectiveSingleton:
@IPWright83 I have written this feature into footwork.js (which is based on knockout). The method employed there accomplishes it using a few things:

* A Custom loader which wraps every component with a custom binding
* The custom binding is used to detect when the component is added to the DOM via the update() method

You can see guts of the implementation at: https://github.com/reflectiveSingleton/footwork/blob/master/source/framework/component/lifecycle.js

IPWright83:
Thanks, I came across someone mentioning custom bindings and have used that
method successfully. Appreciate the reply though.

benquarmby:
@SteveSanderson I am using Knockout exactly as you describe here:

> ...which in production might be all of them, e.g., if you're using AMD and have preloaded a bundle containing everything...

So the following actually seems very attractive, despite the aversion to global flags:

> We can consider going further and having a built-in global flag if it keeps coming up as a requirement (but hopefully it won't be necessary - we avoid global flags as much as possible).

For now I'll look at a custom loader that wraps the default.

Issue #648, opened 2012-09-19
Auto-fill by FF not seen by Knockout.js


beders:
Creating a login form with knockout.js is not working as expected if you use the auto-fill/password feature of Firefox.
Firefox will fill in remembered usernames and passwords, but that change is not seen by the view-model and thus the fields are empty if 'Sign in' is clicked.

I'll add a minimal example of the behavior.
To reproduce: Open index.html, fill in login/password --> The labels login and password are bound as they should.

Now hit sign it --> browser will ask you to store the password. Say yes.

Now reload the page --> login & password fields are pre-filled by browser, but the binding is unaware of that and Sign in will fail.

davidhouweling:
I've been investigating this issue on Safari v. 8.0.5 (the latest I believe) as well as I've noticed that it too doesn't play all that nicely either. When autofill is done, it does fire a textInput event, but the 'data' attribute is the previous value which the user typed in (before the selected autofill value). However, if you do 'target.value' it is the filled in value.

mbest:
Rather than trying to fix this in Knockout, I suggest that if you are running into this problem, to try using https://github.com/tbosch/autofill-event. It this doesn't work well with Knockout (it's designed for Angular), perhaps we can look into creating a Knockout-specific version.

PR #1564, opened 2014-10-01
Add textInput workaround for Firefox and IE bugs regarding auto-fill

mbest:
Discussion in #648.


mbest:
Rather than including this change in Knockout, I suggest that if you are running into this problem, to try using https://github.com/tbosch/autofill-event. It this doesn't work well with Knockout (it's designed for Angular), perhaps we can look into creating a Knockout-specific version.

PR #1576, opened 2014-10-08
computed observables notify "awake" and "asleep" events

mbest:
As described in #1511, this would facilitate the ability to have "spectator" subscriptions using a fairly simple wrapper.
1. A pure computed only awakens when a `change` subscription is registered and goes to sleep when there are no more `change` subscriptions. This means we can use other subscription types that don't affect the computed state. A deferred computed "awakens" when read or when a `change` or `beforeChange` subscription is registered.
2. When awakening, a pure or deferred computed will notify an `awake` event with the current computed value. And when going to sleep, a pure computed will notify an `asleep` event with an `undefined` value.

By overwriting `notifySubscribers`, you could use these new events and the `change` event to generate `spectate` events:

```
ko.subscribable.fn.notifySubscribers = (function (original) {
    return function (value, event) {
        if (!event || event == 'change' || event == 'awake' || event == 'asleep') {
            original.call(this, value, 'spectate');
        }
        original.apply(this, arguments);
    };
}(ko.subscribable.fn.notifySubscribers));
```

### DIFF
--- a/spec/subscribableBehaviors.js
+++ b/spec/subscribableBehaviors.js
@@ -88,9 +88,23 @@ describe('Subscribable', function() {
 
     it('Should be able to retrieve the number of active subscribers', function() {
         var instance = new ko.subscribable();
-        instance.subscribe(function() { });
-        instance.subscribe(function() { }, null, "someSpecificEvent");
+        var sub1 = instance.subscribe(function() { });
+        var sub2 = instance.subscribe(function() { }, null, "someSpecificEvent");
+
         expect(instance.getSubscriptionsCount()).toEqual(2);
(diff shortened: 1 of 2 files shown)


lcorneliussen:
What about also yielding an event "peek" or "peekSleeping" when a sleeping pure computed is peeked?
That would also fall into the "spectate" category.

lcorneliussen:
Could be a better option to fire an "evaluated" event - maybe even instead of awake. What I'm interested in is the first value.

That would then also fire on peek, and where a computed is evalutated by unwrapping or subscribing, but it doesn't register dependencies and gets disposed.

PR #1613, opened 2014-11-13
IE8 throws errors when calling apply with second argument undefined

klinden:
IE8 (and possibly earlier, I haven't tested), isn't able to handle calling apply with undefined for the parameters array. It instead must be called with an empty array. I haven't combed through the entire codebase to see if this is a problem elsewhere, but this is one spot that has caused problems for me in IE8. See http://stackoverflow.com/questions/20432707/knockout-removing-item-from-observablearray-throwing-error-in-ie8 for some more context to the issue. This JSFiddle demonstrates the behavior: http://jsfiddle.net/klinden/3sj2dkzj/3/


SteveSanderson:
Thanks!

PR #1628, opened 2014-11-21
Add hook for ko.onBindingError

brianmhunt:
This PR should exposes a function (`ko.onBindingError`) to hook into binding errors.

This should give greater control to users handling cases such as #1626 

The `function onBindingError(spec)` takes one object, `spec` containing the following properties:

| Property | Meaning |
| --- | --- |
| during | (string) The binding function name (`init` or `update` or `apply`) |
| errorCaptured | (Error) The error that was thrown by the binding function |
| element | (HTML Node) The node the binding was being applied to |
| bindingKey | (string) The name of the binding (`init` or `update` only) |
| valueAccessor | (function) As passed to the binding function (`init` or `update` only) |
| allBindings | (function) As passed to the binding function (`init` or `update` only) |
| bindingContext | (object) As passed to the binding function (contains `$data`, `$rawData`, `$root`, ...) |

Tested on PhantomJS, Chrome, Safari, and Firefox.

### DIFF
--- a/spec/bindingAttributeBehaviors.js
+++ b/spec/bindingAttributeBehaviors.js
@@ -472,6 +472,19 @@ describe('Binding attribute syntax', function() {
         }).toThrow("You cannot apply bindings multiple times to the same element.");
     });
 
+    it("Should call ko.onBindingError when applyBindings is called multiple times for the same element", function () {
+        var obe_calls = 0;
+        this.restoreAfter(ko, 'onBindingError');
+        testNode.innerHTML = "<div data-bind='text: \"Some Text\"'></div>";
+        ko.applyBindings({}, testNode);
+        ko.onBindingError = function (spec) {
+            obe_calls++;
+            expect(spec.during).toEqual('apply');
+        };
+        ko.applyBindings({}, testNode);
+        expect(obe_calls).toEqual(1);
+    });
+
     it('Should allow multiple applyBindings calls for the same element if cleanNode is used', function() {
         testNode.innerHTML = "<div data-bind='text: \"Some Text\"'></div>";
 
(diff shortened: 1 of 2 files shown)


brianmhunt:
Thank you @rniemeyer !

@mbest, @SteveSanderson - I am tagging this for 3.3 - even though I expect it would not make it in. This PR seems a fairly straightforward patch, would be handy for lots of folks, and it would be good to get it into the "stream" for release (expecting it'll be 3.4 or later, but great if it were good enough for 3.3).

The one concern I have is that errors will now be thrown when `handlerUpdateFn` is called, not just when it is bound/initialized. i.e. this differs from the current handling.

I would be very grateful for thoughts on how “mergeable” this looks.

Cheers

SteveSanderson:
@brianmhunt, could you let us know what (if anything) you think we should still do here, in light of #1715 having been merged now?

brianmhunt:
@SteveSanderson – I've been mulling it.

I am guessing the easiest way to preserve the functionality would be to directly attach the metadata to the error instances. We just have to be wary of clobbering anything currently or possibly in the future native to the Error, but the [spec for Errors](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) is pretty bare.

We could just change the `onBindingError` spec and respective calls so the spec looks like this:

```Javascript
    ko.onBindingError = function(error, spec) {
        var bindingText;
        ko.utils.extend(error, spec);
        if (spec.bindingKey) {
            // During: 'init' or initial 'update'
            bindingText = ko.bindingProvider['instance']['getBindingsString'](spec.element);
            error.message = "Unable to process binding \"" + spec.bindingKey + "\" in binding \"" + bindingText + "\"\nMessage: " + error.message;
        } // else: During: 'apply'

        if (ko['onError']) {
          ko['onError'](error);
        } else {
          throw error;
        }
   }
```

Does that sound – roughly – like a reasonable tact?

Incidentally, I noted that `ko.utils.catchFunctionErrors` rethrows an error even if `ko.onError` exists. Might it be preferable to leave it up to the user whether they want the error to rethrow? They can easily add it to their `ko.onError` handler.

Issue #1653, opened 2014-12-13
Lazily force re-evaluation of computed observable


VoiceOfWisdom:
I have a computed observable that I want to lazily execute, instead of executing immeaditly and always returning that value.

``` js
var result = ko.computed({
    read: function () {
      console.log('foo');
      return new Date();
    },
    write: function (newValue) {
        //not relevant
    },
});

//foo has been logged

setTimeout(function(){
  console.log(result().toString());
  //foo is not logged, and the value is the value that was initially returned.
}, 2000)
```
[JSBin](http://jsbin.com/yolazejane/1/edit?js,console,output)

Is there any way to tell knockout to re-evaluate this on the next call?
Looking at the source, there is the _needsEvaluation variable, however there is no way to force that to true that I can see.

VoiceOfWisdom:
@mbest, you were correct about changing the event type allowing me to fix my issue. I have been able to get most of my tests passing by calling subscribe with a `spectate` event type. 

I am running into one remaining issue, I need to manually call `trackChanges` on the `trackArrayChanges` extender. It is not exposed on the computed once you extend it. Would a PR making it public be acceptable?

mbest:
> I need to manually call `trackChanges` on the `trackArrayChanges` extender.

Can you explain this more? Have you looked at the changes made in #1634?

VoiceOfWisdom:
Sure.

I have a lazy observable object, that I dont want to eval until requested, even when subscribed to.
I still want it to support the `arrayChange` event. The `trackArrayChanges` extender calls `subscribe`
inside `trackChanges`, so I need to defer the execution of that function until the `computed` is read.
I can do this by overwritting the `subscribe` function, however without `trackChanges` being public
I cant ever call it to manually set up change tracking.

In #1634 you change to using `beforeSubscriptionAdd` to start tracking changes. While I could do some
tomfoolery with calling `beforeSubscriptionAdd` and intercepting its call to `underlyingBeforeSubscriptionAddFunction`
I think exposing `trackChanges` when you call `trackArrayChanges` is a clean and simple api change.

``` js

console.clear();

var makeLazy = function () {
    var holder = ko.observableArray();

    var needsArrayTracking = false;
    var setupArrayTracking = function () {
        if (needsArrayTracking) {
            needsArrayTracking = false;
            result.trackChanges();
        }
    };

    var notify = function (newValue) {
        result.notifySubscribers(newValue, 'spectate');
    };

    var hasAccessed = false;
[...]

PR #1654, opened 2014-12-14
Added deferSubscribeEvaluation option to computed observables.

VoiceOfWisdom:
Allows for completely lazy computed observables that don't evaluate until they are called.
Fixes issue #1653.

### DIFF
--- a/spec/computedBehaviors.js
+++ b/spec/computedBehaviors.js
@@ -1,0 +1,53 @@
+describe('Computed', function() {
+    it('Should evaluate on subscribe', function () {
+        var evaluated = false;
+        var computed = ko.computed({
+            read: function () {
+                evaluated = true;
+            },
+            deferEvaluation: true
+        });
+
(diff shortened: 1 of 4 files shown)


SteveSanderson:
Thanks for this suggestion!

I'm keen to hear @mbest's views on the correctness of this change, as he has much more insight into all the cases around computeds than anyone else.

I'd be happy enough to have `deferSubscribeEvaluation` functionality, though I'd be keen *not* to export `deferSubscribeEvaluation`/`hasEvaluated`/`onEvaluateImmediate` as public APIs. Those are *really* low-level APIs that virtually nobody is going to understand, and having them would very tightly lock us to increasingly specific implementation details (e.g., around the exact ordering of events).

mbest:
I think that the "plugin" that we discussed in #1653 handles the problem sufficiently, so we don't need to make any changes in Knockout to support this.

PR #1658, opened 2014-12-19
Allow `ko.applyBindings` to accept ´#document-fragment´s

mfeineis:
Adding HTML nodeType 11 (Document Fragment) to the allowed node types for `ko.applyBindings` allows knockout to operate on Shadow DOM Root nodes

### DIFF
--- a/src/binding/bindingAttributeSyntax.js
+++ b/src/binding/bindingAttributeSyntax.js
@@ -430,7 +430,7 @@
             jQueryInstance = window['jQuery'];
         }
 
-        if (rootNode && (rootNode.nodeType !== 1) && (rootNode.nodeType !== 8))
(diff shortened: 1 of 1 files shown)


SteveSanderson:
I agree that supporting Shadow DOM root nodes would be very valuable, and eventually will be essential as it becomes a primary technique for web developers.

However it might require more than this one-line tweak. For example, we might need to change the logic that detects when a bound DOM node has been removed from the document so that it treats Shadow DOM root nodes as equivalent to the primary document root node. Or maybe, for browsers that support Shadow DOM, we can assume they also support DOM mutation events and detect it that way.

If anyone is inclined to investigate this further and account for detecting when nodes have been removed (with tests), hopefully we could merge in such a pull request.

bbrandt:
Would the tests from PR #2479 be what is needed?

Issue #1638, opened 2014-11-27
Need a performant way to compose within foreach


AdamWillden:
I'm running into performance issues using nested [knockout foreach](http://knockoutjs.com/documentation/foreach-binding.html) bindings that heavily utilise [knockout components](http://knockoutjs.com/documentation/component-overview.html). In my case the bindings produce a dynamic table (columns and rows) with optionally modifiable cell contents.

I can use [knockout-repeat](https://github.com/mbest/knockout-repeat) to marginally speed things up but the fact it works on the outer html is causing me issues with components I've made, such that I have to do things to my code that 'breaks' the nice encapsulation I've done with components. The reason I'm forced to break encapsulation is because I'm using table (tbody, tr, td) elements within which you can't put anything 'unusual' (like a div).

Below you can see a stripped down version of what I'm doing currently. I've taken some styling bindings and component parameters out. It produces a nest as follows:
`*` **`<tbody>` foreach (50-100 items)**
`* *` **virtual component `<tr>`**
`* * *` **foreach (<15 items)**
`* * * *` **virtual component `<td>`**

Below is some of the table markup

``` html
<thead>
    <tr data-bind="foreach: visibleColumns">
        <!-- column header markup-->
    </tr>
</thead>
<tbody data-bind="template: { foreach: pageData }">
    <!-- ko component: {
        name: 'table-row',
        params: {
             sourceData: $data,
            columns: $parent.visibleColumns,
            selectMethod: $parent.config.selectMethod,
            isSelected: $parent.getSelectionIndicator($data)
            /*...etc...*/
        }
    } --><tr><td></td></tr><!-- /ko -->
</tbody>
```

and 'table-row' component markup

``` html
<tr>
<!-- ko if: selectMethod === 'check' -->
    <td class="omsui-table-rowSelect" data-bind="toggle: { data: isSelected, stylize: false }">
    {{#if: isSelected}}
        <i class="fa fa-fw fa-lg fa-check-square-o"></i>
    {{/if}}
    {{#ifnot: isSelected}}
        <i class="fa fa-fw fa-lg fa-square-o"></i>
    {{/if}}
    </td>
<!-- /ko -->
<!-- ko foreach: columns -->
    <!-- ko component: {
        name: 'table-cell',
        params: {
            sourceData: $parent.sourceData,
            column: $data,
            paths: $parent.paths,
            pathStatuses: $parent.pathStatuses
        }
    } --><td></td><!-- /ko -->
<!-- /ko -->
[...]

PR #1659, opened 2014-12-19
Execute async components.get callbacks in batches

AdamWillden:
Takes a step towards fixing #1638 

[This still needs unit testing however.](https://github.com/knockout/knockout/issues/1638#issuecomment-67704245)

### DIFF
--- a/src/components/loaderRegistry.js
+++ b/src/components/loaderRegistry.js
@@ -1,6 +1,8 @@
 (function(undefined) {
     var loadingSubscribablesCache = {}, // Tracks component loads that are currently in flight
-        loadedDefinitionsCache = {};    // Tracks component loads that have already completed
+        loadedDefinitionsCache = {},    // Tracks component loads that have already completed
+        callbacks = [],                 // Store each component 'get' callback for batch processing
+        batchTimerID = null;            // The current batch timeout ID
 
     ko.components = {
         get: function(componentName, callback) {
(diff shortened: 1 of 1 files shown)


AdamWillden:
Just adding a note here to read the comments on #1638 regarding animation frame batching. Might be a preferred method to batching. Although my poor in-application 'test' showed showed animation frame to be slower slower than batching I think someone needs to create a simpler performance test online to definitively compare them.

brianmhunt:
Thanks @AdamWillden – I was mulling this a bit, too.

There are quite a few performance comparisons between `requestAnimationFrame` and `setTimeout`, all seeming to show the prior being more performant e.g. http://jsperf.com/requestanimationframe-vs-setinterval-loop/7

I feel like `requestAnimationFrame` is preferable because it is designed to ["deal with things at the most convenient time for the browser"](http://calendar.perfplanet.com/2013/the-runtime-performance-checklist/), meaning it:

1. Does not draw items while they are not visible in the DOM;
2. Does not draw until the browser is ready (i.e. after any repaints, so no duplication before/after);
3. Slows the animation speed while the browser/tab is not focused.

... which will make CPU usage more efficient in general, and therefore also improve battery life. Where the reflows/repaints are complex, it could also dramatically improve performance over a setTimeout.

While experimenting with [knockout-fast-foreach](/brianmhunt/knockout-fast-foreach) I noticed that, where possible, one ought to batch `requestAnimationFrame` as done in this PR, otherwise there is a significant function-call overhead (i.e. iterating over an array of items in one frame is faster than `requestAnimationFrame` iterating over its list of callbacks).

Which is all to say, I think (but am always open to being shown to be wrong :)) that both batching and `requestAnimationFrame` could be used together and that would seem to be the best option.

@AdamWillden - I think it would be interesting to see what sort of performance you see if you combine batching with `requestAnimationFrame`.

AdamWillden:
@brianmhunt ah yeah that sounds like the best approach! Good thinking. 

I do agree the `requestAnimationFrame ` method should be better which is why I don't trust my poor man's test.

In the next week or so I'll combine the two methods and update this PR :-)

brianmhunt:
Awesome, thanks @AdamWillden

AdamWillden:
Right @brianmhunt I've made the changes. Seems to work (not bothered comparing performance to previous). Still needs tests.

brianmhunt:
Awesome, thanks @AdamWillden

mbest:
#1738 now handles this.

Issue #1592, opened 2014-10-28
Components and insertion points


unsafecode:
Speaking of nesting components, I've reworked a bit the latest available source for the component binding to also support the so called *insertion points*: imagine we want a **bootstrap-panel** component that wraps its content into a bootstrap panel. We would like to write our component as 

```html
<bootstrap-panel>
	<header>
		My panel
	</header>
	<main>
		<strong>My content</strong>
	</main>
	<footer>
		My footer
	</footer>
</bootstrap-panel>
```

Then, our template could look like:

```html
<div class="panel panel-default">
    <div class="panel-heading">
        <content select="header" />
    </div>
    <div class="panel-body">
        <content select="main" />
    </div>
    <div class="panel-footer">
       <content select="footer" />
    </div>
</div>
```

The key feature here is the `<content select="..." />` element, which specifies an *insertion point* by means of a *querySelectorAll* expression. At runtime, before binding occurs, and when the component template is loaded, I've set a function that processes all insertion points, looking a the newly available *originalChildNodes*.

The final output, then, will be:

```html
<bootstrap-panel>
	<div class="panel panel-default">
		<div class="panel-heading">
			<header>
				My panel
			</header>
		</div>
		<div class="panel-body">
			<main>
				<strong>My content</strong>
			</main>
		</div>
		<div class="panel-footer">
			<footer>
				My footer
			</footer>
		</div>
	</div>
</bootstrap-panel>
```

I really hope this approach can prove to be useful. At present time, though, I needed to edit the component binding source, since there's no pluggable way the edit it easliy, so if any of you are interested, I can probably build a PR and make it available.

VoiceOfWisdom:
I modified @vamp's interpolation implementation to use the parent binding context, and support durandal style `data-part` selectors.

[Here.](http://jsbin.com/tojahahefo/1/edit?html,js,output)

VoiceOfWisdom:
[New version that uses a binding handler instead of generating a massive with.](http://jsbin.com/boxabebiqu/1/edit?js,output)

novascreen:
i played around with that jsfiddle a bit more and also added the ability to use `<content></content>` for the remaining nodes
http://jsfiddle.net/q4ydg5mz/2/
i'm wondering, should this line `$content = $templateNodes.filter(query);` be using `filter` or rather do a deep `find` to be closer to what shadow dom does?

SteveSanderson:
[...]

PR #1665, opened 2014-12-29
Component transclusion

saynomoo:
Implementation of component transclusion, described for example in #1592. The transclusion code is taken from https://github.com/lelandrichardson/knockout-components

The implementation is done by first exposing the function cloneTemplateIntoElement into ko.components namespace, then the transclusion module replaces that function (in monkeyPatch.js). That is the only required change in the existing code to make the transclusion available.

To make the transclusion-enabled components more flexible, the defaultLoader was also changed a bit: The component definition may provide method findContent, which will allow programmatic use of composable components.

The features implemented in this pull request have been fully tested in transclusionBehaviors.js

### DIFF
--- a/src/components/componentBinding.js
+++ b/src/components/componentBinding.js
@@ -21,7 +21,6 @@
                 },
                 originalChildNodes = ko.utils.makeArray(ko.virtualElements.childNodes(element));
 
-            ko.virtualElements.emptyNode(element);
             ko.utils.domNodeDisposal.addDisposeCallback(element, disposeAssociatedComponentViewModel);
 
             ko.computed(function () {


aldendaniels:
@saynomoo - nice work. Per my comment in #1592, will this ensure that the binding context of the inject child nodes remains that of the parent component?

**My previous comment**
The injected content's binding context needs to be the parent component, not the wrapping component. For example:
```javascript
   <-- ko with:user -->
   <modal-dialog>
      <content>
         <b data-bind="text: firstName"></b>
      </content>
   </modal-dialog>
  <-- /ko -->
```
In this example, `firstName` needs to reference the user binding context, not the modal-dialog binding context.

Also note that the `parents` array should not contain any extra entries.

saynomoo:
@aldendaniels : unfortunately this implementation does not use the binding context of the parent. The feature is implemented with a simple substitution of dom nodes. I agreee that the behavior you suggest would be better.

aldendaniels:
@saynomoo - in case it proves useful - take a look at [this jsbin](http://jsbin.com/boxabebiqu/1/edit?js,output) that @VoiceOfWisdom put together in #1592 illustrating one approach to using the parent binding context.

joeheyming:
+1, is anyone looking at this?

SteveSanderson:
If I'm understanding this correctly, a lot of the functionality proposed here is already included in KO 3.3.0 via the `componentTemplateNodes` context property, which lets you [pass markup into components](http://knockoutjs.com/documentation/component-custom-elements.html#passing-markup-into-components).

Clearly this pull request takes it a bit further, with the whole selectors thing, but couldn't that be achieved without any additional code, like the following?

    <template id="my-component-template">
        Some markup
        
        Some more markup
    </template>

Note that this already takes care of passing through the binding context in KO's standard way.

Could anybody clarify what the goals for this pull request are that would not be adequately met by the KO 3.3.0 functionality? I appreciate it might just be that parts of this conversation predated KO 3.3.0's release.

Thanks for discussing this!

novascreen:
i have a scenario where the approach taken in #1592 seems to provide me with something `ko template:` doesn't, unless i'm missunderstanding something.
Let's say i want to wrap a jQuery plugin in a knockout component to reduce the markup, e.g. a carousel. Now when i use `ko template:` i can not initialize the jQuery plugin in the constructor because the markup is not all there yet and i think this is being discussed in another ticket, but right now i don't know of a way to easily detect when everything is resolved. So by distributing the insertion points before the ViewModel's constructor (in `createViewModel`) i know everything will be there.

I'm still new to knockout so i'm sorry if i misunderstood the current capabilities.

here is some sample code:

```html
<my-carousel>
  <img src="/src/images/iconics/the-great-bear/gallery1-lg.jpg" alt="">
  <img src="/src/images/iconics/the-great-bear/gallery2-lg.jpg" alt="">
  <img src="/src/images/iconics/the-great-bear/gallery3-lg.jpg" alt="">
</my-carousel>
```

```html
<div class="cycle-slideshow" data-cycle-auto-height="16:9" data-cycle-timeout="0">
  <div class="cycle-pager"></div>
  <div class="cycle-prev"></div>
  <div class="cycle-next"></div>
  <content>
    <img src="/src/images/fallback.jpg" alt="">
  </content>
</div>
```

Another note and this is mostly optics, but i like the idea of this looking and working the way Shadow DOM does.

ranjanmadhu:
I tried the solution by @SteveSanderson but it is not working am i missing anything? 
https://jsfiddle.net/madhu_ranjan/Lz1dzub8/2/

barake:
@ranjanmadhu 
The first node in `$componentTemplateNodes` is the node you're searching for, and jQuery is only searching descendants. The following works, but is ugly:
```javascript
$($componentTemplateNodes).find('header').addBack('header')
```

PR #1666, opened 2014-12-29
Pass bindingContext as a parameter to the viewModel constructor

saynomoo:
By providing the bindingContext to the viewModel constructor, we are able to create sub-components with less boilerplate code.

For example the live example 2, from page: http://knockoutjs.com/documentation/extenders.html could be done with components like this:

First register an auxiliar component, which assumes the model has a observable called value
```
    ko.components.register('required-text', {
      viewModel: function(params, bindingContext) {
        this.visible = ko.pureComputed(function(){
          return bindingContext.$rawData.value() == ""
        })
      },
      template: '<span data-bind="visible: visible">This field is required</span>'
    });
```

Bind the context:
```
var appModel = {
  firstName: {value: ko.observable("John")}, 
  lastName: {value: ko.observable("Doe")}
}
 
ko.applyBindings(appModel);
```

With markup:
```
            <div data-bind="with: firstName">
              <input type="text" data-bind="value: value, valueUpdate: 'afterkeydown'" />
              <required-text></required-text>
            </div>
            <div data-bind="with: lastName">
              <input type="text" data-bind="value: value, valueUpdate: 'afterkeydown'" />
              <required-text></required-text>
            </div>
```

Currently the required-text -component would require passing parameters in the markup like:
```
            <div data-bind="with: firstName">
              <input type="text" data-bind="value: value, valueUpdate: 'afterkeydown'" />
              <required-text params="value: value"></required-text>
            </div>
            <div data-bind="with: lastName">
              <input type="text" data-bind="value: value, valueUpdate: 'afterkeydown'" />
              <required-text params="value: value"></required-text>
            </div>
```

### DIFF
--- a/src/components/componentBinding.js
+++ b/src/components/componentBinding.js
@@ -49,7 +49,7 @@
                         throw new Error('Unknown component \'' + componentName + '\'');
                     }
                     cloneTemplateIntoElement(componentName, componentDefinition, element);
-                    var componentViewModel = createViewModel(componentDefinition, element, originalChildNodes, componentParams),
+                    var componentViewModel = createViewModel(componentDefinition, element, originalChildNodes, componentParams, bindingContext),
                         childBindingContext = bindingContext['createChildContext'](componentViewModel, /* dataItemAlias */ undefined, function(ctx) {
                             ctx['$component'] = componentViewModel;
                             ctx['$componentTemplateNodes'] = originalChildNodes;
@@ -75,10 +75,10 @@
         ko.virtualElements.setDomNodeChildren(element, clonedNodesArray);
     }
 
-    function createViewModel(componentDefinition, element, originalChildNodes, componentParams) {
+    function createViewModel(componentDefinition, element, originalChildNodes, componentParams, bindingContext) {
         var componentViewModelFactory = componentDefinition['createViewModel'];
         return componentViewModelFactory
-            ? componentViewModelFactory.call(componentDefinition, componentParams, { 'element': element, 'templateNodes': originalChildNodes })
+            ? componentViewModelFactory.call(componentDefinition, componentParams, { 'element': element, 'templateNodes': originalChildNodes }, bindingContext)
             : componentParams; // Template-only component
     }
 
--- a/src/components/defaultLoader.js
+++ b/src/components/defaultLoader.js
@@ -136,8 +136,8 @@
             // By design, this does *not* supply componentInfo to the constructor, as the intent is that
             // componentInfo contains non-viewmodel data (e.g., the component's element) that should only
             // be used in factory functions, not viewmodel constructors.
-            callback(function (params /*, componentInfo */) {
-                return new viewModelConfig(params);
+            callback(function (params, componentInfo, bindingContext) {
+                return new viewModelConfig(params, bindingContext);
             });
         } else if (typeof viewModelConfig[createViewModelKey] === 'function') {
             // Already a factory function - use it as-is


joutsen:
Speaking about tight coupling, we have quite the opposite experience with the lack of this feature. Not being able to access data in VM constructor other than via 'params' forces developer to always pass data from HTML (isn't it already quite tight coupling?). This disables capability of VM constructor to inject it's application logic from some application business object, e.g. $root (unless this object is always passed via params repeatedly). This in turn disables building DI architectures on top of Knockout components (for the sake of loose coupling :). In other words: by limiting the possibilities to misuse the components one will also limit the most advanced and interesting ways of using the components. We need this feature!

mbest:
I think that if we are going to pass the binding context to a component, it should be as part of the `componentInfo` parameter to `createViewModel`.

joutsen:
It would be also great to pass $element to VM constructor. Same as to custorm binding. This will enable the following configuration
<my-button red small></my-button>
In order to access "variation" attributes from VM I need an access to $element.

reflectiveSingleton:
@joutsen you can do that currently with the params attribute:

```html
<my-button params="variation: [ 'red', 'small' ]"></my-button>
<!-- or some other variant that passes in those differing 'variations' -->
```

joutsen:
Sure I can. But isn't the former way more elegant?

brianmhunt:
I feel like the workarounds are, as @joutsen points out, fairly clunky and un-markup-like.

Another option, also clunky though, is:

```javascript
<my-button red small params='$element: $element'></my-button>
```

Then in the view model function/constructor use `getAttribute` or `$.attr` to check the attributes.

I feel like direct access to `$context` in a component view-model constructor, while it violates the separation of concerns, substantially eases this pain from a practical perspective. Make no mistake though, exposing `$context` opens the floor for lots of poor design choices.

joutsen:
Not just that. Please take into account that my-button takes it's label and click handler from application business object. To access that object we need $root. So the complete example would be
```html
<my-button red small id='saveButton' params='$element: $element, $root:$root'></my-button>
```
And if passing element and bindingContext to VM constructor opens the door for bad designs, why this is done for custom bindings then and also why [knockout-components](https://github.com/lelandrichardson/knockout-components) does it?

mbest:
> It would be also great to pass $element to VM constructor

It's already passed to [`createViewModel`](http://knockoutjs.com/documentation/component-registration.html#a-createviewmodel-factory-function).

rniemeyer:
I agree with @mbest that I think it makes the most sense to include the context in the `componentInfo` parameter of `createViewModel`.

webketje:
+1 @mbest & @rniemeyer; it makes most sense to pass the context as a property on `componentInfo`

SteveSanderson:
We originally made a deliberate decision *not* to supply any view information (element, binding context, etc.) to component constructors by default, because we thought it would guide people to make bad design choices. Our compromise was to supply it only to `createViewModel` so a developer can still do this if they want to, but it's deliberately de-emphasised. If we pass it to the viewmodel constructor (e.g., as a property of `componentInfo`) then we set people up to directly mutate their DOM from their component viewmodel, which eliminates many of the benefits.

I would therefore propose not to do this. Developers who don't care about separating their view from component viewmodels can already make their own `createViewModel` function that passes whatever information they want into their component viewmodel. 

@mbest / @rniemeyer / @brianmhunt - if you feel otherwise, please say so. I'm willing to be overruled here - I'm just stating my preference!

joutsen:
Do I understand correctly that once we know the element of the component from componentInfo, we can actually get the binding context of it using ko.contextFor(element)? If I would know this from the very beginning I wouldn't propose this patch at all. Thanks for solution.

ThomasMichon:
I agree with @SteveSanderson here; we should not pass the binding context to the constructor of the view model by default; specialized components can use the alternate ```createViewModel``` factory method, within which they can extract a binding context using ```ko.contextFor(componentInfo.element)```.

I do not think we should go through with this pull request. I have a hard time as it is convincing my colleagues to write "pure" view model classes; giving all components easy access to the binding context will discourage component authors from separating concerns.

mbest:
I'm closing this, since there's an existing way to get the binding context without much trouble:

```
ko.contextFor(componentInfo.element);
```

joutsen:
Fine with me. The described solution has been extensively tested and works well in all our situations.

Issue #1525, opened 2014-09-08
Split out Knockout Components from core into a separate module


kmalakoff:
Regarding https://github.com/knockout/knockout/issues/1517, I've been giving it some more thought and believe you should consider removing the components framework from Knockout core and releasing it separately.

The scope of Knockout core should be more around data-binding rather than application architecture.

kmalakoff:
@stalniy I really do believe I should have to dive into the internals of the Knockout build system to use the Knockout library. 

If you provide a way to bower install and npm install your proposal, I'll take it.

brianmhunt:
@kmalakoff I'm with you — it'd be delusional to expect users to recompile Knockout to remove components. As awesome as components are, I am also on "team-component-plugin". That said it's not practical to rip it out now. What *can* be done is keeping this issue in mind and making sure that the knockout components code remain at least nominally loosely coupled in future patches.

Speculating, it'll probably be 4.0 or later before that happens, and if components are someday transformed into a plugin it'll probably (and not to be presumptuous, but just thinking out loud) coincide with a change in the nature of plugins e.g. #1589 .

Over time I have come to notice based on my own practices that there are some “first-class” plugins that make sense to bundle with the core package (e.g. components, knockout-punches, knockout-projections), and a slew of ”second-class” packages that are hosted elsewhere. The delineation between first and second class should only ever be hosting-for-convenience, with no special design privileges in the core that prejudice the second-class. (In this sense I feel "first" and "second" class may be derogatory language already ... maybe "core" plugins and just "plugins", but for the purpose of this paragraph, it's illustrative).

That said, I do not know what the other maintainers are thinking the direction of knockout components ought to be, but I'm sure it'll come up in discussion, and I'll raise this if no-one else does. :)

rniemeyer:
Backwards compatibility has always been a key concern and goal for KO, so it is not practical to pull components into a plugin in the short-term. 

Including this functionality in core vs. as a plugin was discussed during the original design. In the end, we wanted the components functionality to be light and extensible (plugins can be built on top of it), but available to new/all KO developers from day one as a pattern for development vs. new folks commonly ending up with large, hard-to-maintain view models as their application grows.  It has always been possible to write modular, reusable code with KO, but the barrier was often high for those that were new to KO.

kmalakoff:
In Knockback, I provide two versions: 1) core 2) full. 

Users can then choose...a full version with the major plugins bundled or start from core and add the plugins they want. 

Easy peasy.

Issue #1517, opened 2014-09-03
Component API should not directly use require 


kmalakoff:
I've been looking into https://github.com/webpack/webpack/issues/424 related to failed compile-time analysis due to Knockout's new component framework. The following causes side effects and I'd like to make a case for an API change:
- [passing require on AMD](https://github.com/knockout/knockout/commit/847e1f199eb2057b40a2c7938b4df332450302f9)
- [assuming a require-based module loader](https://github.com/knockout/knockout/commit/847e1f199eb2057b40a2c7938b4df332450302f9#commitcomment-7555943) - in possiblyGetConfigFromAmd
- [always using async loading](https://github.com/knockout/knockout/issues/1504) - causes flickering for me when I am using synchronous loading

At the same time of looking into the webpack issue, I've also been exploring components and serendipitously, I wrote my own loader before trying the Knockout one. These were the principles I followed:
- do not assume the component is written with Knockout - this means the API should have no notion of View Models so the API should work on elements, etc
- allow for DOM cloning - do not even ask for the template, but only the element
- allow for immediate, synchronous loading
- make no dependency on the module system - the caller should handle the module system so avoid problems with solutions like webpack or even allowing other solutions like Angular's module system  
- use Node.js callbacks - keep the pattern simple ad standard (no special errorCallback)

Here's what the synchronous API could look like:

```
ko.components.register 'component-1', {
   create: (data) -> $(<div>Hello</div>)[0]
   destroy: (el) -> # optional if the component framework removes the element (I'd recommend ko.removeNode(el))
}
```

or

```
ko.components.register 'component-1', {
   create: (data) -> ko.applyBindings({name: 'Bob'}, el = $(<div data-bind="text: name"></div>)[0]); return el
   destroy: (el) -> # optional if the component framework removes the element (I'd recommend ko.removeNode(el))
}
```

Here's the extension for an sync API:

```
ko.components.register 'component-1', {
   create: (data, callback) -> require ['component-1'], (component) -> component.create(callback)
   destroy: (el, callback) -> require ['component-1'], (component) -> component.destroy(el, callback) # or the caller could cache a dispose/destroy/release method directly on the element that they can call synchronously without the callback
}
```

You should be able to check if the create function is asynchronous by checking the number of arguments. componentOptions.create.length == 1 or 2.

[...]

PR #1526, opened 2015-01-09
Component API should not directly use require

mbest:
Regarding https://github.com/knockout/knockout/issues/1517, the component API should be updated to remove directly using require so that static analysis works by build systems: https://github.com/webpack/webpack/issues/424.

Alternatively, you could move the components module into its own separate module: https://github.com/knockout/knockout/issues/1525

### DIFF
--- a/build/fragments/amd-pre.js
+++ b/build/fragments/amd-pre.js
@@ -1,14 +1,13 @@
 (function(factory) {
     // Support three module loading scenarios
-    if (typeof require === 'function' && typeof exports === 'object' && typeof module === 'object') {
-        // [1] CommonJS/Node.js
-        var target = module['exports'] || exports; // module.exports is for Node.js
-        factory(target, require);
-    } else if (typeof define === 'function' && define['amd']) {
-        // [2] AMD anonymous module
+    if (typeof define === 'function' && define['amd']) {
+        // [1] AMD anonymous module
         define(['exports', 'require'], factory);
+    } else if (typeof require === 'function' && typeof exports === 'object' && typeof module === 'object') {
+        // [2] CommonJS/Node.js
+        factory(module['exports'] || exports);  // module.exports is for Node.js
     } else {
         // [3] No module loader (plain <script> tag) - put directly in global namespace
         factory(window['ko'] = {});
     }
-}(function(koExports, require){
+}(function(koExports, amdRequire){
--- a/src/components/defaultLoader.js
+++ b/src/components/defaultLoader.js
@@ -194,8 +194,8 @@
     function possiblyGetConfigFromAmd(errorCallback, config, callback) {
         if (typeof config['require'] === 'string') {
             // The config is the value of an AMD module
-            if (require || window['require']) {
-                (require || window['require'])([config['require']], callback);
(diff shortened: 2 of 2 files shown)


kmalakoff:
@brianmhunt we are using webpack for CommonJS and AMD.

@sokra sorry, I should have just said "the require API through CommonJS". 

By Knockout dropping the passing of require to their factory function for CommonJS, then they lose the ability to pass a require string.

From getConfig(name, callback): `{ template: 'someElementId', viewModel: { require: 'myModule' } }`

The Knockout web component API/documentation is a bit too complex for me to understand so I might be wrong about this being a feature.

mbest:
I've attached the code changes as suggested by @sokra.

@brianmhunt, I don't think using `window['ko'] = window['ko'] || {}` is a good idea. I think loading knockout.js should always set up a clean copy of the object.

brianmhunt:
@mbest You're probably right about a clean copy- it just strikes me as something that'd be the source of mysterious problems if that's how it was being handled, and anyone explicitly including KO by script tags would probably, if they intended to reload KO, be making a copy or removing it from `window` before the reload - otherwise it'd probably be an error. Maybe. Just a thought. :)

brigand:
There should be an overloadable implementation of the 'get async' behavior.  This would allow async loading with various module bundlers.  The most common case is require.js, so I think it's a sensible default, however including support for one library would be very unlike knockout.  Also including support for multiple libraries would bloat the core, and make it even worse.

In my opinion, there should be no async loading of view models by default, or there should be a technique that works without any module systems, where knockout just includes a script with the provided url, and the script calls a function to register the view model.

Require.js specific support should be in a separate plugin, like everything else is.  

About commonjs, there is a require.async which isn't implemented in node/browserify, but is implemented in webpack.

jods4:
I am using KO components in an AMD / require.js context so if my feedback may help, here's my experience.

I started naively, passing a `{ require: 'text!components/myTemplate.html' }` as the template to KO. At first everything was good.
Then I tried to create a few bundles to deploy my application in less than 100 different files. I used the official `r.js` tool. Of course, KO dynamic require code breaks r.js automatic dependency detection.

I could have fixed this by explicitely writing dependencies in my r.js config, but that felt like tedious, non-required work. So I did it another way.
Each one of my KO component lives in its own AMD module. I simply listed the html template as an AMD dependency and passed it as a string to the component view. Everything works nicely again!

Did I loose anything with this transform? Kinda. Now the html template is fetched when my component is registered rather than when it is first instanciated. This doesn't matter at all for me because I'm bundling the template and component in a single file anyway. I could imagine some more complex scenarios where this may be slightly inconvenient, but I think I would handle them without using the require option anyway.
But as a bonus, my components will bind synchronously in a future KO release, which has its fair share of benefits ;)

Regarding the core issue: I agree that it feels wrong to hard-wire require in KO. Especially with HTML5 modules coming up. Moreover, using `require` naively like I did above will do you no good in production, you surely want a more clever loading strategy. It's probably wise to just deprecate the `require` option and let people who want to do complex async view loading at binding time use a custom loader -- or add a new factory function option.

PR #1360, opened 2014-03-18
#1039 build for bower (& migration from Grunt to Gulp)

brianmhunt:
**This PR is keen for review — see comments below.**

This PR changes the task runner from Grunt to Gulp, in an attempt to achieve some desirable improvements in the build and distribution process (including issue #1039).

### New design
Here are some of the advantages of this design:

1. Uses gulp. Gulp, like grunt, is awesome. Here's a [getting started link](http://markgoodyear.com/2014/01/getting-started-with-gulp/).
2. Clean tasks centralization in `gulpfile.js`
3. Get rid of a bunch of superfluous testing code
4. Add jshint linting (`$ gulp lint`)
5. Separate `build` and `build-debug` tasks
6. Automatically rebuilds `spec/runner.html` and also `spec/runner.jquery.html` and `spec/runner.modernizr.html`.
7. Source files used for building are now in a JSON file: `build/fragments/source-references.json`
8. Spec files are now in a JSON file: `spec/helpers/specs.json`
9. Substantially simplified `runner[.*].html`
10. Dependencies on jQuery and Jasmine and Modernizr are now variables in the task config.
11. CI with Travis and Testling (which has [not worked in several months](https://ci.testling.com/knockout/knockout))
12. Uses livereload so that browsers with the unit tests will automatically reload on script changes

### TODO
- [x] Switch to closure compiler
- [x] TravisCI working
- [x] Testling working
- [x] Livereload for local testing
- [x] Add "dist" task that bumps version, creates `dist/knockout[.min].js`, and publishes to npm
- [x] check for trailing spaces in files
- [x] replace ##VERSION##

PR #1675, opened 2015-01-10
Fixed JSHint warnings

garrypas:


brianmhunt:
Thanks @garrypas .

The jshint updates are great, thank you for doing that. I'll try and merge these just after the upcoming 3.3.0 (which should coincide with a gulp build PR #1360 that does auto-jshint).

The native `array.map` is ([by design](http://www.ecma-international.org/ecma-262/5.1/#sec-15.4.4.19)) [much, much slower](http://jsperf.com/map-reduce-named-functions/2) than the `utils.arrayMap` equivalent. Performance aside, is there a benefit you had in mind for this change?

Cheers.

garrypas:
@brianmhunt I just did some benchmarking and you are absolutely correct. That's a big surprise to me. I've therefore reverted the changes related to Array.map().

brianmhunt:
@garrypas Crazy, eh? Thanks for the update.

src/subscribables/observableArray.changeTracking.js
```diff
@@ -107,6 +108,7 @@ ko.extenders['trackArrayChanges'] = function(target) {
 
             case 'pop':
                 offset = arrayLength - 1;
+            break;
```
mbest: This is wrong (as is the earlier added `break`).

src/subscribables/observableArray.changeTracking.js
```diff
@@ -121,7 +123,7 @@ ko.extenders['trackArrayChanges'] = function(target) {
                     endAddIndex = startIndex + argsLength - 2,
                     endIndex = Math.max(endDeleteIndex, endAddIndex),
                     additions = [], deletions = [];
-                for (var index = startIndex, argsIndex = 2; index < endIndex; ++index, ++argsIndex) {
+                for (index = startIndex, argsIndex = 2; index < endIndex; ++index, ++argsIndex) {
```
mbest: Where is `index` defined now?

garrypas: Hi, its on line 104 (was line 103)

brianmhunt: Linter dividends! :)

PR #1677, opened 2015-01-11
Consistent this keywork context for ko.utils.array functions

garrypas:
Consistent this keywork context for ko.utils.array functions #1482

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 ##Getting started
 
+[![Join the chat at https://gitter.im/knockout/knockout](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/knockout/knockout?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
+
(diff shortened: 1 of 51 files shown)


brianmhunt:
Looks good --- do we need tests- the results of this PR *seem* fairly apparent.

What happens when `this` is undefined (as it often would be)? Anything nefarious on any browsers?

garrypas:
I tested it in IE6-IE11 and Chrome and the behaviour is consistent with arrayFirst and behaves the same across these browsers - if a third argument is supplied it is used, otherwise 'this' defaults to the global context.

SteveSanderson:
Looks good - thanks for contributing! I'd be happy to have these changes.

Would you add tests to the pull request?

mbest:
I think we should be cautious about making this change, since a direct function call is generally much faster than using `call`.

mbest:
Merged as 62d568379d7d690a677793d4a851e52816fa6ab9 and 068a0417e215d61eb40381c0f18b564a03254b77

PR #1684, opened 2015-01-25
Added 'throttleSubsequentChanges' rateLimit method

damianpowell:
Hi,

Thanks for Knockout.js. It's a very useful tool.

This change is small but I find it useful when I want to make an immediate change and then throttle subsequent changes. I typically use this in search results where I would clear the results before making a service call, then reinstate the view with new results when the service returns (although it's more useful when there are multiple search services running in parallel).

I hope it's useful to you.

Thanks,
Damian.

### DIFF
--- a/src/subscribables/extenders.js
+++ b/src/subscribables/extenders.js
@@ -30,7 +30,13 @@ ko.extenders = {
             method = options['method'];
         }
 
-        limitFunction = method == 'notifyWhenChangesStop' ?  debounce : throttle;
+        switch (method)
+        {
+        	case 'notifyWhenChangesStop':     limitFunction = debounce;                  break;
+        	case 'throttleSubsequentChanges': limitFunction = throttleSubsequentChanges; break;
+        	default:                          limitFunction = throttle;                  break;
+        }
+
         target.limit(function(callback) {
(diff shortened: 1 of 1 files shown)


brianmhunt:
Many thanks @damianpowell - it looks like it could be quite useful. Before merging for a pubic release we should add a few unit tests to verify compatibility across platforms and prevent regressions (and any contributions in this vein would be most welcome!) 

Cheers

damianpowell:
Hi Brian,

You're welcome! I agree about the tests. I had a look (an admittedly brief one) for existing tests around the rateLimit extender but I didn't spot anything obvious. If you point me in the right direction, I'll gladly add to them.

TTFN,
Damian.

brianmhunt:
Hi Damian,

If I were putting the tests together, I think I'd aim to put them in around here: 

https://github.com/knockout/knockout/blob/ddcfbfc89f877d9e0c44b4269f13c1062484f9b1/spec/asyncBehaviors.js#L114

You will see a few similar tests that should offer a decent baseline.

Cheers

mbest:
Rather than adding new rateLimit methods into Knockout, with #2405, you can just use whatever method you like.

PR #1704, opened 2015-02-24
Update selectedOptions binding to not reset option selected status if not necessary

srd8580:
It causes the IE issue outlined here:
http://stackoverflow.com/questions/28683205/knockoutjs-multiselect-issue-in-internet-explorer

No side effects detected.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "knockout",
-  "version": "3.2.0",
+  "version": "3.3.0",
   "homepage": "http://knockoutjs.com/",
   "description": "Knockout makes it easier to create rich, responsive UIs with JavaScript",
   "main": "dist/knockout.js",
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "knockout",
   "description": "Knockout makes it easier to create rich, responsive UIs with JavaScript",
   "homepage": "http://knockoutjs.com/",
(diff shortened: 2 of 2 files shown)


SteveSanderson:
Thanks for this!

Any chance you could add a test to show that this makes a difference? Otherwise this behavior could get lost in a future update, since looking at the code it's not obvious that your change would cause a change of behaviour.

srd8580:
Hi Steve, thanks for looking at this.  

Unfortunately, I'm at a loss to create a Jasmine test for this as you implied, there actually is no change in behaviour at all.  The issue this change resolves is purely performance related.  I am unfortunately still supporting IE8, and the performance improvement on multi-select elements with a large number of options is dramatic with this change in IE8.  It goes virtually unnoticed in other browsers, however.

I would happily put together a JS Fiddle if it would help getting it merged?

srd8580:
I almost forgot, this JSFiddle http://jsfiddle.net/unp9j9dc/1 shows a UI "bounce" issue in IE also, again, not reproducible in Jasmine or non-IE browsers.

mbest:
I decided to fix this a different way (see the attached commit) when I noticed that Chrome also changes the scroll position when updating the `<select>` from the view model. I ended up including a similar check as yours, though, to prevent the `<select>` from flashing in IE.

srd8580:
This is great guys!  Thanks a lot.

PR #1718, opened 2015-03-09
ko.toJS no longer replaces RegExp objects with plain objects

dancras:


brianmhunt:
Thanks @dancras – I am tagging this for milestone 3.4.0. I do not foresee issues with this patch, but I am leaving it open for review later in the release cycle just in case.

Cheers

SteveSanderson:
Great - thanks!

PR #1732, opened 2015-03-17
(fixes #1731) Expose dependentObservable-options in debug mode

### LINKED ISSUE #1731: Debugging: Expose computed options (including read/write functions)
When debugging it helps a lot to be able to see a computeds options (read, write, e.g.) in the developer tools.

    if (readFunction && typeof readFunction == "object") {
        // Single-parameter syntax - everything is on this "options" param
        options = readFunction;
        readFunction = options["read"];
    } else {
        // Multi-parameter syntax - construct the options according to the params passed
        options = options || {};
        if (!readFunction)
            readFunction = options["read"];
    +    else
    +       options.read = readFunction;
    }
    if (typeof readFunction != "function")
        throw new Error("Pass a function that returns the value of the ko.computed");

    + if (DEBUG) dependentObservable._options = options;

lcorneliussen:
Suggestion for #1731: Expose dependentObservable-options in debug mode

### DIFF
--- a/src/subscribables/dependentObservable.js
+++ b/src/subscribables/dependentObservable.js
@@ -17,9 +17,13 @@ ko.computed = ko.dependentObservable = function (evaluatorFunctionOrOptions, eva
         options = options || {};
         if (!readFunction)
             readFunction = options["read"];
+        else // write the read function back to the options
+            options.read = readFunction;
     }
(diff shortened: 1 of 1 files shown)


SteveSanderson:
Thanks - merged! (Slightly tweaked, but should be the same effect - let me know if not).

PR #1733, opened 2015-03-17
nativeTemplateEngine: Multi-node template

lcorneliussen:
Allow creating templates from multiple nodes directly (without "root" node)


SteveSanderson:
Thanks for the suggestion. Are you sure this one-line tweak is all that's needed to make this work properly everywhere? I thought the reason we didn't already do this is that it's quite nontrivial to ensure things update and get disposed properly in all cases if there isn't a single root node.

If you are confident this is all that's needed, it would be a cool feature to add :+1: . Would you consider expanding this to a full pull request with tests to cover all the relevant scenarios around updates and disposal?

lcorneliussen:
Hi, just saw that I didn't answer this yet :(

Actually I'm not sure if disposal works as expected.

But i looks like #1839 could enable this; one would then have to add the elements to the root node and then clone/parse...

PR #1735, opened 2015-03-19
Publish a "cache" event on latestValue change

lcorneliussen:
For pure computeds the "change"-event does not actually fire on every change, but only for changes on active observables. (not on awake and peek) Also, the subscriptions to "change" lead to computation.

This is a necessary hook for #1511.

### DIFF
--- a/src/subscribables/dependentObservable.js
+++ b/src/subscribables/dependentObservable.js
@@ -150,6 +150,8 @@ ko.computed = ko.dependentObservable = function (evaluatorFunctionOrOptions, eva
 
                 _latestValue = newValue;
                 if (DEBUG) dependentObservable._latestValue = _latestValue;
+                
+                notify(_latestValue, "cache");
 
                 if (isSleeping) {
                     dependentObservable.updateVersion();


SteveSanderson:
I'm struggling to follow exactly what scenario this enables. Your description mentions it's a prerequisite for #1511, but that issue is closed already.

Could you clarify what problem this solves?

lcorneliussen:
What is the state here? Looks like I didn't follow up on your comment @SteveSanderson - but @craigcav seemed to confirm a "cache" event would solve the introduced breaking change from 3.0 to 3.1.

Just trying to update to 3.4 and saw that "cache" isn't included, so I need to patch it again.

lcorneliussen:
The idea is to observe any new computation results without causing any side effects.

mbest:
@lcorneliussen Without patching Knockout, you can get a "spectate" event as I [mentioned before](https://github.com/knockout/knockout/pull/1511#issuecomment-56913452). I hope that helps.

lcorneliussen:
IIRC: A peek to a spleeping does neither fire change or awake, but it still creates and caches a value.

We had a conversation there....
>Actually this is not yet solved. Tested a bit more...
>When I subscribe to a computed that doesn*t register dependencies, it won't awake and hence i won't get any notification about the new value.
>
>What about an "evaluated" event that just publishes all changes done to _latestValue?
>
>One might ask why I'd subscribe to something that won't ever change :dizzy_face:
>
>But if i want to spectate/observe a computed without triggering evaluation, I still need to get hold of the new value if it then is a) peeked, b) unwrapped, c) subscribed to - regardless of wether it registers dependencies or not.

mbest:
I've create a pull request for a `spectate` event (I like that better than `cache`). It includes the event for both computed and regular observables, plus has tests.

PR #1738, opened 2015-03-20
Deferred updates, part 1: task scheduler, extender, tests

mbest:
The additions here provide the means to efficiently make some actions asynchronous. The `ko.tasks` API includes the following methods for scheduling or canceling tasks:

`ko.tasks.schedule( callback )` - Schedules a task to run asynchronously; returns a task handle
`ko.tasks.cancel( handle )` - Cancels a scheduled task; does nothing if the task has already run

The following method are advanced and are useful for testing or for integrating with another task management system.

`ko.tasks.runTasks( )` - Immediately runs all scheduled tasks. 
`ko.tasks.scheduler( function )` - Overwrite this method to define a different function for scheduling when to run tasks.

The `deferred` extender uses the above task scheduler to set up deferred notification for observables.

### DIFF
--- a/spec/asyncBehaviors.js
+++ b/spec/asyncBehaviors.js
@@ -612,3 +612,181 @@ describe('Rate-limited', function() {
        });
     });
 });
+
+describe('Deferred', function() {
+    beforeEach(function() {
+        jasmine.Clock.useMock();
+    });
+
+    describe('Observable', function() {
+        it('Should delay notifications', function() {
+            var observable = ko.observable().extend({deferred:true});
+            var notifySpy = jasmine.createSpy('notifySpy');
+            observable.subscribe(notifySpy);
+
+            observable('A');
+            expect(notifySpy).not.toHaveBeenCalled();
+
+            jasmine.Clock.tick(1);
+            expect(notifySpy.argsForCall).toEqual([ ['A'] ]);
+        });
+
+        it('Should notify subscribers about only latest value', function() {
+            var observable = ko.observable().extend({notify:'always', deferred:true});  // include notify:'always' to ensure notifications weren't suppressed by some other means
+            var notifySpy = jasmine.createSpy('notifySpy');
+            observable.subscribe(notifySpy);
+
+            observable('A');
+            observable('B');
+
+            jasmine.Clock.tick(1);
+            expect(notifySpy.argsForCall).toEqual([ ['B'] ]);
+        });
+
+        it('Set to false initially, should maintain synchronous notification', function() {
+            var observable = ko.observable().extend({deferred:false});
+            var notifySpy = jasmine.createSpy('notifySpy');
+            observable.subscribe(notifySpy);
+
+            observable('A');
+            expect(notifySpy.argsForCall).toEqual([ ['A'] ]);
+        });
+
+        it('Should suppress notification when value is changed/reverted', function() {
+            var observable = ko.observable('original').extend({deferred:true});
+            var notifySpy = jasmine.createSpy('notifySpy');
+            observable.subscribe(notifySpy);
+
+            observable('new');
+            expect(observable()).toEqual('new');
+            observable('original');
+
+            jasmine.Clock.tick(1);
+            expect(notifySpy).not.toHaveBeenCalled();
+            expect(observable()).toEqual('original');
+        });
+
+        it('Set to false, should turn off deferred notification if already turned on', function() {
+            var observable = ko.observable().extend({deferred:true});
+            var notifySpy = jasmine.createSpy('notifySpy');
+            observable.subscribe(notifySpy);
+
+            // First, notifications are deferred
+            observable('A');
+            expect(notifySpy).not.toHaveBeenCalled();
+            jasmine.Clock.tick(1);
+            expect(notifySpy.argsForCall).toEqual([ ['A'] ]);
+
+            notifySpy.reset();
+            observable.extend({deferred:false});
+
+            // Now, they are synchronous
+            observable('B');
+            expect(notifySpy.argsForCall).toEqual([ ['B'] ]);
+        });
+    });
+
+    describe('Observable Array change tracking', function() {
+        it('Should provide correct changelist when multiple updates are merged into one notification', function() {
+            var myArray = ko.observableArray(['Alpha', 'Beta']).extend({deferred:true}),
+                changelist;
+
+            myArray.subscribe(function(changes) {
+                changelist = changes;
+            }, null, 'arrayChange');
+
+            myArray.push('Gamma');
+            myArray.push('Delta');
+            jasmine.Clock.tick(1);
+            expect(changelist).toEqual([
+                { status : 'added', value : 'Gamma', index : 2 },
+                { status : 'added', value : 'Delta', index : 3 }
+            ]);
+
+            changelist = undefined;
+            myArray.shift();
+            myArray.shift();
+            jasmine.Clock.tick(1);
+            expect(changelist).toEqual([
+                { status : 'deleted', value : 'Alpha', index : 0 },
+                { status : 'deleted', value : 'Beta', index : 1 }
+            ]);
+
+            changelist = undefined;
+            myArray.push('Epsilon');
+            myArray.pop();
+            jasmine.Clock.tick(1);
+            expect(changelist).toEqualOneOf([[], undefined]);
+        });
+    });
+
+    describe('Computed Observable', function() {
+        it('Should defer notification of changes and minimize evaluation', function () {
+            var timesEvaluated = 0,
+                data = ko.observable('A'),
+                computed = ko.computed(function () { ++timesEvaluated; return data(); }).extend({deferred:true}),
+                notifySpy = jasmine.createSpy('notifySpy'),
+                subscription = computed.subscribe(notifySpy);
+
+            expect(computed()).toEqual('A');
+            expect(timesEvaluated).toEqual(1);
+            jasmine.Clock.tick(1);
+            expect(notifySpy).not.toHaveBeenCalled();
+
+            data('B');
+            expect(timesEvaluated).toEqual(1);  // not immediately evaluated
+            expect(computed()).toEqual('B');
+            expect(timesEvaluated).toEqual(2);
+            expect(notifySpy).not.toHaveBeenCalled();
+            expect(notifySpy).not.toHaveBeenCalled();
+
+            jasmine.Clock.tick(1);
+            expect(notifySpy.calls.length).toEqual(1);
+            expect(notifySpy.argsForCall).toEqual([ ['B'] ]);
+        });
+
+        it('Should notify first change of computed with deferEvaluation if value is changed to undefined', function () {
+            var data = ko.observable('A'),
+                computed = ko.computed(data, null, {deferEvaluation: true}).extend({deferred:true}),
+                notifySpy = jasmine.createSpy('notifySpy'),
+                subscription = computed.subscribe(notifySpy);
+
+            expect(computed()).toEqual('A');
+
+            data(undefined);
+            expect(computed()).toEqual(undefined);
+            expect(notifySpy).not.toHaveBeenCalled();
+
+            jasmine.Clock.tick(1);
+            expect(notifySpy.calls.length).toEqual(1);
+            expect(notifySpy.argsForCall).toEqual([ [undefined] ]);
+        });
+
+        it('Should notify first change to pure computed after awakening if value changed to last notified value', function() {
+            var data = ko.observable('A'),
+                computed = ko.pureComputed(data).extend({deferred:true}),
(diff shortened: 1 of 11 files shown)


brianmhunt:
The only question I have about this is how it relates to ES6 Promises, specifically whether Promises could integrate with or substitute for this.

Having reviewed the code I feel like this PR is frying a different kettle of fish than Promises, so to speak, but Michael I just wanted to comment in case I am wrong – since Promises seem to be becoming the *de facto* asynchronous built-in behaviour (and they are, IMO, an excellent design), it would be good if they played nice together.

This is just Friday-night food for thought. :grin:

mbest:
> The only question I have about this is how it relates to ES6 Promises, specifically whether Promises could integrate with or substitute for this.

Promises are meant to help you deal with the results of asynchronous operations, whereas `ko.tasks` is meant to help make things asynchronous.

mbest:
Promises can also be used in a sort of hackish way to [schedule an asynchronous operation](http://stackoverflow.com/questions/27647742/promise-resolve-then-vs-setimmediate-vs-nexttick), using `Promise.resolve().then()`. As such, any Promise implementation must include or use a task scheduler.

mbest:
I mentioned before that `ko.tasks` could be useful for components as well. I put together a version that includes that change and then ran the example from #1504 against it.

Before: http://jsfiddle.net/mbest/xwy5nydm/3/embedded/result/
After: http://jsfiddle.net/mbest/xwy5nydm/2/embedded/result/

mbest:
Since it's a simple enough change, I've included the component change here as well.

@brianmhunt - Did I answer your question about the relation to Promises? Do you have any other comments regarding the API or implementation?

src/tasks.js
```diff
@@ -0,0 +1,117 @@
...
+        handleOrigin = 0,
+        processingTask,
+        observer;
+
+    if (observer = window['MutationObserver'] || window['WebKitMutationObserver']) {
+        // Chrome 18+, Firefox 14+, IE 11+, Opera 15+, Safari 6+; borrowed from https://github.com/petkaantonov/bluebird
+        scheduler = (function (callback) {
+            var div = document.createElement("div");
+            new observer(callback).observe(div, {attributes: true});
+            return function () { div.classList.toggle("foo"); };
```
brianmhunt: `classList` is not supported by Opera Mini 8 per http://caniuse.com/#feat=classlist

src/tasks.js
```diff
@@ -0,0 +1,117 @@
...
+                script.onreadystatechange = null;
+                document.documentElement.removeChild(script);
+                script = null;
+                callback();
+            };
+            document.documentElement.appendChild(script);
+        };
+    } else {
+        scheduler = function (callback) {
+            setTimeout(callback, 0);
```
brianmhunt: Arguably, since KO is a user interface library, requestAnimationFrame [is better](https://github.com/knockout/knockout/pull/1659#issuecomment-73881796).

I'm sure though for testing it is better to use `setTimeout`, so it may make sense to make this mockable/overloadable.

brianmhunt: It may make sense to have a generic `ko.utils.runAsync(fn)` that uses animation frames by default, falls back to `setTimeout`, and is overloadable for testing. This would be used elsewhere, too.

brianmhunt:
Thanks Michael. I added a couple of comments to the code, but they're not really important - more asides. Your explanation on Promises was good. I think it's good to include the component change, too.

My one concern is when there are a bunch of asynchronous events e.g.

```javascript
ko.tasks.schedule(vFn)
ko.tasks.schedule(wFn)
ko.tasks.runTasks()
```

Is there a way to tell when `vFn` and `wFn` (arbitrary function names) have completed? Is this desirable?

In `thenable` lingo (which is I think what the desirable Promises-compatibility would end up being), one would capture the events like this:

```javascript
ko.tasks.schedule(vFn)
ko.tasks.schedule(wFn)
ko.tasks.runTasks().then(function on_success() { }, function on_error() {})
```

i.e. `runTasks` returns a *thenable* function that calls either the `on_success` or `on_error` when complete. One can then convert this to a Promise, if desired, `Promise.resolve(ko.tasks.runTasks())`.

Is this reasonably doable and worthwhile? I am not sure on either point, honestly. Certainly a lot of the new APIs are handling asynchronicity this way (e.g. WebCrypto and `fetch`), and I think it *may* be handy to know when all the scheduled tasks are completed. Do you think so too?

src/tasks.js
mbest: It doesn't look like there are any browsers that support `MutationObserver` but don't support `classList`.

mbest:
> Is there a way to tell when `vFn` and `wFn` have completed? Is this desirable?

When using a Promise to get the result of the asynchronous action, the scheduled function would resolve the promise itself:

```javascript
var p1 = new Promise(function(resolve, reject) {       
    ko.tasks.schedule(function() {
        // Do something that needs to be asynchronous
        resolve(/* The result of our action */);
    });
});
```

> `runTasks` returns a *thenable* function that calls either the `on_success` or `on_error` when complete.

`ko.tasks.runTasks` is meant only for testing and for advanced use, such as integration with another task system. It runs whatever tasks have been scheduled immediately (i.e. synchronously). It's not really even necessary since we test using a mock scheduler. If you think `runTasks` is confusing, we could probably just take it out.

src/tasks.js
brianmhunt: Right you are - Looks good, thanks.

brianmhunt:
I could be convinced that we should make an effort to simplify the scheduled function by making it thenable so  `Promise.resolve(ko.tasks.schedule(vFn))` would work, but I think it's an unnecessary stretch. The way you wrap the task in a Promise is the correct/canonical way, I think.

Without a good use case, and without being able to reap async tasks, my sense is that `runTasks` might be confusing (people feeling the need to use it, even though it's not necessary), and misused (people expecting it to finish or callback when tasks, even async ones, complete). I lean towards pulling it, but am open to it being left in. :)

I hope this has been helpful feedback. :)

mbest:
@brianmhunt - Thanks for the additional feedback. I've removed `runTasks`, which led me to make some minor simplifications to the code.

mbest:
Regarding the async scheduler, I decide to test the various methods to compare their effectiveness at reducing layouts and paints. My test updates one observable synchronously and another asynchronously using the selected method. Both toggle a section of the page, using the `if` binding.

*setTimeout* - Two layouts, two paints.
*postMessge* - Two layouts, two paints.
*requestAnimationFrame* - Varies. In Chrome, one paint, but sometimes two layouts. In IE, two layouts and two paints.
*MutationObserver* - Always one layout and one paint.
*onreadystatechange* - (IE only) Always one layout and one paint.

brianmhunt:
Thanks @mbest. Great feedback on the testing.

Doing a little reading on [`MutationObserver` in bluebird](https://github.com/petkaantonov/bluebird/issues/22) and it sounds like `MutationObserver` can starve the rendering process. I don't understand this issue well enough to know if it is a real concern. Even if it is a concern, there are hybrid solutions referred to, such as mutation observer that periodically throws a `setTimeout` instead.

The real benefit of the animation frame would seem to be the power saving for unfocused tabs, which is at odds with performance.

If the aim is to maximize performance, then the task scheduler should probably be a `MutationObserver`, subject to problems noted above with rendering starvation. If it is to bulk animations into power-conscious blocks, then `requestAnimationFrame` would seem to be preferable. I am not sure the other solutions present any clear benefit over these choices, do they? I feel like, at this level, the aim is to maximize performance.

Incidentally, because an error in one task should not derail the lot, I believe it might be valuable if the [`task()`](https://github.com/knockout/knockout/pull/1738/files#diff-60456e6ecd00dc384b528b655db9a89cR46) call were isolated in a `try/catch`. Ideally it should leave open the possibility of a call to a KO error handler e.g. #1715.

mbest:
Ok, I've merged this now.

> Incidentally, because an error in one task should not derail the lot, I believe it might be valuable if the task() call were isolated in a try/catch.

We can revisit this when working on #1715. For now, an exception in a task will interrupt task processing, but it won't prevent the remaining tasks from running. This is handled here: https://github.com/knockout/knockout/blob/master/src/tasks.js#L52

brianmhunt:
> ... but it won't prevent the remaining tasks from running.

Oh right - I missed that. Looks great. Thanks, Michael.

rniemeyer:
Good stuff @mbest !

spec/asyncBehaviors.js
```diff
@@ -612,3 +612,180 @@ describe('Rate-limited', function() {
...
+            expect(changelist).toEqual([
+                { status : 'deleted', value : 'Alpha', index : 0 },
+                { status : 'deleted', value : 'Beta', index : 1 }
+            ]);
+
+            changelist = undefined;
+            myArray.push('Epsilon');
+            myArray.pop();
+            jasmine.Clock.tick(1);
+            expect(changelist).toEqualOneOf([[], undefined]);
```
SteveSanderson: Why are both acceptable? Are we inconsistent across browsers?

spec/taskBehaviors.js
```diff
@@ -0,0 +1,267 @@
...
+        };
+
+        ko.tasks.schedule(func.bind(null, 1));
+        ko.tasks.schedule(func.bind(null, 2));
+
+        jasmine.Clock.tick(1);
+        expect(runValues).toEqual([1,2]);
+    });
+
+    it('Should run tasks again if scheduled after a previous run', function() {
```
SteveSanderson: I'm not sure this spec covers what its name seems to suggest, unless I'm misinterpreting.

The name suggests (to me) that it covers what happens when you schedule the same function instance more than once, but given that it uses fn.bind to construct the param for ko.tasks.schedule, it's a different function instance getting scheduled each time.

Should this be changed to pass the same function instance each time?

spec/taskBehaviors.js
```diff
@@ -0,0 +1,267 @@
...
+
+            // Try to cancel the first task
+            ko.tasks.cancel(handle1);
+
+            // But nothing should happen; the second task will run in the next iteration
+            jasmine.Clock.tick(1);
+            expect(runValues).toEqual([1,2]);
+        });
+
+        it('Should work correctly after a task throws an exception', function() {
```
SteveSanderson: I'm not confident that this behavior is what we'd ideally want. This test shows that it can cancel a task asynchronously, but *only* if one of the preceding but independently scheduled tasks throws an exception. If someone has behavior that relies on cancellation not occurring because it is attempted asynchronously, they will be surprised if sometimes cancellation does succeed because of some completely unrelated task throwing.

Overall I find the whole idea of task callbacks getting delayed by completely unrelated exceptions to be a bit concerning (e.g., as also specified on line 97 above), as it's unpredictable action-at-a-distance. Is this behaviour essential for any reason, or could we change it so that the rest of the queue continues to be processed synchronously even if one entry throws?

src/tasks.js
```diff
@@ -0,0 +1,103 @@
+ko.tasks = (function () {
```
SteveSanderson: This file includes snippets from other MIT-licensed projects, which is fine, but what do you think we should do about the licensing comments? The MIT license says you need to include a copy of the license if you use a "substantial portion" of the product, but never clarifies what "substantial" means or who adjudicates that. Other people who've discussed this seem to be risk-averse and suggest including the third-party license if you use any of their code at all.

Can we at least copy and past in the licenses for the two relevant projects as block comments here, inline in this file? Better still if they survive the minification process while remaining inline here in the code and not having to move to the top.

spec/asyncBehaviors.js
vamp: I think that undefined should be removed

mbest: I copied this from my plugin, where it think it was meant to pass with different versions of Knockout. The correct result should be `undefined`.

spec/taskBehaviors.js
mbest: Again, I had copied this from my plugin, which has the ability to provide arguments separately. The argument isn't necessary for the test, so I'll change it to use the same function instance.

spec/taskBehaviors.js
mbest: Normally, it wouldn't be possible to cancel a task in this manner since the exception would be uncaught (we can catch it because we're using the mock scheduler). I could modify the test to work regardless by putting the `cancel` call within a previously scheduled task, but this may not matter if we change how exceptions are handled.

src/tasks.js
mbest: I wouldn't call either of these "substantial", and the code is not copied exactly (I've made minor changes). Still, I think it makes sense to include more detail about the license. So how is this?

```
// From https://github.com/petkaantonov/bluebird * Copyright (c) 2014 Petka Antonov * License: MIT
```

mbest: Also, if we want the license statement in the minified file, it will need to be a string instead of a comment.

PR #1746, opened 2015-03-31
External templates with default templating engine

jtulach:
I was searching for a way to define template via
```html
<script type="text/html" id="templateId">
template text
</script>
```
but with the ''template text'' being in a separate file. I have not found the exact solution, the closest one being [this tutorial](http://www.knockmeout.net/2011/10/ko-13-preview-part-3-template-sources.html). Based on it I figured out I can redefine behavior of ko.templateSources.domElement.prototype['text'] to do what I need. I think the solution could be generally useful and I am proposing to include it in the nativeTemplatingEngine itself.

With my change one can define a template via 
```html
<script src="your.template.html" type="text/html"></script> 
```
and once the template is needed the nativeTemplatingEngine will asynchronously load it from given source (in this case your.template.html).

If you find the overall direction of my pull request sane, let me know what else should be polished, so the request is mergeable.

### DIFF
--- a/src/templating/templateSources.js
+++ b/src/templating/templateSources.js
@@ -37,8 +37,24 @@
                                  : tagNameLower === "textarea" ? "value"
                                  : "innerHTML";
 
-        if (arguments.length == 0) {
-            return this.domElement[elemContentsProperty];
+       if (arguments.length == 0) {
+            var el = this.domElement;
+            var val = el[elemContentsProperty];
+            if (tagNameLower === "script" && !val && el['src']) {
+                val = ko.observable("Loading...<br/>");
+                var xhr = new XMLHttpRequest();
+                xhr.open('GET', this.domElement['src'], true);
+                xhr.setRequestHeader('Content-Type', 'text/html; charset=utf-8');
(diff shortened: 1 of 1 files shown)


vamp:
Sorry, but my personal opinion - this "feature" is not needed in knockout core (you can override/extend template source, create you custom binding around template or use component).

SteveSanderson:
This functionality (asynchronously fetching external templates) is covered by [components](http://knockoutjs.com/documentation/component-overview.html). If you're keen to do it without components, other third party plugins exist to do that (search for "knockout external templates").

Since we're emphasising components for this in core, I don't think it's something we'd want to change in the template system for the forseeable future. Hope that's OK!

jtulach:
I see. There is the "require: 'text!files/component-like-widget.html'" use-case in components. I understand you want to emphasise use of components, but shouldn't knockout (for the sake of consistency) support the same format for templates as well?

SteveSanderson:
It's not really a matter of 'using the same format'. Templates themselves are more low-level infrastructure. The business of fetching external templates asynchronously is a higher-level concern, which is rightly the job of the components system.

Issue #1748, opened 2015-03-31
Documentation for ko.computedObservable


wallymathieu:
The [documentation](http://knockoutjs.com/documentation/computedObservables.html) for computed obserable does not mention that computed is not needed in many cases. Since it looks like computed is used internally when binding expressions and properties to a view.

```
this.fullName = function() {
    return this.firstName() + " " + this.lastName();
};
```
using:

```
The name is <span data-bind="text: fullName()"></span>
```

Also there are some issues with using computed willy nilly:
[knockout-cleaning-up](http://www.knockmeout.net/2014/10/knockout-cleaning-up.html)

Perhaps some additional documentation on the page computed observable section?

waldheinz:
When doing it the suggested way, the `fullName` won't be updated when either `firstName` or `lastName` change, which in my understanding the main selling point for using `computed`s. But I admit there may be "fire and forget" scenarios where this can be a valid thing to do. Albeit it will be a pain to debug when the requirements change.

wallymathieu:
Well, it looks like it is updating in : http://jsfiddle.net/ozzymcduff/t1cafytg/

waldheinz:
Oh wow, ok, so the magic happens in the `text:` part of `data-bind='text: fullName()'`. Good to know!

wallymathieu:
It's nothing special for the text binding. You can get the same behavior for your custom bindings. Try out this example by making the name longer than 10 characters: http://jsfiddle.net/ozzymcduff/k9zk0721/

SteveSanderson:
Thanks for the suggestion. I haven't found many people to be confused by the existing docs, so I don't think it necessary to prioritise changing this. If you do feel strongly please consider submitting a pull request on those docs - in the meantime I'll close this to tidy up. Thanks the the feedback though!

wallymathieu:
I don't feel strongly about it. I have met a lot of confusion about it
though. People starting to program knockout that start by thinking that
they need to use computed.

PR #1750, opened 2015-04-06
When computed obserable is only used in UI

wallymathieu:
In order for people to be less confused about when to use computed the documentation could provide an alternative in the end of computed obserable page. This is related to issue #1748


SteveSanderson:
Thanks - merged!

I made a slight change to the phrasing to avoid implying that all other computeds have to be manually disposed (they don't, unless they reference something that lives longer than them, which would not be the case in this example).

wallymathieu:
Sweet! Thanks!

PR #1751, opened 2015-04-06
1715 async error capture callback

SteveSanderson:
Here's a ready-to-go version of the pull request previously at #1715.

Discussion about this can continue in #1715. There's still a decision to be made about this API.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -335,14 +335,31 @@ ko.utils = (function () {
             return element && element.tagName && element.tagName.toLowerCase();
         },
 
+        catchFunctionErrors: function (delegate) {
+            return ko.onError ? function () {
+                try {
+                    return delegate.apply(this, arguments);
+                } catch (e) {
+                    ko.onError && ko.onError(e);
+                    throw e;
+                }
+            } : delegate;
+        },
+
+        setTimeout: function (handler, timeout) {
+            return setTimeout(ko.utils.catchFunctionErrors(handler), timeout);
+        },
+
         registerEventHandler: function (element, eventType, handler) {
+            var wrappedHandler = ko.utils.catchFunctionErrors(handler);
+
             var mustUseAttachEvent = ieVersion && eventsThatMustBeRegisteredUsingAttachEvent[eventType];
             if (!mustUseAttachEvent && jQueryInstance) {
-                jQueryInstance(element)['bind'](eventType, handler);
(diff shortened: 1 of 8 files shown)


mbest:
I wonder if for `ko.tasks`, we could wrap the function that processes tasks instead of each task.

src/utils.js
```diff
@@ -335,14 +335,31 @@ ko.utils = (function () {
...
+                    return delegate.apply(this, arguments);
+                } catch (e) {
+                    ko['onError'] && ko['onError'](e);
+                    throw e;
+                }
+            } : delegate;
+        },
+
+        setTimeout: function (handler, timeout) {
+            return setTimeout(ko.utils.catchFunctionErrors(handler), timeout);
```
brianmhunt: Would this be better as a mutation observer, animation frame, deferral to, or refactor of `ko.tasks` [scheduler](https://github.com/knockout/knockout/blob/7d263dcbca18eba03cc7dca5f5f73d354aa77a31/src/tasks.js#L7)? That would feel more consistent – is there logic in keeping it separate?

Perhaps, if so, then this should not be `ko.utils.setTimeout` but rather `ko.utils.runAsync` (or something similar).

Purely thoughts & suggestions, of course. :)

mbest: All the instances of `setTimeout` in Knockout actually need some sort of delay and thus can't use `ko.tasks.schedule`.

brianmhunt: Got it, thanks @mbest.

I wonder about `animationFrame` for improved performance/battery life (as [discussed elsewhere](https://github.com/knockout/knockout/pull/1659#issuecomment-73881796))?

mbest: It's possible that `requestAnimationFrame` could work in some cases, but we'd need to test each case.

SteveSanderson:
Merged.

My understanding of the feedback above is that the conclusion is the existing implementation is fine. If I'm misreading and we should change something, please let me know.

brianmhunt:
Wonderful– Glad to see this make it in, thanks @SteveSanderson .

Thanks @mbest for the comment on `requestAnimationFrame` – if it could be a drop-in replacement for `setTimeout` without too much fenangling I think it's probably worthwhile, otherwise I'd venture that it's pretty far down the list of priorities unless mobile/battery devices are a target.

Issue #1728, opened 2015-03-13
Include deferred updates in Knockout core


mbest:
My Deferred Updates plugin has been quite popular as a means of improving the performance of Knockout applications. But it's quite a hassle to keep it updated with each new version of Knockout, while maintaining backwards compatibility with previous versions. This is mainly because it replaces some of Knockout's internal components such as `ko.computed`.

I think the framework is already in place within Knockout now, with the rate-limit and version features, to make deferred updates a very simple addition.

rniemeyer:
I am all for it. Sounds like a great idea.

reflectiveSingleton:
+1 for this

brianmhunt:
Great suggestion @mbest; I like the idea of this functionality being in core, though I am mindful enough of SteveSanderson's [comment](https://github.com/knockout/knockout/pull/1543#issuecomment-70330822) to note it here:

> We have so much functionality around computeds by now that the logic in `dependentObservable.js` is really intricate... I think we've done a great job of keeping it robust by having such extensive test coverage, but it's absolutely at the limit of what I think is practical to maintain.

Which is not to deter merging this – only that we should be mindful of the possibility that we are travelling down the path of a monolithic design pattern, that may prove a challenge to maintain and evolve.

One option may include eventually decoupling the observable system from the binding system.

:+1:

mbest:
@brianmhunt You are right that we need to be cautious, but my hope is that the changes to `dependentObservable.js` will be very minimal.

I'd like to use this space to do a little brainstorming about this feature. I think a good way to start is to go through parts of the plugin and examine if they should be ported to Knockout.

* In the plugin, all pending deferred updates are added to a queue and handled by a single scheduled processing function. In other words, it includes a task manager. If we were running only in Node.js, `process.nextTick` might be a sufficient replacement, but there isn't any equivalent facility in browsers. So I think this will be a necessary feature, which could also prove useful for other areas of Knockout, such as components (see #1659).

* The plugin allows you to turn deferred updates on and off. The feature is on by default, but it can be later turned off, either globally or for a specific computed observable. We'll definitely need some sort of API to enable deferred updates, and using `ko.computed.deferUpdates = true;` might be a good enough method. But I'm more inclined towards a one-way action that requires calling a function, such as `ko.computed.enableDeferredUpdates();`

* The plugin defers all `change` subscriptions. `rateLimit` also does this by delaying the `change` notification, so this should just work naturally. But it does mean that all observables will need to be "deferred".

[...]

Issue #1603, opened 2014-11-07
Custom Element name must be entered as lowercase string


kthejoker:
If you register your component with a string containing non lowercase alphas, it fails to render if you use it as a custom element directly.

//works
ko.components.register('randomthing', { require: 'mvvm/components/randomthing' } );

//doesn't work (doesn't fail, just elements don't render at all)
ko.components.register('randomThing', { require: 'mvvm/components/randomthing' } );

Note this is irrespective of whether the HTML itself has mixed case, uppercase, etc.

UPDATE: I see this now in the documentation, but it reads a bit too soft. Something like "You MUST use a lowercase string to use the component as a custom element name."

mbest:
Similar issue: #1510

reflectiveSingleton:
+1 for the PEBKAC notification proposed by @brianmhunt

SteveSanderson:
I'm not certainly what the best way to log a warning would be, because it's intentionally legal to use other non-lowercase component names (as long as you don't try to use them as custom elements). Logging a warning would be a breaking change (some people have their unit tests set up to fail if there are any errors or warnings logged, such as my current team), and a very annoying one for anyone who was doing this deliberately.

brianmhunt:
Perhaps the warning on non-lowercase component names could be conditional? e.g. "ko.settings.strictComponentNames = true` (I'm making up `ko.settings` as some theoretical box of options).

kthejoker:
Or rather than console.warn could we just use console.info if a custom
component is generated and no element is found? Or as Brian alluded to,
some sort of "use strict" equivalent setting to force the issue to the
surface?

The problem is the error is hard to spot and since it requires rigidity
where HTML spec does not, the documentation should be much closer to the
code than where it currently sits.

brianmhunt:
Another option is for KO to register all component names twice – once with the original case sensitive name it is given, another the lowercase version (if it differs from the original). So `randomThing` becomes a component accessible as `randomThing` and `randomthing`.

Obviously this can result in name collisions (and is API-breaking as a result), but it is in practice what I had actually been doing anyway so I think it may be a of a convention/habit of some folks. I expect most folks would be conscious of components being distinguished only by case.

rocketmonkeys:
[...]

PR #1753, opened 2015-04-08
Deferred updates, part 2: default option

mbest:
After reading @brianmhunt's [comment](https://github.com/knockout/knockout/issues/1603#issuecomment-90278568) on #1603, I thought of using the same API for controlling whether to use global deferred updates. But I like `ko.options` instead of `ko.settings`, so we have `ko.options.deferUpdates`.

Following the information I presented in #1728, this also includes a method to run pending tasks immediately. Instead of `ko.tasks.runTasks`, the name I had used in #1738, I decided on `ko.tasks.runEarly`, which I think makes its purpose more clear.

### DIFF
--- a/spec/taskBehaviors.js
+++ b/spec/taskBehaviors.js
@@ -5,7 +5,8 @@ describe('Tasks', function() {
 
     afterEach(function() {
         // Check that task schedule is clear after each test
-        expect(ko.tasks.length()).toEqual(0);
+        expect(ko.tasks.reset()).toEqual(0);
+        jasmine.Clock.reset();
     });
 
     it('Should run in next execution cycle', function() {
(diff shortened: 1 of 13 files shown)


brianmhunt:
I like both `ko.tasks.runEarly` and `ko.options`.

IanYates:
:+1: - The new naming makes the intent more clear

spec/taskBehaviors.js
```diff
@@ -205,6 +206,74 @@ describe('Tasks', function() {
...
+                // Schedule another task; it will be run after this one completes
+                ko.tasks.schedule(func.bind(null, 3));
+            });
+            ko.tasks.schedule(func.bind(null, 1));
+
+            jasmine.Clock.tick(1);
+            expect(runValues).toEqual([1,2,3]);
+        });
+
+        it('Should keep correct state if a task throws an exception', function() {
```
SteveSanderson: I'm not sure that runEarly should throw if one of the handlers it runs throws. This means *everywhere* you use runEarly, you have to put all your code in try/catch, because you have no idea whether some completely unrelated task will throw. And people can use it as a way of detecting whether or not at least one scheduled handler does throw (without telling you *which* handler), which encourages excessive use of runEarly, undermining the benefit.

Could we change this so runEarly doesn't throw? Throwing doesn't seem to communicate anything useful, as you have no idea what threw (and I don't think we should add a way of reporting what threw, as it would lead to really weird use cases). Then all task exceptions are definitely going to land in the same place - the `onError` callback (or `window.onerror` if you catch at that level).

spec/taskBehaviors.js
```diff
@@ -205,6 +206,74 @@ describe('Tasks', function() {
...
+            var runValues = [];
+            var func = function(value) {
+                runValues.push(value);
+            };
+            ko.tasks.schedule(func.bind(null, 1));
+            ko.tasks.schedule(function() {
+                expect(runValues).toEqual([1]);
+                ko.tasks.runEarly();
+                ko.tasks.schedule(func.bind(null, 3));      // This never gets called
```
SteveSanderson: As per my comment above, I would find it less surprising and more useful if this *did* get called.

src/tasks.js
```diff
@@ -89,10 +87,14 @@ ko.tasks = (function () {
             }
         },
 
-        // For testing only: return the queue length
-        length: function () {
-            return taskQueueLength;
-        }
+        // For testing only: reset the queue and return the previous queue length
+        reset: function () {
```
SteveSanderson: Is 'reset' exported? Doesn't appear to be exported explicitly, but the fact that specs use it suggests its name doesn't get minified. If this is really only for testing, could we rename it to something like 'resetForTesting'? Otherwise, with such a important-sounding name, people will probably start using it and relying on implementation quirks.

mbest:
@SteveSanderson I'm glad you brought up exception handling since it was something that I felt a bit unsure about.

The main advantage of falling out of the task loop on an exception is that the error gets reported right away in the correct context. We already have somewhat of a precedence for this because if you update an observable, it can trigger any number of manual subscriptions, computed observables, and bindings to run, and errors will both cancel the remaining updates and get thrown back to the caller (the code updating the observable).

I'm also not too concerned about the remaining tasks being rescheduled after an exception, since if using `MutationObserver` or `onreadystatechange`, it still happens before the UI is redrawn.

But, I do agree with your concerns about errors when using `runEarly`. At least when using `runEarly`, we can handle the errors asynchronously. I've seen this done before with something like this:

```
try {
    //...
} catch (e) {
    setTimeout(function() { throw e; }, 0);
}
```

mbest:
After trying a couple a methods, I decided that the simplest and most clear way is to have all task errors deferred using `setTimeout`. So this should now work as you've suggested, @SteveSanderson.

mbest:
I'm merging this now so we can move on to the next part.

PR #1756, opened 2015-04-08
Component named "watch" will give an error in Firefox

rniemeyer:
Naming a component `watch` gives the following warning in Firefox 37:

`Error: Component watch is already registered`

I've created a gist of a barebones application which demonstrates this:
https://gist.github.com/msteward/c3a0298322277adae29e

This error is not thrown in Chrome or IE. For now I can choose another name but curious as to why this would fail in only Firefox?


rniemeyer:
Code added to this issue.

mbest:
I made a small addition to the test. Otherwise looks great.

rniemeyer:
Looks good @mbest

SteveSanderson:
Excellent - thanks!

PR #1760, opened 2015-04-09
Added cleanDescendants to match cleanNode and applyBindingsToDescendants

### LINKED ISSUE #1621: Add cleanDescendants
There is a `ko.cleanNode(node)` method for cleaning up after `ko.applyBindings(vm, node)`, but there is no equivalent to `ko.applyBindingsToDescendants(vm, node)`. I propose a `ko.cleanDescendants(node)` method that will only clean up the descendants of the node, not the node itself.

mariusGundersen:
This fixes #1621

### DIFF
--- a/src/utils.domNodeDisposal.js
+++ b/src/utils.domNodeDisposal.js
@@ -90,7 +90,7 @@ ko.utils.domNodeDisposal = new (function () {
                 node.parentNode.removeChild(node);
         },
 
-        cleanExternalData : function (node) {
+        "cleanExternalData" : function (node) {
             // Special support for jQuery here because it's so commonly used.
             // Many jQuery plugins (including jquery.tmpl) store data using jQuery's equivalent of domData
             // so notify it to tear down any resources associated with the node & descendants here.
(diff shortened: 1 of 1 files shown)


src/utils.domNodeDisposal.js
```diff
@@ -85,7 +90,7 @@ ko.utils.domNodeDisposal = new (function () {
                 node.parentNode.removeChild(node);
         },
 
-        "cleanExternalData" : function (node) {
+        cleanExternalData : function (node) {
```
SteveSanderson: Does this un-export `cleanExternalData`, and is that deliberate? I thought this was an intentional extensibility point.

mariusGundersen: I just found it weird that this name was quoted when none of the others were. Should it be in the ko.exportSymbol list at the bottom of the file?

mariusGundersen:
Reverted the change that probably had unintended consequenses

mbest:
I've merged a version of this change now.

PR #1761, opened 2015-04-09
Improve the propagation of change notifications

dancras:
This PR solves two relatively high impact issues (for me at least):

 - Subscribers being notified many times following a single change to an observable.
 - Subscribers being notified with intermediate states which exist while the changes to computeds are propagating.

Currently I've found I'm abusing rateLimit: 0, adding superfluous conditionals to manage the intermediate states, and spending too much time planning dependency graphs (resulting in some very fragile code).

This solution is the result of a recent experiment of mine which can be [found here](https://github.com/dancras/reactiveComparison).

It works by queuing notifications to dependencies, and intercepting reads of computeds in order to evaluate the dependency graph breadth-first rather than depth-first.

It's fully unit tested, and it affected some existing tests because they are now performing fewer evaluations and skipping intermediate states. I had 63 failures running the browser test runner both before and after my changes; perhaps I'm missing something there?

I've also tried it out on the code [here](http://casumo.com) and everything seems to be working nicely.

Some possible future additions include:

 - Add a ko.bufferedUpdate function for updating multiple observables with only one re-evaluation of the dependency graph
 - Add a ko.sideEffects function so that code which performs side effects outside of the dependency graph can be performed last. This in combination with something like [fastdom](https://github.com/wilsonpage/fastdom) could really improve performance in the future.

Curious to hear your thoughts.

### DIFF
--- a/build/fragments/source-references.js
+++ b/build/fragments/source-references.js
@@ -8,6 +8,7 @@ knockoutDebugCallback([
     'src/utils.domManipulation.js',
     'src/memoization.js',
     'src/tasks.js',
+    'src/subscribables/notifyQueue.js',
     'src/subscribables/extenders.js',
     'src/subscribables/subscribable.js',
     'src/subscribables/dependencyDetection.js',
--- a/spec/asyncBehaviors.js
+++ b/spec/asyncBehaviors.js
@@ -418,30 +418,26 @@ describe('Rate-limited', function() {
         });
 
         it('Should delay change notifications and evaluation', function() {
+
             var observable = ko.observable();
             var evalSpy = jasmine.createSpy('evalSpy');
             var computed = ko.computed(function () { evalSpy(observable()); return observable(); }).extend({rateLimit:500});
             var notifySpy = jasmine.createSpy('notifySpy');
             computed.subscribe(notifySpy);
-            var beforeChangeSpy = jasmine.createSpy('beforeChangeSpy')
-                .andCallFake(function(value) {expect(computed()).toBe(value); });
+            var beforeChangeSpy = jasmine.createSpy('beforeChangeSpy');
             computed.subscribe(beforeChangeSpy, null, 'beforeChange');
 
             // Observable is changed, but notification is delayed
             evalSpy.reset();
             observable('a');
             expect(evalSpy).not.toHaveBeenCalled();
-            expect(computed()).toEqual('a');
-            expect(evalSpy).toHaveBeenCalledWith('a');      // evaluation happens when computed is accessed
-            expect(notifySpy).not.toHaveBeenCalled();       // but notification is still delayed
+            expect(computed()).toEqual(undefined);
+            expect(evalSpy).not.toHaveBeenCalled();         // nothing changes until the rate limit timer is over
+            expect(notifySpy).not.toHaveBeenCalled();       // and no notification happens
             expect(beforeChangeSpy).toHaveBeenCalledWith(undefined);    // beforeChange notification happens right away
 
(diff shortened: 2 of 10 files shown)


AdamWillden:
For what it's worth I'm also abusing rateLimit: 0 due to this same issue. 

In fact I've got an issue that I can't resolve using rateLlimitLimit: 0. I'm intrigued to know if this would be a solution.

dancras:
Here's the tag I've been using to experiment with: https://github.com/dancras/knockout/tree/v3.4.1x

AdamWillden:
Thanks I'll take a look when I can! Are there scenarios you know of where performance is reduced?

dancras:
No, it should really be a performance improvement, although I think it would take a pretty large dependency graph to see any real gains.

tedsteen:
:+1:

SteveSanderson:
Hi @dancras - thanks for suggesting this!

Are you aware of the work already in progress on [deferred updates](https://github.com/knockout/knockout/issues/1728)? If I'm understanding your proposal, it sounds like these cover the same ground. Would you be able to clarify how your proposal differs from deferred updates? Perhaps your requirements will already be met by deferred updates.

mbest:
It's possible that deferred updates will also solve the problem described and solved here, but I'm not sure. The issue is that `ko.pureComputed` can reverse the order of subscriptions from what you might expect and what happens when using `ko.computed`. If we look at the first example from @dancras's test:

```
var a = ko.observable('a'),
    b = ko.pureComputed(function() {
        return a() + 'b';
    }),
    c = ko.pureComputed(function() {
        return a() + b() + 'c';
    }),
    spy = jasmine.createSpy('callback');

c.subscribe(spy);
```

When using `ko.computed`, the order of subscriptions is `b->a` and then `c->a`, `c->b`. So when `a` is updated, it notifies `b` first. The order will be `a, b, c, c`.

When using `ko.pureComputed`, `c` begins subscribing before `b`, so the order is `c->a`, `c->b`, `b->a`. The order of updates is `a, c, b, c`. The order of the observables in `c` makes a difference. If it `return b() + a() + 'c';`, then we would get updates in the order `a, b, c, c` like above.

mbest:
I decided to run the above tests using the deferred updates branch. The simple graph passes, but the complex graph still ends up with 2 notifications instead of 1. Analyzing the simple case, the deferred updates code will schedule the updates as `a, c, b, c`, but the second scheduling of `c` will clear out the first one, so we end up with `a, b, c`.

What I have planned for part 3 should address the issues with the complex graph, and I might just borrow that test for it.

dancras:
Thanks for the interest guys. I wasn't aware of the deferred updates branch. It certainly seems to have a similar impact from a glance. I need to spend some time looking into it to better appreciate the differences.

From what I understand so far one major difference is that deferred updates makes all change notifications asynchronous. Could this mean the browser performs additional renders while the changes propagate through a large graph?

Also if you have the following graph with deferred updates:

```
     B - D - E
A<                > F
          C
```

Would a change to A trigger two updates to F because the top path would take several more stack frames to reach F?

I noticed following comments from @mbest that the complex graph test was failing for computeds so I pushed a solution for that.

The next thing I'll be considering is how to optimise writes which happen while evaluating the graph.

mbest:
> Would a change to A trigger two updates to F because the top path would take several more stack frames to reach F?

In the current iteration, it would get updated twice, so that's something I plan to address in a "part 3".

dancras:
I think this is redundant with the release of 3.4.0. Its complexity got a little out of hand anyway.

PR #1773, opened 2015-04-15
Fixes #1772: simpleHtmlParse and components starting with thead, tbody, etc.

### LINKED ISSUE #1772: simpleHtmlParse() fails to parse components having names beginning with thead, tbody, tfoot, tr, th and td
If your custom component is named ```<traffic-map></traffic-map>```, it will generate invalid HTML that gets discarded by innerHTML.

This is also the case for other nested table elements.

birkestroem:
simpleHtmlParse() fails to parse components having names beginning with thead, tbody, tfoot, tr, th and td


brianmhunt:
This change seems innocuous yet necessary. :smile:

mbest:
I've merged this with changes. See 4c19469

PR #1774, opened 2015-04-16
Adding the preferJQueryEvents option as a way to avoid using jquery eventing in knockout.

dzearing:
Usage:
```html
<head>
    <script src="knockout.js"></script>
    <script type="text/javascript">
        ko.options.preferJQueryEvents = false;
    </script>
   <script src="jquery.js"></script>
</head>
```
// Now all events fired from knockout should not have jquery in the callstack. It is assumed the option is set before applyBindings is called.

### DIFF
--- a/spec/utilsBehaviors.js
+++ b/spec/utilsBehaviors.js
@@ -344,14 +344,14 @@ describe('Function.bind', function() {
 });
 
 describe('registerEventHandler', function() {
-    it ('should not use jQuery eventing with noJQueryEvents options', function() {
+    it ('should not use jQuery eventing with useJQueryEventing option set to false', function() {
         var jQueryLoaded = (typeof jQuery !== 'undefined');
         var element = document.createElement('DIV');
         var eventFired = false;
         var jQueryUsed = false;
 
         // Set the option to true.
-        ko.options.noJQueryEvents = true;
+        ko.options.useJQueryEventing = false;
 
         // If jQuery is present, verify jQuery is not used in event binding.
         if (jQueryLoaded) {
@@ -365,7 +365,7 @@ describe('registerEventHandler', function() {
         ko.utils.triggerEvent(element, 'click');
 
         // Reset the option.
-        ko.options.noJQueryEvents = false;
+        ko.options.useJQueryEventing = true;
 
         expect(!jQueryLoaded || (eventFired && !jQueryUsed)).toBe(true);
     });
--- a/spec/utilsBehaviors.js
+++ b/spec/utilsBehaviors.js
@@ -343,3 +343,31 @@ describe('Function.bind', function() {
     });
 });
 
+describe('registerEventHandler', function() {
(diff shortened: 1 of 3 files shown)


SteveSanderson:
Thanks for this!

Is it possible to add a test for this change? When we run the tests, we do so both with and without jQuery, so it would make sense to have a test that we don't get jQuery event behaviour in the presence of this flag.

mbest:
Should this option also affect `ko.utils.triggerEvent`?

dzearing:
I'll look into the tests. 

Good call @mbest. I will take a look.

dzearing:
Added a test to validate the option locks out jquery eventing.

Updated trigger.

I ran this with runner.html and runner.html?jquery=1, and validated it works in both cases. I'm not sure of a clever way to distinguish trigger behavior coming from jquery vs not.

dzearing:
ping!

stalniy:
It's better to use direct terminology. I'd like to see `useJqueryEvents = true/false` instead of `noJqueryEvents = true/false`.

At least it's hard to understand for my brain the meaning of `noJqueryEvents = false`

brianmhunt:
I agree with @stalniy – The double-negative is too much for my brain. :)

dzearing:
Renamed to "useJQueryEventing". Good call.

mbest:
When I was reading your comment, I at first thought it said "everything" instead of "eventing".

dzearing:
Happy to rename, just give me something everyone is satisified with, so we can finish this off. I can change to useJqueryEvents if it is more straightforward. In my opinion options are always going to be a little mysterious to use, even if named something completely innocent, so naming it useJQueryEvents vs useJqueryEventsWhenAvailable is still going to require a little digging to understand what that implies. It's requires a little digging just to even know that jquery is being referenced from knockout in the first place, I would have not guessed that.

brianmhunt:
@dzearing How does `preferJQueryEvents` ring with you?

dzearing:
Done, I like it.

dzearing:
Any other feedback?

SteveSanderson:
Looks great! And thanks for making the option name change - sounds ideal now.

Merged. I think the other core team members will have seen this by now (since @mbest and @brianmhunt have commented here) but if not and any of you have concerns please let us know.

rniemeyer:
I thought that this one seemed fine. I would generally prefer new options to match the default/current behavior when undefined/falsey, but did agree that the double-negative was confusing.

mbest:
If we'd rather have an option that defaults to `false`, how about `useOnlyNativeEvents`?

rniemeyer:
>  If we'd rather have an option that defaults to false, how about useOnlyNativeEvents?

+1

rniemeyer:
Note: option was updated to `useOnlyNativeEvents` here: 09b9e93f29a7ec7a5282fe9deb2a750d9e48f8f4

brianmhunt:
@rniemeyer I updated the title too, just in case someone wanders onto this :)

PR #1778, opened 2015-04-19
#1591 Test for checked with writeable computed

Roarster:
This is a failing test for checked binding backed to a writeable computed observable that is backed by an array. Currently writing to this will throw an error because the code assumes the writeable computed is an array and attempts to perform array operations on it.

I'm not sure the correct approach for fixing this.  To me it seems like you could update ```ko.utils.addOrRemoveItem``` to deal with the case where it's a writeable computed or have ```updateModel``` in the checked binding do something different if it's a writeable computed.  I'm happy to take a look at fixing this if someone can suggest the best approach.


Roarster:
Closing this since #1591 has now been merged.

PR #1781, opened 2015-04-30
Handle empty, non-blank string in default css binding

### LINKED ISSUE #1431: Bug: data-bind="css: ('   ')" crashes the application
In this case `ko.utils.arrayForEach` in `toggleDomNodeCssClass` gets null and fails with null reference exception.

npbee:
In the default CSS binding, trim the `value` before attempting to toggle the
DOM node CSS classes.  An empty, non-blank string is evaulated as truthy.  The
`toggleDomNodeCssClass` uses an `if (classNames) ...` condition to check for
the existence of a class name before trying to match and loop through them.
Since the non-empty string was being passed, it was attempting to use the
`arrayForEach` method on a `null` value.

Fixes #1431

### DIFF
--- a/src/binding/defaultBindings/css.js
+++ b/src/binding/defaultBindings/css.js
@@ -8,7 +8,7 @@ ko.bindingHandlers['css'] = {
                 ko.utils.toggleDomNodeCssClass(element, className, shouldHaveClass);
             });
         } else {
-            value = String(value || ''); // Make sure we don't try to store or set a non-string value
+            value = String(value || '').trim(); // Make sure we don't try to store or set a non-string value
             ko.utils.toggleDomNodeCssClass(element, element[classesWrittenByBindingKey], false);
             element[classesWrittenByBindingKey] = value;
             ko.utils.toggleDomNodeCssClass(element, value, true);
--- a/src/binding/defaultBindings/css.js
+++ b/src/binding/defaultBindings/css.js
@@ -8,7 +8,7 @@ ko.bindingHandlers['css'] = {
                 ko.utils.toggleDomNodeCssClass(element, className, shouldHaveClass);
             });
         } else {
-            value = String(ko.utils.stringTrim(value) || ''); // Make sure we don't try to store or set a non-string value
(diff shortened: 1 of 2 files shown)


brianmhunt:
Thanks @npbee 

Not all browsers supported by Knockout have `String.prototype.trim`. What you want is `ko.utils.stringTrim`.

Also, if we are going to merge this it would need a unit test, if you please.

Cheers

npbee:
Ah, yes thanks!  I'll make sure to keep a closer eye on compatibility for any future PR's.  I've made the change.

src/binding/defaultBindings/css.js
```diff
@@ -8,7 +8,7 @@ ko.bindingHandlers['css'] = {
                 ko.utils.toggleDomNodeCssClass(element, className, shouldHaveClass);
             });
         } else {
-            value = String(value || ''); // Make sure we don't try to store or set a non-string value
+            value = String(ko.utils.stringTrim(value) || ''); // Make sure we don't try to store or set a non-string value
```
brianmhunt: Not a big issue, but I think the `ko.utils.stringTrim` should probably be outside the `String(value || '')` i.e. `ko.utils.stringTrim(String(value) || '')`. The `stringTrim` function [looks pretty safe](https://github.com/knockout/knockout/blob/8535832f0bd22aeecb2aba53311bf55b77a7d6ad/src/utils.js#L294) though.

Worth noting: http://stackoverflow.com/a/3945236/19212 ; we could probably get rid of the `String` constructor itself, since `stringTrim` calls `toString`, meaning we could just use `ko.utils.stringTrim(value || '')`.

npbee:
Agreed on moving the `ko.utils.stringTrim` outside of the string constructor.  I had it both ways and couldn't decide which was better.

Regarding removing the `String` constructor, I'd definitely be a fan of that but it looks like `ko.utils.stringTrim` only calls `toString` if the `.trim` method is not available.  So we could potentially be calling `.trim` on a non-string.

brianmhunt:
On the `String` constructor - right you are, thanks. We'd want the `String` constructor there for cases where there's an `object` with a `trim` property.

mbest:
Thanks. Looks good.

Issue #1783, opened 2015-05-06
Fixed: IE8/IE9 cannot use `<template id="my-template">` instead of `<script id="my-template">`


franktopel:
Wrapping templates in `<template>`-tags instead of `<script type="text/html">`-tags does not work in Internet Explorer 8 and 9.

Supporting IE 8 and 9 in your knockout project forces you to use `<script>`-tags to wrap your templates. Using `<template>`-tags leads to these templates being evaluated at page load rather than when the template is being instantiated at run-time - which ofc leads to the bindings within the template failing because in most cases there will not yet be data/the objects being used available.

mbest:
Implementation note: This should be as simple as adding `<template>` to the list of elements to skip while binding, in addition to `<script>` and `<textarea>`.

franktopel:
I've tested your suggestion and it looks like that's all that must be done.

I've forked `bindingAttributeSyntax.js` and implemented your suggested fix in the fork:

https://github.com/knockout/knockout/blob/32628d0d8ce1089555eccfcfdacb8fc860edcbf2/src/binding/bindingAttributeSyntax.js

franktopel:
I've added a pull request for the fix: https://github.com/knockout/knockout/pull/1785/files

mbest:
Fixed.

PR #1785, opened 2015-05-07
Fix for #1783: Added <template> to the list of elements to skip while…

franktopel:
… binding

This fixes https://github.com/knockout/knockout/issues/1783


mbest:
Thanks. I merged this change along with a small spec test to verify.

mbest:
Note that the issue affect all versions of Internet Explorer, not just 8 and 9.

PR #1591, opened 2015-05-07
Binding checkboxes to writeable computedobservable array

mbest:
I found an issue when binding a list of checkboxes to a computedobservable. It thinks that push and splice are available in addOrRemoveItem which they are not. It is fixed with this code. Could you please implement it:

 addOrRemoveItem: function(array, value, included) {
            var existingEntryIndex = ko.utils.arrayIndexOf(ko.utils.peekObservable(array), value);
            if (existingEntryIndex < 0) {
                if (included) {
                    // Original: array.push(value)
                    var arr = array();
                    arr.push(value);
                    array(arr);
                }
            } else {
                if (!included) {
                    // Original: array.splice(existingEntryIndex, 1);
                    var arr = array();
                    arr.splice(existingEntryIndex, 1);
                    array(arr);
                }
            }
        },

### DIFF
--- a/spec/defaultBindings/checkedBehaviors.js
+++ b/spec/defaultBindings/checkedBehaviors.js
@@ -156,6 +156,34 @@ describe('Binding: Checked', function() {
         expect(testNode.childNodes[0].checked).toEqual(false);
     });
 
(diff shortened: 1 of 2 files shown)


mbest:
I've attached a possible fix for this problem to this issue. Let me know what you think.

PR #1709, opened 2015-05-07
compareArrays 'dontLimitMoves' option is no longer respected

mbest:
In commit [`0e0559`](https://github.com/knockout/knockout/commit/0e0559bdfb67e707a45afe7707fc8a228aab56de), support for the obscure `dontLimitMoves` option was removed from `compareArrays.js`. I'm unsure whether this was deliberate.

There is still some hangover code that interprets a boolean `options` value as meaning `{ 'dontLimitMoves': yourBooleanOptionsValue }`, but no other code actually uses the `dontLimitMoves` setting anywhere. I guess this flag was somehow missed from test coverage.

I'm not saying we should put it back, unless anyone has a practical need for it (in which case, I'm surprised this wasn't reported already).

### DIFF
--- a/spec/arrayEditDetectionBehaviors.js
+++ b/spec/arrayEditDetectionBehaviors.js
@@ -88,4 +88,18 @@ describe('Compare Arrays', function() {
             { status: "deleted", value: "E", index: 4, moved: 2 }
         ]);
     });
+
+    it('Should honor "dontLimitMoves" option', function() {
+        // In order to test this, we must have a scenario in which a move is not recognized as such without the option.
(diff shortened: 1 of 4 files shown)


mbest:
I've attached my implementation of the fix here. I decided against making `dontLimitMoves` the default for `arrayChange`, but have made it possible to change the setting by applying the extender again.

I've also included spec tests to check that `dontLimitMoves` works.

mbest:
I'm going ahead with the `sparse` option being always `true` because it's likely that plugins or code in general that uses `arrayChange` won't expect to see `retained` entries.

PR #1380, opened 2015-05-08
observableArray sort function confusion

mbest:
`observableArray.sort()` is just a wrapper for `Array.sort` and thus sorts the underlying array, returning the array. I did a quick search and found many examples of using `Array.sort`, and not one of them used the return value. See the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort), for example.

But when sorting an observable array, I often see the return value of the `sort` function being used. For example, here's a [StackOverflow answer](http://stackoverflow.com/a/14513294/1287183) that suggests using `sort` in a binding like `foreach: items.sort(...)`. The biggest problem with this approach is that `observableArray.sort` doesn't establish a dependency on the array, so the binding won't get updated.

Of course, some people might use `foreach: items().sort(...)` instead. This provides the dependency on the array, but has a different problem--that other subscribers of the array don't get notified that the array was sorted.

To clarify the intended use of `observableArray.sort()`, I suggest that we change it to not return the array. To support displaying a sorted version of an observable array, we could add `observableArray.sortCopy()` (or `sorted()`, etc.) that will sort and return a copy the array.

### DIFF
--- a/spec/observableArrayBehaviors.js
+++ b/spec/observableArrayBehaviors.js
@@ -277,6 +277,14 @@ describe('Observable Array', function() {
         expect(timesEvaluated).toEqual(1);
     });
 
+    it('Should return the observableArray reference from "sort" and "reverse"', function() {
+        expect(testObservableArray.reverse()).toBe(testObservableArray);
+        expect(testObservableArray.sort()).toBe(testObservableArray);
+
+        // Verify that reverse and sort notified their changes
+        expect(notifiedValues).toEqual([ [3, 2, 1], [1, 2, 3] ]);
+    });
+
     it('Should inherit any properties defined on ko.subscribable.fn, ko.observable.fn, or ko.observableArray.fn', function() {
         this.after(function() {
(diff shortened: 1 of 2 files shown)


mbest:
I've attached the changes here for review.

mbest:
Since there were no complaints, I'm merging this.

wilson0x4d:
This is a really elegant solution, actually. I'm incredibly grateful, and shamefully I admit that until today I've never taken the time to look at the code backing these methods.

On that note, after looking over the majority of code I admit Knockout is some amazingly lean and concise code (where I've typically assumed otherwise while being impressed with its performance over the years), to wit, I want to thank everyone that has ever touched the code; especially @mbest, @SteveSanderson and @rniemeyer for all their hard work. Without you guys we'd certainly all be stuck rolling our own ad hoc solutions or, worse, using something excessive and unnecessary instead.

Issue #1775, opened 2015-04-17
Subscription/notification order issues


mbest:
Knockout notifies subscribers in the same order that they subscribe to an observable, although this isn't specifically covered by a test. As an experiment, I tried modifying `notifySubscribers` to notify in reverse order to see which tests failed. The main ones that failed were for the `if` binding, the `tempalte/if` option, and the `checked` binding. Those tests each demonstrate a condition where two computed observables depend on the same observable but also have an unspecified dependency between themselves.

As @dancras has described in #1761, the ordering of notifications can affect how many times a computed observable is evaluated and whether you get *intermediate* values. Logically, a proper ordering would involve notifying computed observables based on their depth, or breadth-first. As long as Knockout knows about all the relationships, it would be fairly easy to calculate an observable's depth. But as I've described above, Knockout already includes instances where dependencies aren't specified, and it's likely that user applications include these as well. Reordering by depth could break those relationships.

When using standard computed observables, the fact that they generally have to be defined in order by depth means that notifications will be approximately ordered by depth. But when using the `pure` or `deferEvaluation` options, this ordering can be completely opposite, in which the deepest subscribers are notified first.

While working on #1543, I ran into the problem described above—that a pure computed observable might report intermediate values, depending on the order of the subscriptions. I ended up putting in some extra effort to maintain that behavior in the new code paths, so that a particular pure computed observable would behave consistently. But I think the better solution is to make sure that the subscription order for pure computed observables matches what you'd generally get with plain computed observables, which will fix the intermediate values problem.

mbest:
While working on developing this over the last couple weeks, I've further refined the requirements as follows:

1. Any reordering should only occur when subscriptions are recursive due to computed observables are being awakened.
2. The reordering code should support being interrupted and use a stack to support recursive computed contexts (such as within control-flow bindings).
2. No additional code or memory should be used when reordering isn't needed (or as little as possible).
3. The reordering result should be such that when using deferred updates, pure computed observables are updated in a breadth-first order, and there are no duplicate updates.

mbest:
Maybe we can revisit this later, (I still have a copy of the code I worked on for this [note to self: branch name is `fix-pureComputed-dependency-order`].) but for now I'm closing the issue.

Issue #1758, opened 2015-04-09
Controlling which parts of the UI have their bindings deferred


SteveSanderson:
If I understand correctly, then the new "deferred" functionality applies to DOM bindings based on the global `ko.options.deferUpdates` flag. If (and only if) you have enabled `deferUpdates`, then the internal computeds we create to watch for observable changes will all themselves be deferred/batched which means repeated UI updates will be coalesced to a single update, which is great. The global flag means this is backwards-compatible and allows new apps to gain the benefits.

But for large existing apps it might not help much - many will be unable to enable `deferUpdates` globally because something will inevitably break (you'll have some functionality that relies on synchronous non-coalesced notifications, or you have tests that do, or you're just not confident you could even figure out if enabling the flag breaks anything).

So, to benefit existing apps, would it make sense to enable this in the UI on a more granular level? Here are two possibilities:

### On a per-binding-handler basis

Some binding handlers work correctly when deferred (hopefully all of the built-in ones). Others would break (imagine a 'log' binding that appends to a log on each update - it would miss the coalesced notifications, even if the observable it's watching is itself explicitly not deferred). Perhaps binding handlers should be able declare whether they support deferred notification, and if they do, this would override the global default (at least negatively, when declaring "I don't support deferred" - not sure there's a good use case for overriding the default positively, and you could already override positively just by wrapping your `update` code in a deferred computed).

Then you can enable `deferUpdates` globally but turn it off for your custom binding handlers where either they don't support "deferred" operation, or where you're not confident you can even figure out whether they support it.

Possible drawback: If people try to subclass or combine binding handlers by writing `ko.bindingHandlers.newHandler = { ... code that invokes ko.bindingHandlers.otherHandler.update(), etc. ... }`, it won't bring the setting from `otherHandler` with it, and hence may try to operate in deferred mode when the underlying handler doesn't support it.

### On a DOM-subtree basis

Let's say you have a massive app and you have no idea which of your custom binding handlers or other code would break if you enabled `deferUpdates` globally. You don't have time or inclination to read every line of your code and reason about the timings of its interactions with other code, and/or you have a massive suite of tests that does break if you enable `deferUpdates` because it assumes synchronicity everywhere. So you're not going to enable it globally.

[...]

PR #1795, opened 2015-05-15
Deferred updates, part 3: dependency tree scheduling

mbest:
This is the third, and final, part of supporting deferred updates in Knockout (#1728).

With multiple deferred computed observables in a dependency tree, use "dirty" events to schedule the complete set of updates in the correct order and mark each computed observable as "dirty" so it can be evaluated if needed with the latest value.

### DIFF
--- a/spec/asyncBehaviors.js
+++ b/spec/asyncBehaviors.js
@@ -826,6 +826,42 @@ describe('Deferred', function() {
             expect(dependentComputed()).toEqual('C');
         });
 
+        it('Should delay update of dependent pure computed observable', function() {
+            var data = ko.observable('A'),
+                deferredComputed = ko.computed(data).extend({deferred:true}),
+                dependentComputed = ko.pureComputed(deferredComputed);
+
+            expect(dependentComputed()).toEqual('A');
+
+            data('B');
+            expect(deferredComputed()).toEqual('B');
+            expect(dependentComputed()).toEqual('A');
+
+            data('C');
+            expect(dependentComputed()).toEqual('A');
+
+            jasmine.Clock.tick(1);
+            expect(dependentComputed()).toEqual('C');
+        });
+
+        it('Should delay update of dependent rate-limited computed observable', function() {
+            var data = ko.observable('A'),
+                deferredComputed = ko.computed(data).extend({deferred:true}),
+                dependentComputed = ko.computed(deferredComputed).extend({rateLimit: 500});
+
+            expect(dependentComputed()).toEqual('A');
+
+            data('B');
+            expect(deferredComputed()).toEqual('B');
+            expect(dependentComputed()).toEqual('A');
+
+            data('C');
+            expect(dependentComputed()).toEqual('A');
+
+            jasmine.Clock.tick(1);      // Cause the "change" notification to propogate to the rate-limited computed
+            expect(dependentComputed()).toEqual('C');
+        });
+
         it('Should *not* delay update of dependent deferred computed observable', function () {
             var data = ko.observable('A'),
                 timesEvaluated = 0,
@@ -863,6 +899,30 @@ describe('Deferred', function() {
 
             jasmine.Clock.tick(1);
             expect(notifySpy.argsForCall).toEqual([ ['B'] ]);
+        });
+
+        it('Is superseded by rate-limit', function() {
+            this.restoreAfter(ko.options, 'deferUpdates');
+            ko.options.deferUpdates = true;
+
+            var data = ko.observable('A'),
+                deferredComputed = ko.computed(data),
+                dependentComputed = ko.computed(deferredComputed).extend({rateLimit: 500}),
+                notifySpy = jasmine.createSpy('notifySpy'),
+                subscription = dependentComputed.subscribe(notifySpy);
+
+            expect(dependentComputed()).toEqual('A');
+
+            data('B');
+            expect(deferredComputed()).toEqual('B');
+            expect(dependentComputed()).toEqual('A');       // rate-limited computed doesn't respond to "dirty" events
+
+            jasmine.Clock.tick(1);      // notifies "change" event to the rate-limited computed
+            expect(dependentComputed()).toEqual('B');       // updated on-demand
+            expect(notifySpy).not.toHaveBeenCalled();       // but no notification
+
+            jasmine.Clock.tick(500);
+            expect(notifySpy.argsForCall).toEqual([ ['B'] ]);       // rate-limited computed notifies after the specified delay
         });
 
         it('Should minimize evaluation at the end of a complex graph', function() {
--- a/spec/asyncBehaviors.js
+++ b/spec/asyncBehaviors.js
@@ -636,6 +636,17 @@ describe('Deferred', function() {
             expect(notifySpy.argsForCall).toEqual([ ['A'] ]);
         });
 
+        it('Should throw if you attempt to turn off deferred', function() {
+            // As of commit 6d5d786, the 'deferred' option cannot be deactivated (once activated for
+            // a given observable).
+            var observable = ko.observable();
+
+            observable.extend({deferred: true});
+            expect(function() {
+                observable.extend({deferred: false});    
+            }).toThrow('The \'deferred\' extender only accepts the value \'true\', because it is not supported to turn deferral off once enabled.');
+        });
+
         it('Should notify subscribers about only latest value', function() {
             var observable = ko.observable().extend({notify:'always', deferred:true});  // include notify:'always' to ensure notifications weren't suppressed by some other means
             var notifySpy = jasmine.createSpy('notifySpy');
--- a/spec/asyncBehaviors.js
+++ b/spec/asyncBehaviors.js
@@ -808,6 +808,46 @@ describe('Deferred', function() {
             expect(notifySpy.argsForCall).toEqual([ ['B'] ]);
         });
 
+        it('Should delay update of dependent computed observable', function() {
+            var data = ko.observable('A'),
+                deferredComputed = ko.computed(data).extend({deferred:true}),
+                dependentComputed = ko.computed(deferredComputed);
+
+            expect(dependentComputed()).toEqual('A');
+
+            data('B');
+            expect(deferredComputed()).toEqual('B');
+            expect(dependentComputed()).toEqual('A');
+
+            data('C');
+            expect(dependentComputed()).toEqual('A');
+
+            jasmine.Clock.tick(1);
+            expect(dependentComputed()).toEqual('C');
+        });
+
+        it('Should *not* delay update of dependent deferred computed observable', function () {
+            var data = ko.observable('A'),
+                timesEvaluated = 0,
+                computed1 = ko.computed(function () { return data() + 'X'; }).extend({deferred:true}),
+                computed2 = ko.computed(function () { timesEvaluated++; return computed1() + 'Y'; }).extend({deferred:true}),
+                notifySpy = jasmine.createSpy('notifySpy'),
+                subscription = computed2.subscribe(notifySpy);
+
+            expect(computed2()).toEqual('AXY');
+            expect(timesEvaluated).toEqual(1);
+
+            data('B');
+            expect(computed2()).toEqual('BXY');
+            expect(timesEvaluated).toEqual(2);
+            expect(notifySpy).not.toHaveBeenCalled();
+
+            jasmine.Clock.tick(1);
+            expect(computed2()).toEqual('BXY');
+            expect(timesEvaluated).toEqual(2);      // Verify that the computed wasn't evaluated again unnecessarily
+            expect(notifySpy.argsForCall).toEqual([ ['BXY'] ]);
+        });
+
         it('Is default behavior when "ko.options.deferUpdates" is "true"', function() {
             this.restoreAfter(ko.options, 'deferUpdates');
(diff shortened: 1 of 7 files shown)


mbest:
Originally, in #1728, I thought that we could use the observable versioning system, which is used to efficiently to update sleeping pure computed observables, to accomplish a similar task for deferred updates. At the very least, I wanted to try it and compare it to using a `dirty` event. After completing #1753 (part 2), though, I took a detour with #1775 to develop a system to properly order subscriptions, especially for deferred updates. After spending a couple of weeks working on it, I realized that it wouldn't be enough to accomplish the goal for deferred updates: that pure computed observables are updated in a breadth-first order, and there are no duplicate updates. Combining it with version checking might help, but the whole thing was big and not very efficient.

So I went back to trying the `dirty` event, understanding that it could solve both problems for deferred updates. Scheduling (or rescheduling) the update of each computed observable as it receives `dirty` notifications results in the correct update order. And because this marks each observable as *dirty* in the process, it efficiently causes them to evaluate to the latest value if accessed while an update is pending.

spec/asyncBindingBehaviors.js
```diff
@@ -125,4 +125,13 @@ describe("Deferred bindings", function() {
...
     });
+
+    it('Should throw an exception for value binding on multiple select boxes', function() {
+        testNode.innerHTML = "<select data-bind=\"options: ['abc','def','ghi'], value: x\"></select><select data-bind=\"options: ['xyz','uvw'], value: x\"></select>";
+        var observable = ko.observable();
+        ko.applyBindings({ x: observable }, testNode);
+        expect(function() {
+            jasmine.Clock.tick(1);
+        }).toThrowContaining('Too much recursion');
+    });
```
mbest: This test presents a concrete example of how the recursive task check could come into play (and is the reason I added the check in the (deferred updates plugin)[https://github.com/mbest/knockout-deferred-updates/issues/17]). It's interesting, though, that this example doesn't result in an infinite loop before the code changes here.

src/subscribables/dependencyDetection.js
```diff
@@ -60,6 +60,5 @@ ko.computedContext = ko.dependencyDetection = (function () {
 ko.exportSymbol('computedContext', ko.computedContext);
 ko.exportSymbol('computedContext.getDependenciesCount', ko.computedContext.getDependenciesCount);
 ko.exportSymbol('computedContext.isInitial', ko.computedContext.isInitial);
-ko.exportSymbol('computedContext.isSleeping', ko.computedContext.isSleeping);
```
mbest: I noticed that we were exporting a non-existent property, so I took it out.

mbest:
@SteveSanderson, I'm hoping to get your feedback here. This obviously involves some code changes within `dependentObservable.js`, but I hope that the logic is isolated enough to be easily digestible.

spec/asyncBehaviors.js
```diff
@@ -808,6 +808,82 @@ describe('Deferred', function() {
...
+            expect(dependentComputed()).toEqual('A');
+
+            data('C');
+            expect(dependentComputed()).toEqual('A');
+
+            jasmine.Clock.tick(1);      // Cause the "change" notification to propogate to the rate-limited computed
+            expect(dependentComputed()).toEqual('C');
+        });
+
+        it('Should *not* delay update of dependent deferred computed observable', function () {
```
SteveSanderson: This subtlety - how notifications and re-evaluations differ between regular, ratelimited, and deferred computeds is very subtle indeed and is going to require careful attention in docs. Probably some big table of comparisons, with notes that provide justifications for all the differences.

I think I can guess why these differences exist, but it was still surprising to me at first.

spec/asyncBehaviors.js
```diff
@@ -824,5 +900,66 @@ describe('Deferred', function() {
...
+                subscription = dependentComputed.subscribe(notifySpy);
+
+            expect(dependentComputed()).toEqual('A');
+
+            data('B');
+            expect(deferredComputed()).toEqual('B');
+            expect(dependentComputed()).toEqual('A');       // rate-limited computed doesn't respond to "dirty" events
+
+            jasmine.Clock.tick(1);      // notifies "change" event to the rate-limited computed
+            expect(dependentComputed()).toEqual('B');       // updated on-demand
```
SteveSanderson: This seems really weird. The dependent computed updates on demand, but *only* from the start of the next execution cycle? What's the use case for such a special and potentially surprising behaviour? I wouldn't have anticipated this. We can't assume devs are going to know about internal details like a "dirty" event.

spec/asyncBindingBehaviors.js
SteveSanderson: Are you regarding this as not a breaking change because the previous behavior was not well defined? I'm not sure what the previous behavior would have been, as it does seem impossible for 'x' to satisfy the requirements of both 'select' boxes simultaneously.

src/subscribables/dependentObservable.js
```diff
@@ -57,6 +57,34 @@ ko.computed = ko.dependentObservable = function (evaluatorFunctionOrOptions, eva
         isSleeping = false;
     }
 
+    function subscribeToDependency(target) {
+        if (target._deferUpdates) {
+            var dirtySub = target.subscribe(function () {
+                    // Process "dirty" events if both us and the target are using deferred updates
+                    if (dependentObservable._deferUpdates && !_isBeingEvaluated) {
+                        dependentObservable._evalRateLimited();
```
SteveSanderson: Should _evalRateLimited be renamed, if it's now being used in non-rate-limted scenarios too?

src/subscribables/dependentObservable.js
```diff
@@ -57,6 +57,34 @@ ko.computed = ko.dependentObservable = function (evaluatorFunctionOrOptions, eva
         isSleeping = false;
     }
 
+    function subscribeToDependency(target) {
```
SteveSanderson: This function will be quite a lot more expensive than the usual 'subscribe' if we go into the 'deferred' code path, both in memory and CPU use, as now it constructs several extra function instances with their own closures. This once for each dependency, every time any 'deferred' computed is re-evaluated.

It might be OK, but is there any possibility of dropping these dynamically created closures in favour of static function instances and variables that get mutated?

src/subscribables/subscribable.js
```diff
@@ -129,7 +129,8 @@ var ko_subscribable_fn = {
         } else {
             var total = 0;
             ko.utils.objectForEach(this._subscriptions, function(eventName, subscriptions) {
-                total += subscriptions.length;
+                if (eventName != 'dirty')
```
SteveSanderson: `!==`, maybe?

Minor point, I know!

SteveSanderson:
Thanks for this! Comments and questions added.

spec/asyncBehaviors.js
mbest: Now that you bring this up, I can see how it would be confusing. I'll see how I could make it work the other way.

spec/asyncBindingBehaviors.js
mbest: Right. I'd say that the previous behavior was undefined.

src/subscribables/dependentObservable.js
mbest: I had originally written this with static functions but then realized I needed to access `target` in one of them. I'll take another look at it, though, to see if there's another way. Also this will only run once for each dependency because when a computed is re-evaluated, it saves and reuses previous subscriptions.

src/subscribables/dependentObservable.js
mbest: Makes sense.

src/subscribables/subscribable.js
mbest: Sure.

mbest:
Another point to considered before merging this is that "dirty" events will cause a computed to re-evaluate even if the underlying observable doesn't *change*. In my plugin, I dealt with this by having two levels of "dirty". A `dirty` event marks a computed as *possibly* dirty, and a `change` event marks it as *actually* dirty. When *possibly* dirty, it will be evaluated on demand, but only if *actually*  dirty will it be evaluated during task processing. Do you think I should try to introduce the same logic here?

mbest:
@SteveSanderson I've made some changes that address most of the concerns you brought up. Please check them over and give me your thoughts.

spec/asyncBindingBehaviors.js
```diff
@@ -125,4 +125,14 @@ describe("Deferred bindings", function() {
         expect(testNode.childNodes[0]).toContainHtml('<span data-bind="text: childprop">moving child</span><span data-bind="text: childprop">first child</span><span data-bind="text: childprop">second child</span>');
         expect(testNode.childNodes[0].childNodes[targetIndex]).not.toBe(itemNode);    // node was create anew so it's not the same
     });
+
+    it('Should not throw an exception for value binding on multiple select boxes', function() {
+        testNode.innerHTML = "<select data-bind=\"options: ['abc','def','ghi'], value: x\"></select><select data-bind=\"options: ['xyz','uvw'], value: x\"></select>";
```
mbest: Given that the non-deferred scenario doesn't throw an exception, I decided to give this some more thought. I realized that if the computed observables used for updating the binding don't respond to "dirty" events, that would stop the recursion as long as the underlying observable suppresses notifications for non-changing updates. So this will still cause a "recursion" exception if the values are objects instead of strings. @SteveSanderson, do you think this change is helpful or not?

SteveSanderson: If this brings the deferred behavior more into line with non-deferred, then I guess it's a good thing, but it's hard to think of a case where someone would be doing this and expect something useful to happen :)

SteveSanderson:
Thanks again for implementing this, and sorry it's taken a few weeks to complete the review.

This looks good to me - the notion of 'dirty' events propagating through any unbroken chain of deferreds to spread awareness that a change is coming, and hence that they should schedule a re-evaluation (and re-evaluate proactively if read) seems like a good way to model this.

I'm happy for this to go in (**so please merge it if you're ready, or let me know and I will**), but I just want to ask a couple of clarifying questions in case you want to add further tweaks:

* What's the reason why we can no longer support turning off `deferred`? Is it that we'd have to recreate all the subscriptions to dependencies (and would it be prohibitively difficult to implement that)? I know you put in the comment that "it would result in an observable that appears to be rate-limited but actually updates synchronously" but I'm unsure why that behavior is necessary.
 * I've added a check that you don't try to use `.extend({ deferred : false })` as that code would actually *enable* deferred :)
 * It seems odd that you actually can turn off `deferred` by turning on `rateLimit` (which you can do at arbitrary times). Would this lead to bugs? Why is it safe to enable `rateLimit` and hence disable `deferred`, if you can't just disable `deferred` on its own?
* Is it true that you can't have an observable/computed that is both deferred and rate-limited at the same time?
 * I hope not, because then there are so many combinations of [observable/computed/pureComputed] * [deferred on/off] * [rateLimit on/off] and I don't have any idea how the deferred+rateLimited combinations should behave
 * Should we add an exception if you try to use both extenders on the same thing?

mbest:
@SteveSanderson, I'm glad to get your feedback. Hopefully we can get this in soon.

> What's the reason why we can no longer support turning off deferred?

One reason is that I realized that it creates a new class of observables that would need to be thoroughly tested (and the implementation probably adjusted accordingly). Another reason is that I decided that I didn't want to encourage turning off deferred updates for a individual observable when using the global `deferUpdates` option, mostly because I couldn't decide how such a scenario should behave.

> Is it true that you can't have an observable/computed that is both deferred and rate-limited at the same time?

That's correct. Both use the internal `limit` function and setting one turns off the other.

> Why is it safe to enable rateLimit and hence disable deferred, if you can't just disable deferred on its own?

When using deferred updates globally, it still makes sense that people will sometimes want to use `rateLimit` for specific observables. The rate-limited observable will behave as expected whether used with other normal or deferred observables.

> Should we add an exception if you try to use both extenders on the same thing?

We need to support setting `rateLimit` on a deferred observable but not vice versa. We could add an exception for the latter case, although I don't think it's necessary.

spec/asyncBindingBehaviors.js
mbest: I think we may need to take a closer look at how bindings interact with the deferred updates feature, as you've described out in #1758. So this code may need to change some more anyway.

brianmhunt:
>> Is it true that you can't have an observable/computed that is both deferred and rate-limited at the same time?
> That's correct. Both use the internal limit function and setting one turns off the other.

Is deferring analogous to `rateLimit: 0` (or `1`), much like `setTimeout(fn, 0)`? If not, how are they different?

DanielCarbajal:
Date: Fri, 3 Jul 2015 20:11:23 -0700

mbest:
@brianmhunt That's a good question. If you use `deferred` for a single observable, it could be equivalent to `rateLimit: 0`, at least on older browsers. Once you have multiple deferred observables, you'll see more benefit because all updates happen together with no delay between them. This might help:

1. *Default* - Notifications happen immediately and synchronously.
2. *Deferred* - Notifications happen asynchronously as soon as possible, hopefully before any redraws.
3. *Rate-limit* - Notifications happen after a specified period of time (minimum of 2-10 ms depending on the browser).

brianmhunt:
@mbest Got it – that's very helpful (and probably worthwhile to reference in the docs!).

mbest:
This is merged now.

PR #1788, opened 2015-05-19
textInput error in IE9

mbest:
In IE9, notifySubscriber eats(destroys) HTML when undoing/CTRL+Z on a textarea with maxlength extender.

Reproduction steps:
0. Using IE9 (not IE11 with compatibility set to IE9)
1. Navigate to http://jsfiddle.net/h4r91bcg/3/ 
2. Paste in a text inside the textarea that is too long. Ex. "In snooker," (without the double quotes)
3. Hit CTRL+Z or right click undo
4. Repeat the paste/undo combo

Expected Result:
1. Text area clears the contents with no side effects to the surrounding DOM

Actual Result:
1. Undo eats up DOM due to IE9 bug.

Workaround (c/o Michael Best!)
1. Define pre databind tags to the text itself and the text.view
2. Further extend the ko.pureComputed function by adding in ".extend({ rateLimit: 0, notify: 'always' });" to the end
3. See http://jsfiddle.net/h4r91bcg/10/

### DIFF
--- a/src/binding/defaultBindings/textInput.js
+++ b/src/binding/defaultBindings/textInput.js
@@ -72,6 +72,10 @@ ko.bindingHandlers['textInput'] = {
             }
         };
 
+        // IE9 will mess up the DOM if you handle events synchronously which results in DOM changes (such as other bindings);
+        // so we'll make sure all updates are asynchronous
+        var ieUpdateModel = ko.utils.ieVersion == 9 ? deferUpdateModel : updateModel;
+
         var updateView = function () {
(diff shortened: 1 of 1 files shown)


mbest:
So this fixes the problem by delaying all updates from the binding in IE9.

mbest:
Since this is a simple change and seems to fix the problem described here (and possibly others), I'll merge this in.

chamchu:
Thanks Michael, sorry I didn't get a chance to thank you earlier. This looks good.

PR #1798, opened 2015-05-21
Fixing false-y values for checkedValue of lone checkboxes

mattliegey:
Checkboxes do not show as checked when checkedValue === modelValue if modelValue is false-y.

Example:
```javascript
model = {
  bool: ko.observable(false)
}

<input type="checkbox" data-bind="
    checked: bool,
    checkedValue: false
">
```

I would expect the checkbox to be checked as both checkedValue and model.bool both have the same value. Or is there a better way to set up a checkbox with a checkedValue of false/0/etc..

### DIFF
--- a/src/binding/defaultBindings/checked.js
+++ b/src/binding/defaultBindings/checked.js
@@ -66,8 +66,8 @@ ko.bindingHandlers['checked'] = {
                 // When a checkbox is bound to an array, being checked represents its value being present in that array
                 element.checked = ko.utils.arrayIndexOf(modelValue, checkedValue()) >= 0;
             } else if (isCheckbox) {
-                // When a checkbox is bound to any other value (not an array), being checked represents the value being trueish
-                element.checked = modelValue;
+                // When a checkbox is bound to any other value (not an array), being checked represents the value being trueish or equal to checkedValue()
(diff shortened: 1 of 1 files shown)


mbest:
The `checkedValue` option is only used when `checked` is bound to an array or when used for radio buttons.

mattliegey:
Is there a way to get the behavior I'm describing?

mbest:
In your example, what would un-checking the checkbox do?

mattliegey:
That's a good point. As written it would also set it too false. I'm working with a UI where checkboxes resemble switches and in one case the "on" position corresponds to false in our models/API. I suppose the best solution would be to have an intermediate observable to switch the value.

mbest:
> I suppose the best solution would be to have an intermediate observable to switch the value.

That sounds right.

mbest:
> I suppose the best solution would be to have an intermediate observable to switch the value.

That sounds right. Something like this:

```javascript
var bool = ko.observable(false);
var boolFlipped = ko.pureComputed({
    read: function() { return !bool(); },
    write: function(value) { bool(!value); }
});
```

mattliegey:
Yep, exactly what I ended up with. Thanks!

PR #1510, opened 2015-05-25
Knockout shouldn't define a "custom" element if a component name matches a standard element name

mbest:
For example, the following will basically break your app:

    ko.components.register('div', { });

demo: http://jsfiddle.net/mbest/hxqdbho9/

### DIFF
--- a/spec/components/customElementBehaviors.js
+++ b/spec/components/customElementBehaviors.js
@@ -26,6 +26,38 @@ describe('Components: Custom elements', function() {
         expect(testNode).toContainHtml('<div>hello <test-component>custom element <span data-bind="text: 123">123</span></test-component></div>');
     });
 
+    it('Inserts components into custom elements with matching non-dashed names', function() {
+        if (jasmine.ieVersion || window.HTMLUnknownElement) {   // Phantomjs 1.x doesn't include HTMLUnknownElement and will fail this test
+            this.after(function () { ko.components.unregister('somefaroutname'); });
+            ko.components.register('somefaroutname', {
+                template: 'custom element <span data-bind="text: 123"></span>'
+            });
+            var initialMarkup = '<div>hello <somefaroutname></somefaroutname></div>';
+            testNode.innerHTML = initialMarkup;
+
+            // Since components are loaded asynchronously, it doesn't show up synchronously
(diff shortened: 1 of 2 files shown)


mbest:
I've attached the code changes to this issue. Please let me know what you think.

reflectiveSingleton:
@mbest I think there might still be an issue with ```svg``` elements...I'm not too sure how common it is to bind to ```svg``` elements but the same issue persists with:

```javascript
ko.components.register('svg', { });
```

That will bork your application if you have any ```svg``` elements that also contain bindings in them because ```svg``` elements evaluate to ```[object HTMLUnknownElement]``` as a string value (a quick check verified this). I created a fiddle to demonstrate it: http://jsfiddle.net/gq16puf9/1/ (as can see I patched in your updated getComponentNameForNode code)

As a note, there may be issues with other elements but the only one I could think someone might bind to would be ```svg```. I think it would affect inner elements of ```svg``` as well such as ```rect```/etc.

Now why someone would call ```ko.components.register('svg', { });``` is beyond me...but I could say the same for registering ```div``` as a component so if that is an issue I would think this is too?

mbest:
@reflectiveSingleton Thanks for testing with `svg` elements. The problem isn't that they evaluate to `[object HTMLUnknownElement]` but that they have a lower case tagName and so match `node.tagName === tagNameLower`, which is meant to check for unknown node types in IE8 and lower. I've modified the code to perform that check only if the browser is IE<=8.

reflectiveSingleton:
ah...my oversight...anyways, glad I could bring the issue up and you were able to resolve it.

jococo:
Will this change disallow us from extending built-in HTML elements?

We are building a markup language that sits on top of HTML. We currently use custom javascript but are looking into switching to Knockout for its Custom Element support and other nice features that Knockout will enable.

We would definitely vote for a flag to allow this to be turned off if it bars us from achieving this requirement.

mbest:
@jococo Can you provide more detail about why you'd want to extend a built-in element?

jococo:
The markup we layering on to HTML is very specific to our product, some tags we will want to change or enhance their purpose. We also will be providing a means for developers who use or product to create their own extensions to the language.

PR #1808, opened 2015-06-10
Update component-overview.md

gdelhumeau:
The current example does not work. The included file must declare an AMD component. This is confusing when we try to use the example as it is presented.

### DIFF
--- a/documentation/component-overview.md
+++ b/documentation/component-overview.md
@@ -126,21 +126,23 @@ Now `like-or-dislike` can be consumed in the same way as before, using either a 
     <button data-bind="click: addProduct">Add a product</button>
 {% endcapture %}
 {% capture live_example_viewmodel %}
-    function Product(name, rating) {
-        this.name = name;
-        this.userRating = ko.observable(rating || null);
-    }
+    define(['knockout'], function(ko) {
+        function Product(name, rating) {
+            this.name = name;
+            this.userRating = ko.observable(rating || null);
+        }
+    
+        function MyViewModel() {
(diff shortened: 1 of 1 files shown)


rniemeyer:
These changes break the live example on the page. The current example uses require.js to load the viewmodel/template as part of the component, but otherwise uses the global KO without AMD/require.js.

I can understand how this is could be confusing and is not a normal setup.

To make this work, we would potentially have to change the block like:

```js
define("MyViewModel", ["knockout"], function(ko) {
    function Product(name, rating) {
        this.name = name;
        this.userRating = ko.observable(rating || null);
    }

    function MyViewModel() {
        this.products = ko.observableArray(); // Start empty
    }

    MyViewModel.prototype.addProduct = function() {
        var name = 'Product ' + (this.products().length + 1);
        this.products.push(new Product(name));
    };

    return MyViewModel;
});

require(["knockout", "MyViewModel"], function(ko, MyViewModel) {
    ko.applyBindings(new MyViewModel(), document.getElementById("component-amd"));
});
``

We have to use a named module "MyViewModel", as this is not being loaded from a file. In addition to that, the registration of the component happened with the default KO instance which is not the same as the one required through AMD, so we would potentially have to either wrap the registration in a module and require it before applyBindings or setup the require config (global-scripts.html) to define the knockout module as `window.ko` so they are the same.

We need to find a balance between a simple example and one that is not confusing. Maybe it would be appropriate to add a note to the example to indicate that normally this would be part of a module.

PR #1256, opened 2015-06-15
How should foreach binding deal with items mid-animation when updating afterwards

mbest:
From: http://stackoverflow.com/q/14828688/1287183

When adding an item while the `beforeRemove` animation is running, the removed element is moved to the bottom of the list instead of staying in its position until the animation has finished and the element is removed.

Fiddle: http://jsfiddle.net/bPP5Q/8/

### DIFF
--- a/spec/defaultBindings/foreachBehaviors.js
+++ b/spec/defaultBindings/foreachBehaviors.js
@@ -144,35 +144,53 @@ describe('Binding: Foreach', function() {
     });
 
     it('Should be able to supply afterAdd and beforeRemove callbacks', function() {
-        testNode.innerHTML = "<div data-bind='foreach: { data: someItems, afterAdd: myAfterAdd, beforeRemove: myBeforeRemove }'><span data-bind='text: childprop'></span></div>";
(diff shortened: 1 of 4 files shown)


mbest:
I've worked on a fix for this problem, which I've attached to this issue. The fixed version can be seen demonstrated here: http://jsfiddle.net/mbest/8k8V5/2734/

mbest:
The fix is to have the `foreach` binding continue tracking the "removed" nodes until they are actually removed. This means that the `beforeRemove` handler must remove all of the nodes in order for the binding to stop tracking them. The original jsFiddle example didn't remove non-element nodes (such as text nodes) and so the binding would track all removed items forever.

mbest:
@rniemeyer, Can you take a look at this and give me your thoughts?

rniemeyer:
@mbest - sure, will probably be early next week, but happy to take a look

rniemeyer:
@mbest - great fix. works very well.

mbest:
Thanks!

PR #1838, opened 2015-08-03
Update template binding recursion comment

borkxs:
Bringing comments up to date to reflect addition of template to `bindingDoesNotRecurseIntoElementTypes`.


mbest:
Good catch.

PR #1840, opened 2015-08-07
Make observables 2-3 times faster to instantiate, and reduce their memory use by over half

SteveSanderson:
I've been doing some perf investigations for the very large KO-based web app I work on. Although KO is only a small slice of our overall CPU use, there are some worthwile opportunities for improvement.

In particular I've found that the instantiation of `ko.observable` and `ko.computed` take up a large percentage of the time our app spends inside KO code. Also, since we create such a huge number of them, the memory use and resulting GC time is quite noticeable too.

## Net result

This pull request makes `ko.observable` about 2-3 times faster to instantiate. On a test where I instantiate 1 million of them, the time drops from roughly 2.7s to roughly 1.2s on Chrome, and from roughly 4.5s to 1.5s on Firefox.

This pull request also reduces the memory use by over half. According to the Chrome profiler, the memory overhead of each observable reduces from about 500 bytes to about 225 bytes with this change.

## Approach

The most expensive thing seems to be constructing function instances. A `ko.observable` inherently needs at least one unique one (since an observable *is* a function instance), but we can avoid the need to construct additional ones (for `peek`, `valueHasUpdated`, etc.) by moving those to `ko.observable.fn`. Making this work means that `_latestValue` can no longer be a closure variable, but instead needs to be a property on the observable instance.

A further minor speedup comes from streamlining how we inherit from `ko.subscribable`. For browsers that support `__proto__` modification, we don't have to add `ko.subscribable.fn` to the prototype chain on each instantiation, because it comes for free when we inherit from `ko.observable.fn`.

I tried various approaches to speed up creation of the one remaining function instance in each `ko.observable`, but every alternative I tried was slower. For example, I tried having a single shared function and then cloning it using `.bind()`, but that was much slower. I also tried using ES2015 inheritance (i.e., `extends Function`), and that's unbelievably slow.

## Drawbacks

None that I know of. Compatibility should be identical (unless you do something mad like identity-comparing `someObservable.peek` function instances). We can even retain the encapsulation of `_latestValue` in non-debug mode by using an ES2015 `Symbol` where available.

The only behavior change I can think of is that now, in debug mode, you can write to the private `_latestValue` property and it will silently change the value of the observable, whereas before if you did that it would not change the value of the observable since it was a separate copy of the value. I don't think this is a problem because that's obviously not a supported scenario (it's a private property) and it is not the case in release builds (at least, on browsers that support `Symbol`).

## Next steps

I'm going to look into `ko.computed` and see if similar techniques reduce its setup, evaluation, and memory costs.

### DIFF
--- a/src/subscribables/observable.js
+++ b/src/subscribables/observable.js
@@ -1,12 +1,15 @@
+var shouldUseSymbol = !DEBUG && typeof Symbol === 'function';
+var latestValueSymbol = shouldUseSymbol ? Symbol('_latestValue') : '_latestValue';
+
 ko.observable = function (initialValue) {
     function observable() {
         if (arguments.length > 0) {
             // Write
 
             // Ignore writes if the value hasn't changed
-            if (observable.isDifferent(observable._latestValue, arguments[0])) {
+            if (observable.isDifferent(observable[latestValueSymbol], arguments[0])) {
                 observable.valueWillMutate();
-                observable._latestValue = arguments[0];
+                observable[latestValueSymbol] = arguments[0];
                 observable.valueHasMutated();
             }
             return this; // Permits chained assignments
@@ -14,11 +17,11 @@ ko.observable = function (initialValue) {
         else {
             // Read
             ko.dependencyDetection.registerDependency(observable); // The caller only needs to be notified of changes if they did a "read" operation
-            return observable._latestValue;
+            return observable[latestValueSymbol];
         }
     }
 
-    observable._latestValue = initialValue;
+    observable[latestValueSymbol] = initialValue;
 
     // Inherit from 'subscribable'
     if (ko.utils.canSetPrototype) {
@@ -31,7 +34,7 @@ ko.observable = function (initialValue) {
 
     // Inherit from 'observable'
     ko.utils.setPrototypeOfOrExtend(observable, observableFn);
-    
+
     if (ko.options['deferUpdates']) {
         ko.extenders['deferred'](observable, true);
     }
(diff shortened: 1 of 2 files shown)


rniemeyer:
@SteveSanderson That looks great and makes perfect sense to me.

src/subscribables/observable.js
```diff
@@ -17,20 +17,23 @@ ko.observable = function (initialValue) {
...
-    observable.peek = function() { return _latestValue };
-    observable.valueHasMutated = function () { observable["notifySubscribers"](_latestValue); }
-    observable.valueWillMutate = function () { observable["notifySubscribers"](_latestValue, "beforeChange"); }
+    observable[latestValueSymbol] = initialValue;
 
-    ko.exportProperty(observable, 'peek', observable.peek);
-    ko.exportProperty(observable, "valueHasMutated", observable.valueHasMutated);
-    ko.exportProperty(observable, "valueWillMutate", observable.valueWillMutate);
+    // Inherit from 'subscribable'
+    if (ko.utils.canSetPrototype) {
```
brianmhunt: Should the logic in this block be part of some generic `ko.utils.adopt(fn, from, proto)`, that can be reused for `ko.observable`, `observableArray`, `computed` and user-defined observables?

SteveSanderson: That's a good question, but I'm unsure how such a thing would work. We've already factored out part of this logic (`ko.utils.setPrototypeOfOrExtend`).

I'm going to perform some similar changes to `ko.computed` - when that's done we'll have a clearer idea of the commonalities and what might be factored out.

SteveSanderson:
Looks like there were no objections so I've merged it.

Brian, I take your point about possibly being able to factor out more, and will look for an opportunity to do that as I work on computeds.

brianmhunt:
:+1: Awesome.

harry8525:
Seems like we should have one precomputed prototype to extend here instead of possibly 3 (at least 2):

// Inherit from 'subscribable' 
27     if (ko.utils.canSetPrototype) { 
28         // 'subscribable' will come for free when we inherit from 'observable', so just set up the internal state needed for subscribables 
29         ko.subscribable['fn'].init(observable); 
30     } else { 
31         // 'subscribable' won't be on the prototype chain unless we put it there directly 
32         ko.subscribable.call(observable); 
33     } 
34 
 
35     // Inherit from 'observable' 
36     ko.utils.setPrototypeOfOrExtend(observable, observableFn); 
37 
 
38     if (ko.options['deferUpdates']) { 
39         ko.extenders['deferred'](observable, true); 
40     }

PR #1855, opened 2015-08-21
Fix for potential memory leak

massimoahmadi:
We detected that when we remove a component binding from the page (via a knockout if binding) then for some reason our viewModel still lives in the heap. The retainer (in the memory profiler) seems to be the "currentViewModel" variable... I wonder if it is required once the viewModel has been disposed? Because setting it to null like this seems to fix the problem and completely remove the viewModel from the heap. Is there another way this gets cleaned up that we're missing?

### DIFF
--- a/src/components/componentBinding.js
+++ b/src/components/componentBinding.js
@@ -11,7 +11,7 @@
                     if (typeof currentViewModelDispose === 'function') {
                         currentViewModelDispose.call(currentViewModel);
                     }
-
+                    currentViewModel = null;
                     // Any in-flight loading operation is no longer relevant, so make sure we ignore its completion
                     currentLoadingOperationId = null;
                 },


brianmhunt:
Thanks @massimoahmadi – great catch.  This looks okay to me, offhand.  I'll flag this for 3.4 in hopes it'll get reviewed. 

Offhand, it looks straightforward and innocuous to me.

chessels:
It looks like #1855 is the only open issue for the [3.4.0 milestone](/knockout/knockout/milestones/3.4.0). The fix looks okay to me.

It would be great to get 3.4.0 out the door!

brianmhunt:
Thanks @chessels !

mbest:
I've merged this.

PR #1856, opened 2015-08-24
Add afterRenderAll and beforeRenderAll callbacks to foreach binding

zachpainter77:
This pull request will allow the developer define methods to be called before any items are rendered and after alll items have rendered.

### DIFF
--- a/src/binding/defaultBindings/foreach.js
+++ b/src/binding/defaultBindings/foreach.js
@@ -21,6 +21,8 @@ ko.bindingHandlers['foreach'] = {
                 'afterAdd': unwrappedValue['afterAdd'],
                 'beforeRemove': unwrappedValue['beforeRemove'],
                 'afterRender': unwrappedValue['afterRender'],
+                'afterRenderAll': unwrappedValue['afterRenderAll'],
+                'beforeRenderAll': unwrappedValue['beforeRenderAll'],
                 'beforeMove': unwrappedValue['beforeMove'],
                 'afterMove': unwrappedValue['afterMove'],
                 'templateEngine': ko.nativeTemplateEngine.instance
--- a/src/binding/editDetection/arrayToDomNodeChildren.js
+++ b/src/binding/editDetection/arrayToDomNodeChildren.js
@@ -118,6 +118,9 @@
         // Store a copy of the array items we just considered so we can difference it next time
         ko.utils.domData.set(domNode, lastMappingResultDomDataKey, newMappingResult);
 
+        if(options['beforeRenderAll'])
+            options['beforeRenderAll']();
+        
         // Call beforeMove first before any changes have been made to the DOM
         callCallback(options['beforeMove'], itemsForMoveCallbacks);
 
@@ -153,6 +156,9 @@
         // Finally call afterMove and afterAdd callbacks
         callCallback(options['afterMove'], itemsForMoveCallbacks);
         callCallback(options['afterAdd'], itemsForAfterAddCallbacks);
+        
+        if(options['afterRenderAll'])
+            options['afterRenderAll']();
     }
 })();
 


zachpainter77:
@mbest  recently you closed a pull request because you said it do anything.. I am new at git hub and created the pull request incorrectly.. This one definitely does something.. please commit this changes thanks.

zachpainter77:
This is how I am using it... http://jsfiddle.net/zachpainter77/4tLabu56/ 

This fiddle is for getting knockout to work with jquery datatables..

zachpainter77:
@mbest @brianmhunt @SteveSanderson @rniemeyer @paglias @chrisprice 

why is this pull request being ignored???

fervanrijswijk:
This seems to be very handy and a solution for my problem!

brianmhunt:
@zachpainter77  Thanks for the PR.

I would hazard a guess why this has not seen much discussion, namely:

1. everyone is busy :) 
2. 3.4 is upcoming, and that's getting priority
3. there are no unit tests with the PR
4. this changes the exposed API and these changes have not been discussed or considered (especially the implications with some forward-looking changes that might include `foreach`)

fervanrijswijk:
How can I put this code to work? There is no (unminified) bundled knockoutjs where I can test this?

zachpainter77:
Okay, @brianmhunt, I understand... Thank you for taking the time to reply! I really appreciate that.. This pull request does not conflict with existing code.. As for the Unit Tests... I'm not sure what kind of tests that you would want.  I've used the branched version with the pull request in this fiddle --> http://jsfiddle.net/zachpainter77/4tLabu56/ and it works... My branch is located here... --> https://github.com/zachpainter77/zach-knockout.js ... I was hoping that this small pull request that would solve so many problems would make it into 3.4... 

Thanks for your time again...

zachpainter77:
@fervanrijswijk, You can use the debug version of knockout.js @  my fork located here: --> https://github.com/zachpainter77/zach-knockout.js ... here you can download the file or just plop a reference to this version of knokcout from here --> https://rawgit.com/zachpainter77/zach-knockout.js/master/zach-knockout.debug.js

fervanrijswijk:
@zachpainter77 Thanks!

fervanrijswijk:
I'm sorry when I am using this thread for cantacting zachpainter, but I cant find a way to PM.

@zachpainter77  I'm trying trying to use your code, but I've not succeeded yet. Is this the correct way?

```
<div class="panel-body" data-bind="foreach: children, afterRender: renderComplete">
            <!-- ko template: { name: template} --><!-- /ko -->
        </div>
```

zachpainter77:
@fervanrijswijk, here is a simple example of what you are trying to do... http://jsfiddle.net/zachpainter77/n2ZTU/55/

fervanrijswijk:
@zachpainter77 Thanks, i got it to work. But when involving templates the renderAfterAll is not doing what i expect it to do. I have an alert() in the afterRenderAll() handler. After the alert is show the contect apears on screen. I expect the alert to display when al the elements are on the screen. 

Can you confirm that this code does not work when using templates in the foreach-loop?

zachpainter77:
@fervanrijswijk , it works! --> http://jsfiddle.net/zachpainter77/2xt17ehz/

fervanrijswijk:
@zachpainter77 The templatename in my case is a factory which will return the name of the template, that is why there are no quotes. I will see if i have time today to try to makes this work in my project. It obvious does work in your fiddle! Thanks :-)

mbest:
I'm about to start working on adding this feature and leaning towards including the one in #690 since it includes tests. There are some differences between that version and this, and I'd be interested to hear people's opinion on them:

1. #690 names them `before` and `after` instead of  `beforeRenderAll` and `afterRenderAll`.
2. It only calls them for updates, not for the initial render.
3. It doesn't call the callbacks if nothing actually changed.

fervanrijswijk:
point 2: The functionality I am after is that when the initial render takes place, I want to know if all the child elements are rendered as well so i can begin an animation or something. So on update would not be an option for me...

point 3. I can imagin that, you want to fire thos always, maye with an event opbject which contains a property which specifies if changes where made. i.e.: e.changes. This way it is up to the user to do what he wants. strat show loading spinner, stop showing loader spinner

zachpainter77:
@mbest thanks for adding this feature. I'll test it out when I get to work tomorrow. At first glance I like the code because it follows the architecture already in place.. I. E. Creating a function that tests for the function definition and executes if exists. There's more to it but I'll need to check it out tomorrow. I think that the names of the callbacks are to general and unintuitive in my opinion. Before and after. Before and after what? I like my names for the callbacks better. But I really dont like to get hung up on names. I just need to call a function before the binding renders any. And after the binding renders all. But like I said. I'll test out his code tomorrow in further depth.  Thanks again for your time.

fervanrijswijk:
Silly question, what is the expected release version for this improvement? And when is that released scheduled?

mbest:
We just released 3.4.0 so it'll be some months before another release. It may be 3.5.0 or 4.0.0.

fervanrijswijk:
Than i have to find a way to make the above example work for the time being. Thanks for answering :-)

fervanrijswijk:
I did some testing. I got it working in my project. But not all is well. 

I set up a jsFiddle in which I see the same pattern as in my project. On initiate (refresh, first opening the page) the 'afterRenderAll' call is called way before html elements are visible.

Clicking thru the app (simulate this by click on 'run' in jsFiddle), will trigger the 'afterRenderAll' correctly.

Is it possible too make it work on initializing?

zachpainter77:
@mbest I tested the other version shown in https://github.com/knockout/knockout/pull/690 This somewhat breaks my DataTablesForEach custom binding.  Mainly due to the callGlobalCallback function.  Is there any reason why we do not call the afterRenderAll() or after() callback on the first execution?  I absolutely **need** to call the call back on the first execution.  

``` javascript 
 function callGlobalCallback(callback) {
            if (callback && !isFirstExecution && editScript.length > 0)
                callback();
  }
```

I propose changing the callGlobalCallback definition, removing the test for the first execution.:

``` javascript
 function callGlobalCallback(callback) {
            if (callback && editScript.length > 0)
                callback();
  }
```

Maybe I'm missing something or the author of the pull-request had a specific need or something. But I can't seem to understand why you would define a function to execute after all elements are rendered **except** for the first time.

If someone has a special need like that, it seems like they would be able to define that in their actual callback definition on the view model. Something like this? What do you think?

``` 
<ul databind="foreach: { data: someItems, afterRenderAll: myAfterRenderCallBack }">
```

``` javascript
var vm = {
   isFirstExecution: true,
   myAfterRenderCallBack: function(){
       if(this.isFirstExecution){
          this.isFirstExecuition = false;
       } else {
          //doStuff
       }
   }
}
```

Other than the callback not being called on first execution, everything works as expected. I still do not like the names, 'before' and 'after', for the callbacks though.  I think they should be a bit more descriptive.  Since the callbacks are called before the elements bound to the array are rendered, they should contain the word 'render'. Since their is already an 'afterRender' callback that gets called after each element is rendered, the name of the callback should be something like 'afterRenderAll' or 'afterRenderFinished'. Something along those lines? The same goes for the 'before' callback.  Perhaps it should be 'beforeRenderAll' or 'beforeRenderAny' or 'beforeRenderStart'.  Do you agree?

mbest:
Creating a custom wrapper binding that calls these functions isn't hard and seems to work just as well:

```javascript
ko.bindingHandlers.foreach2 = {
    init: ko.bindingHandlers.foreach.init,
    update: function (element, valueAccessor, allBindings, viewModel, bindingContext) {
        var options = ko.unwrap(valueAccessor());
        ko.unwrap(options.data); // needed to set a dependency

        options.beforeRenderAll && options.beforeRenderAll();
        ko.bindingHandlers.foreach.update(element, valueAccessor, allBindings, viewModel, bindingContext);
        options.afterRenderAll && options.afterRenderAll();
    }
};
```

http://jsfiddle.net/2xt17ehz/10/

I'm not saying we can't add support for this to Knockout. I just want to make sure we're actually adding value.

zachpainter77:
@mbest Thanks for the simplification! That definitely works! I used it in a custom binding for jquery datatables. And verified that it works very amazingly.

``` javascript
ko.bindingHandlers.dataTablesForEach = {
	page: 0,
    init: function (element, valueAccessor, allBindingsAccessor, viewModel, bindingContext) {        
    	valueAccessor().data.subscribe(function (changes) {
        	var table = $(element).closest('table').DataTable();
            ko.bindingHandlers.dataTablesForEach.page = table.page();
            table.destroy();
        }, null, 'arrayChange');            
        var nodes = Array.prototype.slice.call(element.childNodes, 0);
        ko.utils.arrayForEach(nodes, function (node) {
        	if (node && node.nodeType !== 1) {
            	node.parentNode.removeChild(node);	
            }
        });
        return ko.bindingHandlers.foreach.init(element, valueAccessor, allBindingsAccessor, viewModel, bindingContext);
    },
    update: function (element, valueAccessor, allBindings, viewModel, bindingContext) {        
        var options = ko.unwrap(valueAccessor()),
            key = 'DataTablesForEach_Initialized';
        ko.unwrap(options.data);       
        ko.bindingHandlers.foreach.update(element, valueAccessor, allBindings, viewModel, bindingContext);
        (function() {
            console.log(options);
            var table = $(element).closest('table').DataTable(options.dataTableOptions);
            if (options.dataTableOptions.paging) {
                if (table.page.info().pages - ko.bindingHandlers.dataTablesForEach.page == 0) 
                    table.page(--ko.bindingHandlers.dataTablesForEach.page).draw(false);                
                else 
                    table.page(ko.bindingHandlers.dataTablesForEach.page).draw(false);                
            }
        })();
        if (!ko.utils.domData.get(element, key) && (options.data || options.length))
            ko.utils.domData.set(element, key, true);
        return { controlsDescendantBindings: true };
    }
}; 

```
I have a question about this line in your code:

``` javascript
ko.unwrap(options.data); // needed to set a dependency

```

I had no clue why removing this was breaking until I found this [How dependency tracking works](http://knockoutjs.com/documentation/computed-dependency-tracking.html). Now I understand! but the first line of this page is "Beginners don’t need to know about this".  Why make getting this functionality so obscure?  I don't think that adding more options can hurt knockout. A good API makes it easier to develop a program by providing **all** the building blocks.  Judging by how much this or different versions of this pull-request has come up, I think that it's safe to say that knockout devs want this feature.  This feature just makes things easier.

Another point is, what if devs want to get the same effect as this without creating a custom wrapper binding? What if they just want to do something like this:

```
<ul data-bind="foreach: { data: someItems, afterRenderAll: function() { /*doSomething*/ } }">
     <li data-bind="text: $data />
</ul>
```
Doesn't it defeat the purpose of the api to have to rewrap the out of the box bindings for something so small? I think that it would be nice for devs to be able to have easily accessible events to define callbacks before any items render and after all items render.  That way knockout will be able to play nice with many jquery plugins that rewrite the html and break the bindings.  **I think this would add tremendous value to knockout!**

brianmhunt:
Just out of curiosity, how should `afterRenderAll` and `beforeRenderAll` work for e.g. [knockout-fast-foreach](/brianmhunt/knockout-fast-foreach), which needs to be asynchronous and which batches changes (in what may be arbitrarily sized sets, depending on the async function).  Is the desired functionality replicable in that context?

zachpainter77:
@brianmhunt Doesn't fast-foreach have a beforeQueueFlush and afterQueueFlush events? do these events basically fire at the same time in the process as the afterRenderAll() and beforeRenderAll() events in my pull-request? (callbacks so folks can do things) :)

I imagine that you could just change those events to whatever name that the knockout devs here decide on for the callbacks? Or perhaps they would go with your naming convention.. 

``` javascript
FastForEach.prototype.processQueue = function () {
  var self = this;
  var lowestIndexChanged = MAX_LIST_SIZE;

  // Callback so folks can do things before the queue flush.
  if (typeof this.beforeQueueFlush === 'function') {
    this.beforeQueueFlush(this.changeQueue);
  }

  ko.utils.arrayForEach(this.changeQueue, function (changeItem) {
    if (typeof changeItem.index === 'number') {
      lowestIndexChanged = Math.min(lowestIndexChanged, changeItem.index);
    }
    // console.log(self.data(), "CI", JSON.stringify(changeItem, null, 2), JSON.stringify($(self.element).text()))
    self[changeItem.status](changeItem);
    // console.log("  ==> ", JSON.stringify($(self.element).text()))
  });
  this.rendering_queued = false;

  // Update our indexes.
  if (!this.noIndex) {
    this.updateIndexes(lowestIndexChanged);
  }

  // Callback so folks can do things.
  if (typeof this.afterQueueFlush === 'function') {
    this.afterQueueFlush(this.changeQueue);
  }
  this.changeQueue = [];
};


```

It looks like someone has already brought to your attention the need for "parity" of the bindings. So that fastforeach can be a drop in replacement for the original foreach binding.  I imagine if a knockout update breaks their existing code then they wouldn't want to update to the newest version of knockout. which is bad right?

[knockout-fast-foreach issue #21](https://github.com/brianmhunt/knockout-fast-foreach/issues/21)

zachpainter77:
@fervanrijswijk I noticed what you are talking about as well... however, this is not an issue with the afterRenderAll or beforeRenderAll callbacks.  This is an issue with jsFiddle and how the script is loaded into the iframe that displays your results.

fervanrijswijk:
@zachpainter77 In my project, not running in an iframe, I see the same behavior. What could be wrong in my project regarding loading the scripts?

zachpainter77:
@fervanrijswijk can you post a link to your project? I'd love to take a look at it and try and figure out what the problem is. You can contact me at http://zachpainter.net

mbest:
BTW, the idea I presented above came from looking through the https://typekit.com/fonts code (the site was mentioned in #690). Given this fairly simple way to achieve the requested functionality without modifying Knockout, I'm going to wait to do anything more with this.

fervanrijswijk:
@zachpainter77 I'm not able to expose the project because it has to much dependency's inside a private network. I will think about some way i could reproduce it more globally...

zachpainter77:
@fervanrijswijk I tested both @mbest 's solution and my solution with the modified knockout source with support for afterRenderAll and beforeRenderAll outside of jsfiddle. And, They both work as expected.  The page loads with no list items, alerts that no items have rendered, renders the items, then alerts that all items have rendered, then turns the text red.  I'm not sure what you are doing in your project, but I can assure you that it is not knockout that is to blame. Please contact me via email and send me a generic example as a .zip file or something, and I'd be glad to help.

zachpainter77:
I tried to upload a .zip but it wont let me.. lol.. 
image

zachpainter77:
@mbest Yes, you can achieve the same results without modifying knockout... but this feature should be out of the box. Users shouldn't have to rewrap the binding for this feature. Why so reluctant to add two simple events?!? It makes no sense. 

IMHO of course..

mbest:
Before adding this to Knockout, I want to make sure we get it right.

MrFrankel:
Hi, I want to add to this feature build on zachpainter77 .
I have added some tests for it and also sent the domNode reference to the callbacks as a param.

Should I create a new merge request or is there a way to add to this one?

mbest:
@MrFrankel It will need to be a new pull request. At this point, I'm still not sure what direction this issue will go in. I think @brianmhunt could help here based on his work on "fastforeach".

mbest:
With #2310, this is mostly now supported. There's now a `childrenComplete` event that can be handled either by explicitly subscribing to it or by adding a `childrenComplete` binding.

codymullins:
@mbest do we need to update the docs for the new bindings?

mbest:
@codymullins Yes. I'll create a 3.5.0 docs branch soon.

Seabizkit:
I dont understand does this or does this not exist in latest 3.5.1, how does one do afterRenderAll 

[zachpainter77](https://github.com/zachpainter77) commented [on Sep 23, 2015](https://github.com/knockout/knockout/pull/1856#issuecomment-142732576)

date on this is 2015 but its not available in 2022 am i missing something?

Issue #1858, opened 2015-08-27
Release version 3.4.0


mbest:
Steps to release:

- [x] Run tests on supported browsers and fix any failing tests.
- [x] Update documentation with new features (see #1873)
- [x] Write release notes.
- [x] Publish release candidate.
- [x] Respond to any urgent bugs.
- [x] Build and create release in Github.
  - [x] Update doc site with latest version.
- [x] Publish to NPM (@mbest)
- [ ] Publish to NuGet (@SteveSanderson).
- [ ] Post on Google groups, blogs, etc. about new release.

brianmhunt:
Great list, thanks @mbest.

For ease of reference: [The Travis-CI build from the #1360 branch](https://travis-ci.org/brianmhunt/knockout/builds/77599280), which has merged all the 3.4.0 changes.

The debug build from #1360 out to be near-bitwise-identical, and here are the errors reported on that branch:

- [ ] [Parse HTML fragments with jQuery](https://travis-ci.org/brianmhunt/knockout/builds/77599280#L149-L156)
- [ ] #1367 [Dependent Observable Should allow long chains without overflowing the stack](https://travis-ci.org/brianmhunt/knockout/builds/77599280#L216)
- [ ] IE 9, 10, and 11 – [A variety of issues](https://travis-ci.org/brianmhunt/knockout/builds/77599280#L1992-L2004):

<pre>
IE 10.0.0 (Windows 8 0.0.0) registerEventHandler should not use jQuery eventing with useOnlyNativeEvents option set to true FAILED
	Expected false to be true.
	Expected false to be true.
IE 9.0.0 (Windows 7 0.0.0) registerEventHandler should not use jQuery eventing with useOnlyNativeEvents option set to true FAILED
	Expected false to be true.
	Expected false to be true.
IE 11.0.0 (Windows 8.1 0.0.0) registerEventHandler should not use jQuery eventing with useOnlyNativeEvents option set to true FAILED
	Expected false to be true.
	Expected false to be true.
IE 9.0.0 (Windows 7 0.0.0) Binding: TextInput Should update observable on input event (on supported browsers) or propertychange event (on old IE) FAILED
	Expected 123 to equal 'some user-entered value'.
IE 11.0.0 (Windows 8.1 0.0.0) Binding: Hasfocus Should not unnecessarily focus or blur an element that is already focused/blurred FAILED
	Expected true to equal false.
</pre>

[...]

PR #1866, opened 2015-09-03
Fix parse html on old IE

SteveSanderson:
I don't think our current `parseHtmlFragment` implementation or its tests are correct. The tests claim success even though the parser output is actually wrong. This is because the assertions are only looking at `theParsedResult.outerHTML` instead of checking that the actual DOM element structure is correct. For custom elements like `<my-element></my-element>`, old IE parses it as two separate nodes - `<my-element>` and the very weird `</my-element>` element instead of one single element.

Fixing it is tricky. There are three cases:

1. **If you're just using KO alone**, then this pull request fixes the implementation and tests. We require you to be using the `document.createElement('my-element')` hack, since that is the only way that IE6-8 will not mangle the DOM. This pull request updates the test to use that hack. Also it updates the parsing code to be compatible with that hack (i.e., run the parsing in the same document context).
2. **If you're using KO and innershiv**, then it's still broken, because innershiv does the parsing inside a separate document context, so the `document.createElement` hack is of no use (it's a different context). Since innershiv itself is deprecated (in favour of html5shiv), and since it's always been broken and there's no reasonable way we can fix this anyway since it's a problem in innershiv, I propose we consider this scenario unsupported.
3. **If you're using KO and jQuery**, then it's still broken, because jQuery's HTML parsing has never supported custom elements on old IE. I am coming to think that, like @brianmhunt suggested in #1858, we should stop using jQuery for HTML parsing altogether. Thoughts on that:
 * Pro: For old browsers, it means we can support the KO+jQuery+oldIE+customelements combination. I can't think of any other way of supporting this combination.
 * Pro: Probably would have no impact on modern browsers, as KO's native parsing is likely to be totally robust - modern browsers don't have the edge cases that old IE does. If anything, our code is probably faster because it deals with fewer old-browser edge cases.
 * Pro: It means we wouldn't need 200f196c
 * Con: There is probably some unknown edge case where jQuery's HTML parser works correctly on old IE and KO's does not. Switching parsers could break some apps on old IE, though we don't know of any specific scenarios.

Overall, I think the decision on (3) comes down to whether it's more important not to break some as-yet-unknown edge-case scenario on old IE, or to enable support for custom elements on old IE. I am tending to prefer the latter (enabling custom elements by dropping the jQuery HTML parser) because old idiosyncratic apps targetting very old IE are probably not even upgrading to the latest KO, whereas newly-written apps that use the latest KO might well want to support IE8. If people report bugs in KO's HTML parser on IE6-8, we can fix them. **Anyone else got a preference on this?**

Further points:
* We should consider adding support for html5shiv as a replacement for the now-deprecated innershiv. I don't think it's urgent, since it only affects IE6-8 and our focus is not on adding features that only a vanishingly small percentage of applications are going to benefit from.
* The change in 200f196c that is meant to disable jQuery HTML parsing for elements with hyphens - I don't think this is a general solution, since it only looks at the top-level element and not any descendants. And changing it to look at all descendants would be a perf concern. It might be worse to have this than not, because some scenarios will appear to work correctly, but then break if the developer makes a seemingly innocuous change to their code (e.g., nesting some markup in a container). @mbest - is my assessment correct here? If so, can we revert 200f196c? I would prefer just not using jQuery HTML parsing at all, since at least the story would be consistent.

### DIFF
--- a/spec/defaultBindings/foreachBehaviors.js
+++ b/spec/defaultBindings/foreachBehaviors.js
@@ -575,23 +575,29 @@ describe('Binding: Foreach', function() {
     });
 
     it('Should be able to output HTML5 elements (even on IE<9, as long as you reference either innershiv.js or jQuery1.7+Modernizr)', function() {
-        // Represents https://github.com/SteveSanderson/knockout/issues/194
-        ko.utils.setHtml(testNode, "<div data-bind='foreach:someitems'><section data-bind='text: $data'></section></div>");
-        var viewModel = {
-            someitems: [ 'Alpha', 'Beta' ]
-        };
-        ko.applyBindings(viewModel, testNode);
-        expect(testNode).toContainHtml('<div data-bind="foreach:someitems"><section data-bind="text: $data">alpha</section><section data-bind="text: $data">beta</section></div>');
+        var isSupported = jasmine.ieVersion >= 9 || window.innerShiv || window.jQuery;
+        if (isSupported) {
+            // Represents https://github.com/SteveSanderson/knockout/issues/194
+            ko.utils.setHtml(testNode, "<div data-bind='foreach:someitems'><section data-bind='text: $data'></section></div>");
+            var viewModel = {
+                someitems: [ 'Alpha', 'Beta' ]
+            };
+            ko.applyBindings(viewModel, testNode);
+            expect(testNode).toContainHtml('<div data-bind="foreach:someitems"><section data-bind="text: $data">alpha</section><section data-bind="text: $data">beta</section></div>');
+        }
     });
 
     it('Should be able to output HTML5 elements within container-less templates (same as above)', function() {
-        // Represents https://github.com/SteveSanderson/knockout/issues/194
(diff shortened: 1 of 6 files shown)


brianmhunt:
> enable support for custom elements on old IE

I favour this as well.

I am in favour of taking out jQuery html parsing – I feel like the downside is pretty marginal compared to the upside, but I am open to being convinced otherwise.

IanYates:
I think this 
`because old idiosyncratic apps targetting very old IE are probably not even upgrading to the latest KO, whereas newly-written apps that use the latest KO might well want to support IE8. If people report bugs in KO's HTML parser on IE6-8, we can fix them
` 

makes the case.

:+1:

mbest:
I think this is great. Thanks, @SteveSanderson. There are a couple of tests that now fail in IE8 because of the lack of innershiv. Should we use the `document.createElement` "hack" to make those tests work?

brianmhunt:
@SteveSanderson  I believe you have the best triage we can do.

I have merged the changes into the #1360 branch, and the [test results are running](https://travis-ci.org/brianmhunt/knockout/builds/78768247).

Would we wish to update the Browser Support page of the docs with the caveats here?

mbest:
I'm happy with this. I've merged it now.

Issue #358, opened 2012-03-04
long chains of computed observables cause stack overflow


haberman:
I'm writing an accounting app and want to make the "current balance" for each transaction a sum of all previous transactions.  Computed observables are a nice way to model this.  But it appears that update chains are delivered recursively and will overflow the stack.  Could this be made iterative instead so that long chains are possible?

Test program:

``` javascript
var first = ko.observable(0);
var last = first;
for (var i = 0; i < 1000; i++) {
  (function() {
    var l = last;
    last = ko.computed(function() { return l() + 1; })
  })();
}
first(1);
```

PR #1622, opened 2015-09-05
long chains without overflowing the stack fails on FF 10-23

mbest:
The test [Dependent Observable Should allow long chains without overflowing the stack](https://github.com/knockout/knockout/blob/05f4cdcbd2ba653f06ec19240fc100056c5e006b/spec/dependentObservableBehaviors.js#L486) fails [on FF 10-23](https://travis-ci.org/brianmhunt/knockout#L654) with `InternalError: too much recusion`.

Noting: #358 & #359.

### DIFF
--- a/src/subscribables/dependentObservable.js
+++ b/src/subscribables/dependentObservable.js
@@ -29,7 +29,8 @@ ko.computed = ko.dependentObservable = function (evaluatorFunctionOrOptions, eva
         disposeWhen: options["disposeWhen"] || options.disposeWhen,
         domNodeDisposalCallback: null,
         dependencyTracking: {},
-        dependenciesCount: 0
+        dependenciesCount: 0,
+        evaluationTimeoutInstance: null
     };
 
     function computedObservable() {
@@ -50,29 +51,6 @@ ko.computed = ko.dependentObservable = function (evaluatorFunctionOrOptions, eva
             return state.latestValue;
         }
     }
-
-    // evaluatePossiblyAsync is defined in the constructor so that it can access the computed properties
-    // and method using a closure instead of "this", meaning that we can pass the function to "subscribe"
-    // without the need to use "bind". Older versions of Firefox seem to use up a lot more stack space
-    // when using a bound function or using "call". See #1622.
-    var evaluationTimeoutInstance = null;
-    computedObservable.evaluatePossiblyAsync = function () {
-        var throttleEvaluationTimeout = computedObservable['throttleEvaluation'];
-        if (throttleEvaluationTimeout && throttleEvaluationTimeout >= 0) {
-            clearTimeout(evaluationTimeoutInstance);
-            evaluationTimeoutInstance = ko.utils.setTimeout(function () {
-                if (computedObservable.evaluateImmediate()) {
-                    computedObservable["notifySubscribers"](state.latestValue);
-                }
-            }, throttleEvaluationTimeout);
-        } else if (computedObservable._evalDelayed) {
-            computedObservable._evalDelayed();
-        } else {
-            if (computedObservable.evaluateImmediate()) {
-                computedObservable["notifySubscribers"](state.latestValue);
-            }
-        }
-    };
 
     computedObservable[computedState] = state;
     computedObservable.hasWriteFunction = typeof writeFunction === "function";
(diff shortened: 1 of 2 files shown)


mbest:
Okay. So this is one way to fix this issue. This more-or-less mirrors what I did originally (688a567): reducing the function call depth for each iteration and eliminating the use of `call` and `bind`.

src/subscribables/dependentObservable.js
```diff
@@ -52,6 +51,29 @@ ko.computed = ko.dependentObservable = function (evaluatorFunctionOrOptions, eva
         }
     }
 
+    // evaluatePossiblyAsync is defined in the constructor so that it can access the computed properties
+    // and method using a closure instead of "this", meaning that we can pass the function to "subscribe"
+    // without the need to use "bind". Older versions of Firefox seem to use up a lot more stack space
+    // when using a bound function or using "call". See #1622.
+    var evaluationTimeoutInstance = null;
+    computedObservable.evaluatePossiblyAsync = function () {
```
brianmhunt: This undoes the performance benefits of #1841; in most cases a closure will be more memory and slower than a function on the prototype chain, notably with newer browsers.

src/subscribables/dependentObservable.js
```diff
@@ -52,6 +51,29 @@ ko.computed = ko.dependentObservable = function (evaluatorFunctionOrOptions, eva
...
+        var throttleEvaluationTimeout = computedObservable['throttleEvaluation'];
+        if (throttleEvaluationTimeout && throttleEvaluationTimeout >= 0) {
+            clearTimeout(evaluationTimeoutInstance);
+            evaluationTimeoutInstance = ko.utils.setTimeout(function () {
```
brianmhunt: Would it make sense if `setTimeout` was replaced with `ko.tasks.schedule`?

src/subscribables/dependentObservable.js
mbest: If we're okay with the test failing in older FF, then we can simply ignore these changes and possibly skip the test for those browsers. After all, the test already failed with version 3.3.0.

src/subscribables/dependentObservable.js
mbest: No, this needs a delay, whereas `ko.tasks.schedule` acts without a delay.

brianmhunt: Got it, thanks.

src/subscribables/dependentObservable.js
brianmhunt: I'm okay with that, and as you say, going forward the use of deferred updates makes this issue go away.  @SteveSanderson ?

I think then the answer might be to just reduce the number of iterations to 100.  Legacy FF should pass with that (I think it succeeds with 130 or so, but not positive– if we go this route, I'll fix it up).

mbest:
I've made a further commit to undo some of my previous changes. This keeps the logic of moving the "change" notification to `evaluatePossiblyAsync`, reducing a couple layers of function calls from the recursion. This still requires reducing the depth value in the test to make it pass but allows for a greater depth than doing nothing.

brianmhunt:
Thanks @mbest – The updated PR looks good to me –– bearing in mind the caveat that I haven't stared at this enough to really understand any potential side-effects or pitfalls.

mbest:
Thanks. I'm happy with it too. I've merged it now.

brianmhunt:
I merged the above into the 1360-gulp branch, but [as you can see](https://travis-ci.org/knockout/knockout/builds/82047980) there is still an odd continuing failure:

> Firefox 38.0.0 (Linux 0.0.0) Dependent Observable Should allow long chains without overflowing the stack FAILED
>	InternalError: too much recursion in /home/travis/build/knockout/knockout/src/subscribables/subscribable.js (line 72)
>	ko_subscribable_fn.notifySubscribers@/home/travis/build/knockout/knockout/src/subscribables/subscribable.js:72:25
>	computedFn.evaluatePossiblyAsync@/home/travis/build/knockout/knockout/src/subscribables/dependentObservable.js:211:17
> 	ko_subscribable_fn.notifySubscribers@/home/travis/build/knockout/knockout/src/subscribables/subscribable.js:72:25
> ... 
> 	ko_subscribable_fn.notifySubscribers@/home/travis/build/knockout/knockout/src/subscribables/subscribable.js:72:25
>	observableFn.valueHasMutated@/home/travis/build/knockout/knockout/src/subscribables/observable.js:46:36
>	observable@/home/travis/build/knockout/knockout/src/subscribables/observable.js:12:17
>	@/home/travis/build/knockout/knockout/spec/dependentObservableBehaviors.js:516:9

Firefox 17 and 24 pass; all three failed [prior builds without this patch](https://travis-ci.org/knockout/knockout/builds/80755508).  Not sure why FF 38 would have this regression.

mbest:
I'll check that out. I had tested it with Firefox 16.

brianmhunt:
Here's [a jsFiddle to show stack depth](http://jsfiddle.net/Z5pLM/82/)

FF 42 has a call stack of 200–262,000; Chrome is 8–20,000.  Haven't tested FF 38 yet.

PR #1870, opened 2015-09-08
Add ko.isPureComputed function and tests

zhaoyu0810:
Added ko.isPureComputed function

### DIFF
--- a/spec/pureComputedBehaviors.js
+++ b/spec/pureComputedBehaviors.js
@@ -5,6 +5,11 @@ describe('Pure Computed', function() {
         expect(ko.isComputed(computed)).toEqual(true);
     });
 
+    it('Should advertise that instances are pure computed', function () {
+        var instance = ko.pureComputed(function () { });
+        expect(ko.isPureComputed(instance)).toEqual(true);
+    });
+
     it('Should require an evaluator function as constructor param', function () {
         expect(function () { ko.pureComputed(); }).toThrow();
     });
--- a/src/subscribables/dependentObservable.js
+++ b/src/subscribables/dependentObservable.js
@@ -440,9 +440,15 @@ ko.isComputed = function (instance) {
     return ko.hasPrototype(instance, ko.computed);
 };
 
+ko.isPureComputed = function (instance) {
+    return ko.hasPrototype(instance, ko.computed) &&
+        instance.getVersion === pureComputedOverrides.getVersion;
+}
+
 ko.exportSymbol('computed', ko.computed);
 ko.exportSymbol('dependentObservable', ko.computed);    // export ko.dependentObservable for backwards compatibility (1.x)
 ko.exportSymbol('isComputed', ko.isComputed);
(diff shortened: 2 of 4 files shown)


src/subscribables/dependentObservable.js
```diff
@@ -440,9 +440,14 @@ ko.isComputed = function (instance) {
     return ko.hasPrototype(instance, ko.computed);
 };
 
+ko.isPureComputed = function (instance) {
+    return (instance && instance._state && instance._state.pure && ko.isComputed(instance)) || false;
```
brianmhunt: This strategy only works in debug builds, since `_state` will be a Symbol.

A better strategy may be to check if the pure functions are the same as the regular e.g. `instance.getVersion` is the same as the regular `ko.computed`/subscribable version.  (the overloaded version is here: https://github.com/zhaoyu0810/knockout/blob/master/src/subscribables/dependentObservable.js#L407)

zhaoyu0810: Hi Brian,

Thank you very much! How about below code?

```javascript
ko.isPureComputed = function (instance) {
    if (!instance) {
        return false;
    }

    return instance.getVersion === pureComputedOverrides.getVersion && ko.isComputed(instance);
}
```

brianmhunt: Awesome. Maybe it can condense it with e.g.

```javascript
ko.isPureComputed = function(instance) {
   return ko.hasPrototype(instance, ko.computed) &&
     instance.getVersion === pureComputedOverrides.getVersion
}
```

This avoids the overhead of an extra function call to `ko.isComputed`, and we can get rid of the extra `!instance` test since it's subsumed by `hasPrototype`.

zhaoyu0810: Agree. Thank you!

brianmhunt:
I'm ok with this. @knockout/owners ?

spec/pureComputedBehaviors.js
```diff
@@ -5,6 +5,11 @@ describe('Pure Computed', function() {
         expect(ko.isComputed(computed)).toEqual(true);
     });
 
+    it('Should advertise that instances are pure computed', function () {
```
brianmhunt: I think there should also IMHO be assertions that the `ko.isPureComputed` fails for regular computeds and observables.

zhaoyu0810: Thanks Brian. I added tests for computeds and observables.

brianmhunt:
Thanks @zhaoyu0810 - looks good to me.

SteveSanderson:
Thanks @zhaoyu0810! Since @brianmhunt has also looked at this, I've merged it.

I made a slight tweak to the implementation of `isPureComputed` so that it would still work if some future subclass of pure computed overrode the `getVersion` method. Hopefully it still satisfies your requirements.

zhaoyu0810:
Thank you very much! @brianmhunt , @SteveSanderson

Issue #1759, opened 2015-04-09
virtual elements broken in chrome


frankabbruzzese:
When a virtual element is inserted after a `<table>` tag its closing tag appears misplaced. In fact chrome adds a tbody tag that overlaps just after the opening comment, but closes the tbody after the closing comment. Thus, the closing comment is lost withing the tbody.

I imagine that there is no feasible solution to this issue, but it should be mentioned in the documentation at least

brianmhunt:
@frankabbruzzese Thanks for the report.

It seems like some of your markup was clobbered, so I cannot tell what the issue is. It may help if you wrap your example in <code>```</code>

Cheers

frankabbruzzese:
Put the following code into an html page, and open it with chrome:
```
<!DOCTYPE html>
<html>
<head> 
</head>
<body>
<table  data-bind="foreach: Matrix">
<!-- ko enableParsing: true, validation: true -->
	<tr  data-bind="foreach: $data">
	<!-- ko enableParsing: true, validation: true -->
            <td>
               <span data-bind="text: Name" ></span> 
            </td>
        <!-- /ko -->
	</tr>
<!-- /ko -->
</table>
</body>
</html>
```
You don't need to include knockout, or to redefine my custom binding, since the problem is the browser adding a tbody tag that breaks the the outermost "enableParsing" virtual element.

If you select " inspect element" in the right click menu you can see how the browser rewtite the above markup:
![commentbreaking](https://cloud.githubusercontent.com/assets/614976/7084884/f82d5ed0-df71-11e4-916d-e2b526bd9ef8.PNG)

As you can see while the opening tag is left out of the tbody the closing tag is put inside the <tbody>.
This depends of the browser ignoring the semantict of ko comments. Similar errors may occur each time a browser insert automatically a tag. So I suppose the problem appears also in other browser and in other circumnstances.

There is no general solution...other than avoiding virtual elements. Luckily, in each case the code may be fixed by adding manually the tag automatically inserted by the browser,...but in the proper way :)

I think, the above problem should be mentioned in the documentation of virtual elements, together with the way to fix it...otherwise people might go crazy when a similar error occurs.

IanYates:
:+1: for getting this into the docs - if it helps save a newbie some hair-pulling time then it's worth it.  
Aside: I remember running into this sort of problem when I started trying the punches library...  Browsers just completely eat plain text like {{if: xyz}} in places like <table> tags.

brianmhunt:
[...]

Issue #914, opened 2013-04-08
subscribe should not only return the new value, but also the previous value


Topener:
Currently, subscribe only returns current value:

```
var.subscribe(function(newVal) {} );
```

However, in my opinion, it should return both the previous and new/current value

```
var.subscribe(function(newVal, oldVal) {} );
```

daedalus28:
I know this is an old ticket, but I just want to comment that @mbest's solution is great but doesn't let you specify the this context. This slight modification does - just adds a context parameter and uses call instead of directly invoking the callback:

```js
ko.subscribable.fn.subscribeChanged = function (callback, context) {
    var savedValue = this.peek();
    return this.subscribe(function (latestValue) {
        var oldValue = savedValue;
        savedValue = latestValue;
        callback.call(context, latestValue, oldValue);
    });
};
```

brianmhunt:
It feels like, semantically, if this were to be implemented then it should be exposed as a type of event e.g. `subscribe(callback, callbackTarget, 'delta')`.

krnlde:
`delta` is a bit misleading, since it is no delta, but two distinguishable values. Something like `full` or `verbose` sounds exactly like what it's doing.

I hope to bump this issue a bit to finally implement it in the ko-core.

Cheers

cervengoc:
Wouldn't it be simple to just add this feature to the default change notification, for example as a second parameter (to keep compatibility)? I wonder wether it adds a significant memory overhead. It would be so much more simple and intuitive.

krnlde:
This would be the bestestest.

mblandfo:
https://jsfiddle.net/tjzo124x/17/

If you set an observable to a simple value like 1 and it was already set at 1, it won't fire.
But, if you set it to a moment from momentjs, and then you set it to an equivalent, cloned moment, it will fire. In fact, if you set it to the exact same moment object, it will fire.

In fact, it seems if you set an observable to an object that === the last value, it will still fire. Anyway, I modified the above function to fix this:

```
subscribeChanged = function(obs, func) {
    var savedValue = obs.peek();
    return obs.subscribe(function (latestValue) {
        var oldValue = savedValue;
        savedValue = latestValue;
        // ko.subscribe doesn't know how to compare moments
        var areEqual = oldValue === latestValue || oldValue && oldValue.isSame && oldValue.isSame(latestValue);
        if (!areEqual) {
            func.call(latestValue, oldValue);
        }
    });
};
```

[...]

PR #1873, opened 2015-09-10
Documentation updates for 3.4.0

mbest:
All of the 3.4.0 updates can go in here (although large updates might be better handled in a sub-branch first).

  - [x] Add note about `ko.options.useOnlyNativeEvents` to some or all pages that describe features that use events: `event`, `click`, `submit`, `value`, `checked`, `textInput`, `selectedOptions`, `hasFocus`.
  - [x] `ko.onError` to catch errors from async code. I'm not sure if needs its own page or should be added to the appropriate pages (`event` binding, components, rate-limit, deferred updates, etc.).
  - [x] Observable Array: `sort` and `reverse` return the `observableArray` and not the `Array`.
  - [x] New page for deferred updates. Possible links from other pages (such as where rate-limiting is already mentioned).
  - [x] Update notes on component `synchronous` option. (The option is not needed anymore to prevent multiple DOM reflows.)
  - [x] `ko.tasks`
  - [x] `ko.isPureComputed`
  - [x] Add Edge to supported browsers.


brianmhunt:
Should we also document `ko.tasks`?

brianmhunt:
Just for ease of reference, two cases should probably be documented because they keep coming up:

1. #1759  --  the hoisting of comments out of `<table>` tags
2. #914 -- `beforeChange` subscription or a change event with the new + old value

Not that these need to go in for 3.4, just thought to note them while we're at it. :grin:

mbest:
> Should we also document ko.tasks?

Yeah, I've added it to the list and am working on it now.

mbest:
I've finally put together the deferred updates documentation.

mbest:
The only Doc item left is `ko.onError`. Any takers, @SteveSanderson, @brianmhunt, @rniemeyer?

brianmhunt:
I'm happy to write the `onError` docs but I can't commit to a timeline right now (probably 2 weeks, give or take)– if someone else wants to pick it up sooner that's super, too. :)

rniemeyer:
I can take `onError` docs

mbest:
Great, @rniemeyer!

brianmhunt:
:+1: @rniemeyer

mbest:
Thanks, Ryan. If there are no complaints, we can consider the 3.4.0 documentation complete.

PR #1876, opened 2015-09-10
Added support for an 'All' event

daedalus28:


daedalus28:
Sorry about all those commits - that's what I get for using the github editor instead of pulling the repo locally. The issues were mostly related to changes that were need when converting this from my local patch strategy to bringing it into the ko core plus a few spacing rules knockout enforces. All tests should be passing now, but I'll happily rebase this to cleanup the history a bit.

src/subscribables/subscribable.js
```diff
@@ -75,6 +79,20 @@ var ko_subscribable_fn = {
                 ko.dependencyDetection.end(); // End suppressing dependency detection
             }
         }
+        // Notify subscribers to the ALL_EVENT event, but not if that was the event that was fired (prevent double events)
+        if (this.hasSubscriptionsForEvent(allEvent) && event != allEvent) {
```
stalniy: Why don't you use notifysubscribers here?
If(event !== all) 
   NotifySubscribers("all")

No need in manual loop, dep detection etc

daedalus28: Because this _is_ the code for `notifySubscribers` - it's based entirely off of the existing implementation. Unless I'm missing something, calling again would result in an infinite loop

src/subscribables/subscribable.js
```diff
@@ -150,8 +168,10 @@ var ko_subscribable_fn = {
 };
 
 ko.exportProperty(ko_subscribable_fn, 'subscribe', ko_subscribable_fn.subscribe);
+ko.exportProperty(ko_subscribable_fn, 'subscribeAll', ko_subscribable_fn.subscribeAll);
 ko.exportProperty(ko_subscribable_fn, 'extend', ko_subscribable_fn.extend);
 ko.exportProperty(ko_subscribable_fn, 'getSubscriptionsCount', ko_subscribable_fn.getSubscriptionsCount);
+ko.exportProperty(ko_subscribable_fn, 'allEvent', allEvent);
```
stalniy: No need to export this

daedalus28: Actually exporting it is useful. If you need to programmatically subscribe to events you may want to subscribe directly to the all event type. It's a covered test case as part of this pull request, actually. SubscribeAll is just a convenience method.

src/subscribables/subscribable.js
mbest: But you're checking `event != allEvent` so that would prevent an infinite loop.

daedalus28: True - but unless I'm missing something, I'd have to publish the original event type because the callback needs to pass it along. It's a core part of the feature - if you subscribeAll you need to know what the event was.

daedalus28: Perhaps we can extend notifySubcribers to take a third param? I'd like to avoid wrapping the value because you can still directly publish on the all event. That would let us pass the original event along and reuse notifySubcribers, but it would change the arity of the function. That could cause issues - for example if people used it as the callback in an _.each to notify events from an object

brianmhunt:
The API that seems most natural would be `subscribe(callback, thisArg, 'all')`

I'm working my head around the substance too, slowly... 😀

daedalus28:
@brianmhunt I agree - the only reason I didn't just call it 'all' was to prevent reserving an event name. While observables publish specific events like change, beforeChange, etc, subscribables don't have any special reserved events to my knowledge. But then again I'm basically using knockout subscribables as an EventEmitter which is probably a bit of an edge case, so maybe it doesn't matter

brianmhunt:
This may or may not be worth taking into consideration :grin:  –

This feels like it is close to the inner-loop, so it could knock on performance.  It also feels like a bit of an edge case (i.e. not a huge demand), and it's not immediately clear looking at the code how this would be used in the ordinary case.  So that said, I think we can open up an innocuous hook that make this possible without impacting performance or having code that confuses me (not a high standard, mind you :wink:).

@daedalus28 – I feel like, but may be mistaken, that you could you do everything you need with a hook at https://github.com/daedalus28/knockout/blob/patch-1/src/subscribables/subscribable.js#L82 like this:

```javascript
if (typeof this.onNotifySubscribers === 'function')
    this.noNotifySubscribers.call(this, valueToNotify, event)

// And somewhere, you set this
ko.subscribable.fn.onNotifySubscribers = function (valueToNotify, event) {}
```

That said, is KO exposing enough to make your end-case possible?  The only thing that might not be exposed, that I see offhand, is `isDisposed`.

Just some food for thought.

src/subscribables/subscribable.js
stalniy: So, do you think copy/paste is better for this case?
Can you pass event name in another way?
Do you think it's ok that "all" has different event name in subscribe callback? What's about consistency?
How does other libraries implement similar functionality?

daedalus28:
@brianmhunt I like the idea of a generic onNotifySubscribers hook. It would probably eliminate the possibility for an 'all' event, but just having a subscribeAll function is acceptable for my use case. To use the hook, I'd need to also deal with subscription management, disposal etc. The biggest question I have is what exactly what the `ko.dependencyDetection.begin` and  `ko.dependencyDetection.end` functions do (and I can't just blindly invoke them as they are private). Can you or @mbest, @SteveSanderson, or @rniemeyer explain what they're for, and how I'd be able to accomplish the same thing with the public API surface?

Also, since it seems like your thinking is pointing towards this hook, does that mean knockout is not interested in supporting a subscribe all in the core? I'm ok with that, but it does seem like a concern for the knockout core. But if that's the consensus, I'll open a new pull request for the hook and close this one.

src/subscribables/subscribable.js
daedalus28: The code duplication is definitely unfortunate and is something I'd refine before this was merged. For consistency, if `notifySubscribers` passed a third callback argument, we could swap the event and third param in the case of a notify subscribers with an 'all' event if you think that would be less confusing. But there's no way around passing the original event - if you subscribe to all events, you need to know what the original event was so you can react accordingly.

Either way, it seems like a onNoify hook is the direction this is heading so it's probably a bit of a moot point.

daedalus28:
Also, if we do go the route of `onNotifySubscribers`, it seems like it would be a relatively small change. Is it something that could make it in the 3.4 release?

brianmhunt:
Thanks @daedalus28 .

```javascript
ko.dependencyDetection.begin()
thisArg.things(param)
ko.dependencyDetection.end()
/// is the same as
ko.ignoreDependencies(things, thisArg, [param])
```

- Docs: http://knockoutjs.com/documentation/computed-dependency-tracking.html
- Code: https://github.com/knockout/knockout/blob/master/src/subscribables/dependencyDetection.js#L39-L46

Insofar as a change is warranted, I personally feel the hook is preferable because it does not substantially increase the API or support surface of the library (unless I'm missing something).  Innocuous, in other words.  But I am open. :)

daedalus28:
Ok - I'd also want to expose ko.subscription as I'd have my own array of subscriptions that could get disposed (and it would be unfortunate to duplicate that code). Also, a hook in the init function would be nice to set up `instance._subscriptionsToAll`, although I could also patch that in by caching `ko.subscribable.fn.init`.

With the hooks in place, my implementation would look like this:

```js
ko.subscribable.fn.onInit = function(instance) {
    instance._subscriptionsToAll = [];
};
ko.subscribable.fn.onNotifySubscribers = function(valueToNotify, event) {
    ko.ignoreDependencies(function() {
        ko.utils.arrayForEach(this._subscriptionsToAll, function(subscription) {
            if (!subscription.isDisposed) // This is why I need ko.subscription and not just an array of callbacks
                subscription.callback(valueToNotify, event);
        });
    }, this);
};
ko.subscribable.fn.subscribeAll = function(callback, callbackTarget) {
    var boundCallback = callbackTarget ? callback.bind(callbackTarget) : callback;
    var subscription = new ko.subscription(this, boundCallback, function() {
        ko.utils.arrayRemoveItem(this._subscriptionsToAll, subscription);
    });
    this._subscriptionsToAll.push(subscription);
    return subscription;
};
```

Apart from a little bit of redundancy, my main concern now with this approach is just that it's separate from things like `getSubscriptionCount`, `hasSubscriptionForEvent`, etc and doesn't have the hook for `afterSubscriptionRemove`, `beforeSubscriptionAdd`, etc - all the stuff that came free before.


Another option instead of exposing `ko.subscription` is to make _subscriptions public (which I believe is another open issue) so I can hook into the existing functions with a similar 'all' event like this pull request:

```js
ko.subscribable.fn.allEvent = 'all';
ko.subscribable.fn.onNotifySubscribers = function(valueToNotify, event) {
    if (event == ko.subscribable.fn.allEvent)
        return;
    ko.ignoreDependencies(function() {
        ko.utils.arrayForEach(this._subscriptions[ko.subscribable.fn.allEvent], function(subscription) {
            if (!subscription.isDisposed) // This is why I need ko.subscription and not just an array of callbacks
                subscription.callback(valueToNotify, event);
        });
    }, this);
};
ko.subscribable.fn.subscribeAll = function(callback, callbackTarget) {
   return this.subscribe(callback, callbackTarget, ko.subscribable.fn.allEvent);
};
```



However, if we can extend notify subscribers to pass the event as the second parameter and split out `notifySubscribers` into `notify` which would be the current implementation and `notifySubscribers` which would be the current implementation plus the conditional call to `onNotifySubscribers`, we don't need to expose anything and I can completely reuse the current `notifySubscribers` implementation
```js
ko.subscribable.fn.allEvent = 'all';
ko.subscribable.fn.onNotifySubscribers = function(valueToNotify, event) {
    if (event != ko.subscribable.fn.allEvent)
        this.notify(valueToNotify, event);
};
ko.subscribable.fn.subscribeAll = function(callback, callbackTarget) {
   return this.subscribe(callback, callbackTarget, ko.subscribable.fn.allEvent);
};
```

daedalus28:
@brianmhunt To clarify, here is my proposal:

1. Rename `notifySubscribers` to `notify`
2. Add `notifySubscribers`, which calls `notify` and then calls `onNotifySubscribers` if it exists and is a function (just like your comment).
3. Make notify additionally pass the event type as the second parameter to the callback (at this line https://github.com/daedalus28/knockout/blob/patch-1/src/subscribables/subscribable.js#L76)

daedalus28:
Sorry, just realized that last option would not actually trigger subscriptions to the all event. What I'd want to do then is not actually bother to split out `notifySubscribers` into `notify` - I'd just make it pass the rest of it's parameters to the callback function (it would have been prettier if notifySubscribers took it's data last instead of having the event at the end):

```js
subscription.callback.apply(subscription.callbackTarget, [valueToNotify].concat(arguments.slice(2));
```

The only other missing piece is to cache the subscription `callbackTarget` at subscribe time, which would be something like this at https://github.com/daedalus28/knockout/blob/patch-1/src/subscribables/subscribable.js#L50:
```js
subscription.callbackTarget = callbackTarget;
```


 With that in place, my implementation would look like this:
```js
ko.subscribable.fn.allEvent = 'all';
ko.subscribable.fn.onNotifySubscribers = function(valueToNotify, event) {
    if (event != ko.subscribable.fn.allEvent)
        this.notifySubscribers(valueToNotify, ko.subscribable.fn.allEvent, event);
};
ko.subscribable.fn.subscribeAll = function(callback, callbackTarget) {
   return this.subscribe(callback, callbackTarget, ko.subscribable.fn.allEvent);
};
```

and the only modifications to be made to the ko core are two lines of code to to pass along additional arguments to the callback and cache the callback context.

daedalus28:
If you don't want to support an arbitrary number of additional parameters, we could simplify the implementation to just pass the third parameter to `notifySubscribers`. It would eliminate the need for caching the context in subscribe and reduce this to a single line change.

stalniy:
Callback is the same story as all event with 1 difference that you can have only one callback. In future devs are forced to write smth like 
Oldcallback = onNotify
Subscription.onNotify = ->
  Old.apply
  NewOne.apply

Instead of
subscription.subscribe(fn, null, all)

Which seems more natural and straightforward (consistent with other Lina). Callback is usually a bad choice if you are not doing FP in JS. Also you will need to solve issues with disposing custom subscription.

daedalus28:
@stalniy that's true, but there are other extension points in knockout that have the same behavior (like subscribable's `beforeSubscriptionAdd` and `afterSubscriptionRemove`).

Regarding subscription disposal, if you read through the discussion I think I've addressed that in all three of the options I've proposed (which the current pull request's changes does not reflect). The latest option reuses the ko methods so subscription disposal and utilities like `getSubscriptionCount` should work out of the box.

The latest proposal would also support `subscription.subscribe(fn, this, 'all')` - `subscribeAll` is once again just some sugar over the subscribe method.

daedalus28:
@brianmhunt does adding support for additional values sent to callbacks from `notifySubscribers` seem like a reasonable change? It can either be fully dynamic by passing an arguments slice and caching callbackTarget, or we can just add a single third parameter that gets passed along. I either need that or to make `_subscriptions` public.

daedalus28:
Closing this since it's probably pretty stale, there doesn't seem to be much interest, and my team and I have generally all moved on anyway and so likely wouldn't have the bandwidth to support this change.

PR #1879, opened 2015-09-14
Fix link extension.

tabone:
Fixed link for "How dependency tracking works" page inside "Pure computed observables" doc page.

screen shot 2015-09-14 at 18 07 51


brianmhunt:
Thanks @tabone!

tabone:
:+1: My pleasure :)

PR #1883, opened 2015-09-21
Fixing utils.domNodeDisposal.cleanExternalData export

moonpyk:
As stated in the documentation (http://knockoutjs.com/documentation/custom-bindings-disposal.html), cleanExternalData is part of ko.utils.domNodeDisposal public API.
The function was not correctly exported, with risks of being mangled during the build step. This patch makes sure the function won't be lost.


mbest:
`cleanExternalData` is always exported because it's defined using quotes.

moonpyk:
I was not aware of this, thanks for the quick reply !

PR #1880, opened 2015-09-24
Removing jQuery HTML parsing by default is breaking

mbest:
I was originally hopeful that we could stop piggybacking on jQuery's HTML parser, at least for modern browsers. But after attempting to upgrade my current project to KO 3.4.0, I can see it's a bit more complicated than that!

Some things that the `$.parseHTML` handles that KO's built-in parser doesn't (even on modern browsers):

 * `<col>` and `<colgroup>` elements and probably others like `<legend>`
 * Markup with special elements like `<thead>` prefixed by comments or whitespace, e.g., `<!-- ko something --><thead>...</thead>`

For KO 3.4.0 to ship without pretty bad breaking changes here, I think we need to:

 * Go back to using `$.parseHTML` when available by default (possibly with some `ko.options` flag to disable that, but it shouldn't be the default)
 * Or, extend KO's native parser to handle the full set of edge cases that jQuery does

Looking at jQuery's source for `$.parseHTML`, I don't feel too enthusiastic about putting that much stuff into KO, so am leaning towards "go back to using jQuery's parser again". I know that jQuery's parser doesn't handle certain bizarrely-named custom elements (e.g., `<th-mything>`) but that's much more of an edge case than standard elements like `<col>` that jQuery does support.

Thoughts?

### DIFF
--- a/spec/lib/loadDependencies.js
+++ b/spec/lib/loadDependencies.js
@@ -40,7 +40,7 @@
         if (url) {
             var shouldInclude = getParam(name);
             if ((dependency.include || shouldInclude) && shouldInclude !== "0" && shouldInclude !== "false") {
-                if (shouldInclude && /^[0-9]+\.[0-9.]+$/.test(shouldInclude)) {
+                if (shouldInclude && shouldInclude !== "1" && shouldInclude !== "true") {
                     url = url.replace(dependency.versionString || 'latest', shouldInclude);
                 }
                 jasmine.addScriptReference(url);
--- a/spec/parseHtmlFragment.js
+++ b/spec/parseHtmlFragment.js
@@ -11,10 +11,10 @@ describe('Parse HTML fragment', function() {
 
     ko.utils.arrayForEach(
     [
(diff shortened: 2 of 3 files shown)


spec/lib/loadDependencies.js
```diff
@@ -40,7 +40,7 @@
         if (url) {
             var shouldInclude = getParam(name);
             if ((dependency.include || shouldInclude) && shouldInclude !== "0" && shouldInclude !== "false") {
-                if (shouldInclude && /^[0-9]+\.[0-9.]+$/.test(shouldInclude)) {
+                if (shouldInclude && shouldInclude !== "1" && shouldInclude !== "true") {
                     url = url.replace(dependency.versionString || 'latest', shouldInclude);
```
mbest: This change allows the use of `jquery=git` to test with the 3.0.0-pre version.

mbest:
So this should be ready to merge. Hopefully this is the last code change for 3.4.0.

PR #1893, opened 2015-09-30
Setting hasFocus to false in IE clears focus from all other elements

mbest:
The behavior of the ```hasFocus``` binding has broken in IE since version ```3.3.0```. Specifically, assigning a value of ```false``` to the binding when the element is rendered now clears focus from all other elements, because it invokes ```element.ownerDocument.body.focus()```.

Here is a very simple example:

``` html
<p tabindex="0" data-bind="hasFocus: true">This should have focus initially.</p>
<p tabindex="0" data-bind="hasFocus: false">This should not have focus.</p>
```

``` javascript
ko.applyBindings()
```

In Chrome and Firefox, the first paragraph starts as focused. In IE, however, neither paragraph is focused.

Here is a JSFiddle of the simple case: http://jsfiddle.net/ThomasMichon/ef3u690p/ (look for a red outline around the focused paragraph).
Here is a more thorough example showing how this behavior disrupts keyboard usage: http://jsfiddle.net/ThomasMichon/5toqrqn3/ (note how the selected item loses focus when a new item is rendered).

My understanding is that passing ```false``` to the ```hasFocus``` binding is simply supposed to blur that one element, not clear focus from all elements. I think we just need just need to ensure that the bound element is the current value of ```document.activeElement``` before using body focus as a way to blur it.

### DIFF
--- a/spec/defaultBindings/hasfocusBehaviors.js
+++ b/spec/defaultBindings/hasfocusBehaviors.js
@@ -84,4 +84,15 @@ describe('Binding: Hasfocus', function() {
         model.isFocused(null);
         expect(didBlurAgain).toEqual(false);
     });
+
+    it('Should not cause unrelated items to lose focus when initialized with false', function () {
+        // See #1893
+        testNode.innerHTML = '<input data-bind="hasFocus: true" value="This should be focused initially" /><input data-bind="hasFocus: false" value="This should not be focused" />';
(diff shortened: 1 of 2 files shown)


mbest:
I've attached a fix and a test based on your example.

mbest:
Merged.

PR #1903, opened 2015-10-13
New beforeRemove behavior can break non-removed items

mbest:
While re-testing the examples included with #1256, I discovered that when removing items from an array, following items might be "cleaned" so that their bindings no longer worked. This happens when the nodes for an item aren't all removed at the same time and `fixUpContinuousNodeArray` exits with the "undefined scenario".

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -255,8 +255,11 @@ ko.utils = (function () {
             // Rules:
             //   [A] Any leading nodes that have been removed should be ignored
             //       These most likely correspond to memoization nodes that were already removed during binding
-            //       See https://github.com/SteveSanderson/knockout/pull/440
-            //   [B] We want to output a continuous series of nodes. So, ignore any nodes that have already been removed,
+            //       See https://github.com/knockout/knockout/pull/440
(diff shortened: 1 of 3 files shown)


mbest:
@rniemeyer Could you please review these changes?

src/utils.js
```diff
@@ -275,8 +282,10 @@ ko.utils = (function () {
                     while (current !== last) {
                         continuousNodeArray.push(current);
                         current = current.nextSibling;
-                        if (!current) // Won't happen, except if the developer has manually removed some DOM elements (then we're in an undefined scenario)
-                            return;
+                        if (!current) { // Won't happen, except if the developer has manually removed some DOM elements (then we're in an undefined scenario)
```
mbest: With the change above (new rule B), we can probably remove this check since the last node should always be there. What do you think, @rniemeyer?

rniemeyer:
@mbest - I'lll review as soon as I can

rniemeyer:
@mbest this looks good and I agree that it should be safe to remove the check that you commented on.

mbest:
Thanks, @rniemeyer. I've merged this now.

Issue #1109, opened 2013-09-19
Computed update bug


blazkovicz:
1) computed depends on observable properties of items in observable array
2) computed initialized
3) we add new items into observable array
result) properties of those new items are not watched by computed
knockout version 2.3.0

mbest:
This topic has come up again while working on 3.4.0. I noticed that the example fiddle works with 3.4.0rc: http://jsfiddle.net/ybmHq/4/

PR #1905, opened 2015-10-13
Endless recursion possible in evaluateImmediate

mbest:
It looks like ```evaluateImmediate``` was changed in PR #1622 so that the calls to ```notifySubscribers``` are now made outside the window during which ```state.isBeingEvaluated``` is ```true```. This is a breaking change because it now produces stack overflows if two computed observables have a circular dependency.

**Update**: I've isolated this to the case where a ```computed``` returns an ```object``` on its evaluation, which is then used to trigger a circular dependency.

### DIFF
--- a/spec/dependentObservableBehaviors.js
+++ b/spec/dependentObservableBehaviors.js
@@ -502,8 +502,8 @@ describe('Dependent Observable', function() {
     it('Should allow long chains without overflowing the stack', function() {
         // maximum with previous code (when running this test only): Chrome 28: 1310, IE 10: 2200; FF 23: 103
         // maximum with changed code: Chrome 28: 2620, +100%, IE 10: 4900, +122%; FF 23: 267, +160%
-        // (per #1622, max depth reduced to pass tests in older FF)
-        var depth = 175;
+        // (per #1622 and #1905, max depth reduced to pass tests in older FF)
+        var depth = 100;
         var first = ko.observable(0);
         var last = first;
         for (var i = 0; i < depth; i++) {
--- a/src/subscribables/dependentObservable.js
+++ b/src/subscribables/dependentObservable.js
@@ -195,24 +195,19 @@ var computedFn = {
     },
     evaluatePossiblyAsync: function () {
         var computedObservable = this,
-            state = computedObservable[computedState],
             throttleEvaluationTimeout = computedObservable['throttleEvaluation'];
         if (throttleEvaluationTimeout && throttleEvaluationTimeout >= 0) {
-            clearTimeout(state.evaluationTimeoutInstance);
-            state.evaluationTimeoutInstance = ko.utils.setTimeout(function () {
-                if (computedObservable.evaluateImmediate()) {
-                    computedObservable["notifySubscribers"](state.latestValue);
-                }
+            clearTimeout(this[computedState].evaluationTimeoutInstance);
+            this[computedState].evaluationTimeoutInstance = ko.utils.setTimeout(function () {
+                computedObservable.evaluateImmediate(true /*notifyChange*/);
             }, throttleEvaluationTimeout);
         } else if (computedObservable._evalDelayed) {
             computedObservable._evalDelayed();
         } else {
-            if (computedObservable.evaluateImmediate()) {
-                computedObservable["notifySubscribers"](state.latestValue);
-            }
+            computedObservable.evaluateImmediate(true /*notifyChange*/);
         }
     },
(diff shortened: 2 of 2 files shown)


mbest:
I've attached the code to revert the changes from #1622, except to further reduce the depth value in the test. @brianmhunt Can you review?

ThomasMichon:
I would argue that the recursion suppression when computed observables have circular dependencies is one of the primary features of Knockout. It's certainly what makes it very easy to write simple two-way data-bindings between view models. You do not have to write your own equivalent of ```isBeingEvaluated``` or some form of ```eventId``` tracker just to prevent circular dependencies, which will inevitably occur as people link more components together.

One thing which surprises me is that only one of my unit tests caught this, and it involved code using the ```knockout-projections``` plugin and the way it handles mapping evaluations. I was actually able to write a workaround in my code, but I don't know where else in my codebase this change will have an effect.

brianmhunt:
@mbest - I think so. Hope to report back today.

mbest:
@ThomasMichon Originally the recursion check was meant to protect a bit of Knockout code that couldn't properly handle computed recursion (#387). Prior to that release (2.1.0), Knockout didn't prevent recursion and we got issues like [this](http://stackoverflow.com/q/9750533/1287183). After 2.1.0, I found only one complaint about the new "feature", #1109, and interestingly the change in 3.4.0rc fixes that example.

I guess there's not necessarily a "better" approach, but it's certainly correct that removing the "feature" from Knockout is a breaking change.

brianmhunt:
@mbest I've merged the changes into #1868. Good and bad news. The good news is the old FF versions now pass. The bad news is that [FF 38 fails](https://travis-ci.org/#L420). Weird.

I also noted that jQuery3 fails a test (but their `-compat` version works) with PhantomJS 1.9.8. I presumed that is a PhantomJS problem, but it also fails on Chrome.  The result in question is:

> Chrome 46.0.2490 (Mac OS X 10.11.0) DOM node disposal If jQuery is referenced, should clear jQuery data when a node is cleaned FAILED

>	Expected {  } to be undefined.

>	Error: Expected {  } to be undefined.

>	    at null.<anonymous> (/.../knockout/spec/domNodeDisposalBehaviors.js:94:50)

Not sure if jQuery will clean that up before beta/release of 3.0; we probably want to delay a workaround until we know what they're doing.

Otherwise I think it looks okay — though to get rid of the recursion we may have to start pulling subscribers like in #359 (though I'm not sure that's the best tact).

mbest:
Thanks, Brian. I tested this with FF 38 and didn't get any errors. I've merged it now.

brianmhunt:
Very good, thanks Michael. I will have to investigate what might be causing the issue with the automated tests, since I am skeptical that it's FF 38.

nmehlei:
If I switch my solution from ko 3.3 with deferred plugin to ko 3.4 (latest master branch commit) with native deferral I still get 'Too much recursion' after processing 5000 task groups.
I hoped that the latest commit would change that. I did not have the time to find out which of my computed's is at fault here (my solution is very big) but obviously the deferred plugin did support my use case whereas the native deferral does not :/

nmehlei:
Well, upon further research into my codebase the infinite recursion only occured in a single area of my code which was not written especially well. After slight optimizations all worked fine, so my previous comment was a bit premature. As it looks right now, my codebase is working fine with native deferral and I will happily upgrade to the release version of 3.4 once it is released.

ThomasMichon:
Given that I have been able to work around this by being a little more careful with the return values from computed observables, you may not need to rush to check this in. I was initially worried that the ```knockout-projections``` plugin by @SteveSanderson was adversely affected by these changes since that is where I was experiencing infinite recursion. However, it seems the issue is confined to returning values of type ```object``` in cases where you cannot override the default ```equalityComparer``` to be less chatty and you have a circular dependency between observables. An example of that happens to be when you use the ```mappingWithDisposeCallback``` of the ```knockout-projections``` plugin.

mbest:
@ThomasMichon Are you suggesting that it might be better to leave in the change that caused this issue for you?

ThomasMichon:
@mbest I am suggesting that if you believe that the changes in question are good for performance optimization and reducing bugs, I can take the hit with respect to infinite recursion. I still have a workaround with respect to circular dependency detection if I need it, so long as it doesn't occur in an external library. These changes *do* have the potential to break some consumers, if they work the way I did, but they are probably better to have overall.

PR #1910, opened 2015-10-19
typed observables

LukasHechenberger:
In one of my current projects I use **typed observables** to easily validate types of values I store in a database later. I thought you might be interested in adding this to the knockout core.

In this fork i added a **ko.registerType**-method and short implementations of a **Bool**, **Number** and **Int** type to show you how this could look like. These types are available as **ko.observableBool** and so on. Types can be subtypes of another.

Each type has some additional methods and attributes. As an example, ko.observableBool has a toggle method: toggling such an observable can easily be done by adding
```<button data-bind="click: aBoolObservable.toggle"></button>```
to the markup. The view model in this case would simply be
```{ aBoolObservable: ko.observableBool() }```.

I created a [Codepen](http://codepen.io/ls-age/pen/OyOmwQ?editors=101) providing a simple example of all the methods and attributes I added to this fork.

If you were interested in adding this to the knockout core, I would be happy to help you implementing it.

### DIFF
--- a/build/fragments/source-references.js
+++ b/build/fragments/source-references.js
@@ -17,6 +17,7 @@ knockoutDebugCallback([
     'src/subscribables/observableArray.changeTracking.js',
     'src/subscribables/dependentObservable.js',
     'src/subscribables/mappingHelpers.js',
+    'src/subscribables/typedObservables.js',
     'src/binding/selectExtensions.js',
     'src/binding/expressionRewriting.js',
     'src/virtualElements.js',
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by ThomasMichon** [comment]
Would defining a set of extenders still have the effect you desire?

``` javascript
ko.extenders['boolType'] = function (target, options) {
    return ko.utils.extend(target, {
        'toggle': function () {
            target(!target.peek());
        }
    });
};

ko.extenders['numberType'] = function (target, options) {
    return ko.utils.extend(target, {
        'increment', function () {
            target(target.peek() + 1);
        },
        'decrement', function () {
            target(target.peek() - 1);
        },
        // ...
    });
};

// ...
```

Wherever you would create an observable, you could apply the appropriate extender:

``` javascript
var showDetails = ko.observable(false).extend({ boolType: {} });
showDetails.toggle();

var viewCount = ko.observable(0).extend({ numberType: {} });
showDetails.increment();
```

An advantage here is that you can apply the extenders to a ```ko.computed``` in addition to a ```ko.observable```.

I also wonder why we ultimately need to bother adding the methods conditionally per type. Seems one would get a lot more power out of having the utility methods always be present for *any* observable, no matter how it is created. One could just apply attach these methods to ```ko.observable.fn``` and then use them wherever you need them.

Issue #1649, opened 2014-12-12
Support html5 tristate checkboxes


davhdavh:
Please add support for html5 tristate checkboxes.

Here is simple example of how it could be done
https://github.com/davhdavh/knockout/commit/5e11af75c20bc50ef7ecf6b719c616fea754df90

c = ko.observable(false);
     &lt;input type="checkbox" data-bind="checked: c, tristateType: 'cycle'" />
clicking the checkbox will now cycle between false, null and true. and the value is null, the css selector :indeterminate will match on the element, allowing styling.

c = ko.observable(null);
     &lt;input type="checkbox" data-bind="checked: c, tristateType: 'none'" />
clicking the checkbox will now cycle between false, true and only allow the 3rd state programmatically.

http://jsfiddle.net/Cine/4soms7hx/1/

krnlde:
This task has been sitting here for quite a while now. What is the state?

mbest:
I hope to get something into the next version. I don't know when that will be though.

DeanNorth:
Maybe this could be added as a new built in binding handler like this?

```
ko.bindingHandlers.indeterminateValue = {
    update: function (element, valueAccessor) {
        var value = ko.utils.unwrapObservable(valueAccessor());
        element.indeterminate = value;
    }
};
```

Example using the computed writable example with checkboxes
https://jsfiddle.net/deannorth/crqwac12/

PR #1924, opened 2015-11-06
Support 3-state checkboxes

lencharest:
HTML5 allows for a 3-state checkbox, and the ARIA spec defines 3 discrete values (see http://www.w3.org/TR/wai-aria/roles#checkbox). Modern browsers can visualize the indeterminate/mixed state, and it can be set programmatically by setting the `indeterminate` property of the corresponding DOM element (see https://css-tricks.com/indeterminate-checkboxes/). This PR modifies the `checked` binding to support 3-state checkboxes by recognizing the special value 'mixed'. Note that the modifications *do not* inject any `aria-` attributes.

### DIFF
--- a/spec/defaultBindings/checkedBehaviors.js
+++ b/spec/defaultBindings/checkedBehaviors.js
@@ -27,9 +27,19 @@ describe('Binding: Checked', function() {
 
         ko.applyBindings({ someProp: myobservable }, testNode);
         expect(testNode.childNodes[0].checked).toEqual(true);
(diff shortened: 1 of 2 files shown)


brianmhunt:
:+1:

mbest:
See #1649

lencharest:
Does this mean that #1649 is going to be merged in to `master`?

mbest:
I like the implementation suggested in #1649 better as is supports a "cycle" feature.

PR #1710, opened 2015-11-20
Allow Dynamic and Static Classes with CSS Binding

mbest:
I am wanting to apply a CSS binding with both a dynamic and static class. I have been reading around on forums for solutions to this and I guess I am looking for a reason not to find a way to implement this into the library. If it is a principle of the library to not support this, that answer would be fine as well. The only way I could think of handling this would be to allow the CSS binding to take an array that contains strings or objects that then get handled appropriately by the binding to apply the classes.

Thanks

### DIFF
--- a/spec/defaultBindings/cssBehaviors.js
+++ b/spec/defaultBindings/cssBehaviors.js
@@ -102,4 +102,31 @@ describe('Binding: CSS classes', function() {
         observable1("my-Rule");
         expect(testNode.childNodes[0].className).toEqual("unrelatedClass1 my-Rule");
     });
+
+    it('Should be aliased as class as well as css', function() {
+        expect(ko.bindingHandlers['class']).toEqual(ko.bindingHandlers.css);
+    });
+
+    it('Should be able to combine "class" and "css" bindings with dynamic and static classes', function () {
+        // This test doesn't cover cases where the static and dynamic bindings try to set or unset the same class name
+        // because the behavior for that scenario isn't defined.
+
+        var booleanProp = new ko.observable(false);
+        var stringProp = new ko.observable("");
+        testNode.innerHTML = "<div class='unrelatedClass' data-bind='css: { staticClass: booleanProp }, class: stringProp'></div>";
+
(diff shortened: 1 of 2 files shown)


mbest:
I've attached the code changes here.

juuxstar:
While a simple implementation, IMO having two bindings which do the same thing adds confusion (especially in the exact case when you need to use both at once since it doesn't matter which you use for the static and which for the dynamic bindings and thus somehow implies that one binding does something else than the other)

I ran into this issue last year and solved it by slightly tweaking the existing css binding to accept an array of values, which can either be Strings or objects (to support both dynamic and static classes in the same binding):

```
var classesWrittenByBindingKey = '__ko__cssValue';
ko.bindingHandlers['class'] = ko.bindingHandlers['css'] = {
    'update': function (element, valueAccessor) {
        var classes = ko.unwrap(valueAccessor());
        var classString = [];

        if (!(classes instanceof Array)) classes = [ classes ];

        classes.forEach(function(clazz) {
            if (clazz !== null && typeof clazz == 'object') {
                ko.utils.objectForEach(clazz, function(className, shouldHaveClass) {
                    ko.utils.toggleDomNodeCssClass(element, className, ko.unwrap(shouldHaveClass));
                });
            }
            else classString.push(String(clazz || '')); // Make sure we don't try to store or set a non-string value
        });

        if (classString.length > 0) {
            classString = classString.join(' ');
            ko.utils.toggleDomNodeCssClass(element, element[classesWrittenByBindingKey], false);
            element[classesWrittenByBindingKey] = classString;
            ko.utils.toggleDomNodeCssClass(element, classString, true);
        }
    }
};
```

so you would thus call it like this:
```
<div data-bind="css : [ 'static-class-' + 1, { 'visible' : isVisible() } ]"></div>
```

mbest:
Perhaps a solution to the confusion would be to make `class` only work for dynamic classes (string value).

mbest:
Please see #1953 for the code change to update `class`.

PR #1944, opened 2015-12-03
Component afterRender functionality

krishnaglick:
Components need a way to run functionality after they're done rendering.
Here's my solution


Domitnator:
When can we expect this feature to be commited to the master branch? Really looking forward to this one!

caseyWebb:
Isn't the `viewModel` executed after render?

rameshr1986:
Hi all is there a way to identify the page transition when all the subcomponents are loaded

rameshr1986:
I meant to show the page once all its components and sub components are loaded

Issue #1943, opened 2015-12-03
Infinite loop in circular computeds


juuxstar:
It looks like computeds are prone to a circular dirtying loop as demonstrated by:

```
ko.options.deferUpdates = true;
a = ko.observable();
b = ko.computed({ read : function() { a(); return d(); }, deferEvaluation : true });
d = ko.computed({ read : function() { a(); return b(); }, deferEvaluation : true });
d();
a('something');
```

This will crap out with a "Maximum call stack size exceeded".  Looks like it's not the evaluation of them but the scheduling for re-evaluation,

Tested with KO 3.4.

https://jsfiddle.net/ye0b2p0w/10/

juuxstar:
I believe the old stand-alone deferred library had a `clearDuplicate` method that dropped duplicate computeds from the refresh task queue.

vamp:
@juuxstar, good catch!
PS: as quick fix - disable deferUpdates

mbest:
So until this can be fixed, you can't use deferred updates with computed observables that depend on each other in a circular loop.

juuxstar:
Thanks for the quick response and suggestions guys.

Disabling deferred updates is a large performance hit in a non-trivial app though and it's not always possible to figure out how computed observable depend on each other.  Again, in a non-trivial app with many computed observables, large non-obvious loops can form and it's hard to figure out how to break the loop.

I'm going to create a fork and attempt a quick fix to the scheduler by checking first if the given callback has already been scheduled on the tasks queue to prevent duplicates (essentially the same solution from the original knockout deferred library).

mbest:
Sounds good, @juuxstar. Here's the code you'll likely want to play with: https://github.com/knockout/knockout/blob/master/src/subscribables/extenders.js#L52

juuxstar:
Hmm, so the above PR helps but it doesn't quite pass all the tests.  I'm not familiar enough with how the task queue is suppose to work to see what the problem is.  Any ideas why the tests are failing?

krnlde:
My 2 cents: require.js solves its circular dependencies by saving all the values in a map to decouple the values from the actual variables/implementation. Could this approach be applied here?

maskmaster:
Any large project is probably going to have circular dependencies between computeds.

This issue is currently stopping us from using deferred updates and is probably a reason other larger projects have problems with enabling it globally. #1758 tries to address migration problems, but this is something that really could enable migrations. This issue should imho have higher priority.

[...]

PR #1945, opened 2015-12-03
fix for #1943

juuxstar:
Checks whether a task is already queued and if so, ignores it.

### DIFF
--- a/src/tasks.js
+++ b/src/tasks.js
@@ -89,6 +89,13 @@ ko.tasks = (function () {
             }
         },
 
+        isScheduled: function(func) {
+            var keys = Object.keys(taskQueue);
+            for (var a = keys.length - 1; a >= 0; a--)
+                if (taskQueue[keys[a]] === func) return true;
+            return false;
+        },
+
         // For testing only: reset the queue and return the previous queue length
         'resetForTesting': function () {
             var length = taskQueueLength - nextIndexToProcess;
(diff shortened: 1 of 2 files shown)


src/tasks.js
```diff
@@ -89,6 +89,13 @@ ko.tasks = (function () {
             }
         },
 
+        isScheduled: function(func) {
+            var keys = Object.keys(taskQueue);
+            for (var a = keys.length - 1; a >= 0; a--)
+                if (taskQueue[keys[a]] === func) return true;
```
brianmhunt: This makes tasks at least O(n<sup>2</sup>) – might this be quite slow?

juuxstar: How n^2?  It's single pass over the queue so that would make scheduling a task go from O(1) to O(n).

It would be somewhat mitigated by the performance improvement of avoiding recomputing computed observables needlessly.  I don't know how often this happens in practice though (other than in the circular dependency edge case)

Another idea would be to use `Set` for the taskQueue which automatically eliminates duplicates.  It also conveniently returns entries in insertion order so easily facilities a queue.  But that's a larger re-factor.  Also Sets are not available on all browsers that Knockout is currently supporting (though to mitigate there are Set [polyfills](https://github.com/WebReflection/es6-collections) )

mbest:
It's probably simpler to check the `handle` than the `callback`.

juuxstar:
There that did it.  Turns out my previous solution did not account for subscribers getting attached after the first scheduled recalculation task.

PR #1946, opened 2015-12-07
Adding method to retrieve ordered list of dependencies for computed observables

WHenderson:
Hi,

I've added a method for retrieving the list of dependencies relating to a computed observable.
This is useful in contexts where you wish to interrogate the dependent observables or their extensions, such as during validation.

Some example pseudo-code

var ajaxData = ko.computed(function () { return vm.toJJSON(); });
var isInvalid = ajaxData.getDependencies().some(function (dependent) { return dependent.hasError(); });

See this post for the (rather short) discussion leading to this pull request: https://groups.google.com/forum/#!topic/knockoutjs/FAsQPtGR2J8

### DIFF
--- a/spec/asyncBehaviors.js
+++ b/spec/asyncBehaviors.js
@@ -557,6 +557,7 @@ describe('Rate-limited', function() {
             expect(computed()).toEqual(1);
             // The computed should not be dependent on the second observable
             expect(computed.getDependenciesCount()).toEqual(1);
+            expect(computed.getDependencies()).toEqual([observableSwitch]);
 
             // Updating the second observable shouldn't re-evaluate computed
             observableValue(2);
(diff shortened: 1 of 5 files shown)


brianmhunt:
Thanks @WHenderson 

In general I think this is a great idea.

It's a separate idea, but pairs well with the idea of named observables (i.e. passing a name param that we could use to identify observables).

mbest:
I like it too. I had previous included the same functionality in my [Deferred Updates plugin](https://github.com/mbest/knockout-deferred-updates/blob/master/knockout-deferred-updates.js#L564).

cervengoc:
If everyone likes it, why still hasn't it been merged yet? :)
I like also the named observables idea, could be a great help when debugging really large apps.

Issue #1949, opened 2015-12-10
knockout not updating values for Safari autocomplete


nzCharlie:
On Safari version 9.0.1 on OSX El Capitan, the observables binded to text type input value is not updating when the auto-complete is triggered. Saw [#1679](https://github.com/knockout/knockout/issues/1679), but the issue has been closed.  

We found that setting [valueUpdate](http://knockoutjs.com/documentation/value-binding.html) to *blur*, we are able to work around this issue.  We've got a patch to add 'blur' event to the list of events used to update for Safari, can submit a pull request if required.

checkphi:
Is there any chance this issue finds it's way into 3.5.0 or 3.4.2?

brianmhunt:
I'm not averse to the `blur` workaround in [tko.binding.core](/knockout/tko.binding.core).  To what bindings should the event be called?

checkphi:
Currently we have this problem when applying the value-binding to an input.

binarysource:
This one got me today, Safari Autofill and using valueUpdate: 'blur' seems to have fixed it.

Is this likely to make it into the next release?

PR #1950, opened 2015-12-10
Fixed knockout not updating values for Safari autocomplete, by using …

nzCharlie:
Fixed knockout not updating values for Safari autocomplete, by using blur event for updating.


nzCharlie:
See [#1949](https://github.com/knockout/knockout/issues/1949).

PR #1951, opened 2015-12-10
Support afterRender for if, ifnot and with bindings

cervengoc:
I saw several related issues but realized that it's still not
implemented.

So here is a basic support for the afterRender callback to if, ifnot and
with bindings. One can use it like this

```
with: data, afterRender: callback
if: condition, afterRender: callback
ifnot: condition, afterRender: callback
```

Semantics are same as with foreach and template bindings.

Added the some specs too, based on the specs found for foreach binding.

As far as other previously mentioned bindings go (visible, etc.) I
personally decided not to implement it for them because for example
visible is semantically different, there is nothing like "rendering" in
this sense.

### DIFF
--- a/src/binding/defaultBindings/ifIfnotWith.js
+++ b/src/binding/defaultBindings/ifIfnotWith.js
@@ -21,7 +21,12 @@ function makeWithIfBinding(bindingKey, isWith, isNot, makeContextCallback) {
                         if (!isFirstRender) {
                             ko.virtualElements.setDomNodeChildren(element, ko.utils.cloneNodes(savedNodes));
                         }
-                        ko.applyBindingsToDescendants(makeContextCallback ? makeContextCallback(bindingContext, rawValue) : bindingContext, element);
+                        var newContext = makeContextCallback ? makeContextCallback(bindingContext, rawValue) : bindingContext;
+                        ko.applyBindingsToDescendants(newContext, element);
+                        if (element.childNodes.length) {
+                          if (allBindings.has('afterRender'))
+                            ko.dependencyDetection.ignore(allBindings.get('afterRender'), null, [element.childNodes, newContext['$data']]);
+                        }
                     } else {
                         ko.virtualElements.emptyNode(element);
                     }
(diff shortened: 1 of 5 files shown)


src/binding/defaultBindings/ifIfnotWith.js
```diff
@@ -21,7 +21,13 @@ function makeWithIfBinding(bindingKey, isWith, isNot, makeContextCallback) {
                         if (!isFirstRender) {
                             ko.virtualElements.setDomNodeChildren(element, ko.utils.cloneNodes(savedNodes));
                         }
-                        ko.applyBindingsToDescendants(makeContextCallback ? makeContextCallback(bindingContext, rawValue) : bindingContext, element);
+                        var newContext = makeContextCallback ? makeContextCallback(bindingContext, rawValue) : bindingContext;
+                        ko.applyBindingsToDescendants(newContext, element);
+                        if (element.childNodes.length) {
```
brianmhunt: Could be written:

```
if (element.childNodes.length && allBindings.has('afterRender')) {
```

src/templating/templating.js
```diff
@@ -224,7 +224,7 @@
             if (typeof bindingValue == "string" || bindingValue['name']) {
                 // It's a named template - clear the element
                 ko.virtualElements.emptyNode(element);
-            } else if ('nodes' in bindingValue) {
+            } else if (bindingValue['nodes']) {
```
brianmhunt: Why change this?

cervengoc: Wanted to have this as a separate PR, Needed for a scenario where I wanted
to conditionally bind nodes.
2016. júl. 8. du. 4:36 ezt írta ("Brian M Hunt" <notifications@github.com>):

cervengoc: I will clean this up as soon as I have some time, but I guess I don't need to rush since there are tons of PRs sitting here.
Anyway I had a situation similar to this:

```
var model = {
  dataTemplate: ko.observable(), // can be an ID string, HTML string or Node array
  data: ko.observable()
}

...

function getTemplateNodes(template) {
  if (template && template.length) return template;
  if ("template is HTML string")
    return ko.utils.parseHtml(template);
  return null;
}

...

<div data-bind="template: { name: isTemplateName(dataTemplate()) ? dataTemplate() : null, nodes: getTemplateNodes(dataTemplate()) }"></div>
```

With the current implementation KO will always use the `nodes` binding whenever it's presented in the bound object, regardless of what it returns. I needed here a way to be able to have a name or have `nodes` conditionally based on the `dataTemplate` observable.

I checked several history entries to have an idea wether the current behavior is be intention or not, but haven't found any.

I will put together a separate PR for it for discussion. For now I extended the `template` binding to reach the proposed behavior.

Issue #1958, opened 2015-12-15
ko.isObservable(ko.computed) returns true although it shouldn't


mbest:
This causes `ko.toJSON(ko)` to throw an error "Pass a function that returns the value of the ko.computed".

PR #1970, opened 2016-01-07
Fix for issue #1958, ko.isObservable(ko.computed) returns true

MrFrankel:
Testing 'fn' object on tested object to verify its not a constructor object

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -9,7 +9,7 @@ module.exports = function(grunt) {
         fragments: './build/fragments/',
         banner: '/*!\n' +
                 ' * Knockout JavaScript library v<%= pkg.version %>\n' +
-                ' * (c) Steven Sanderson - <%= pkg.homepage %>\n' +
+                ' * (c) The Knockout.js team - <%= pkg.homepage %>\n' +
                 ' * License: <%= pkg.licenses[0].type %> (<%= pkg.licenses[0].url %>)\n' +
                 ' */\n\n',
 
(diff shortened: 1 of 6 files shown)


spec/observableBehaviors.js
```diff
@@ -10,6 +10,11 @@ describe('Observable', function() {
         expect(ko.isObservable(instance)).toEqual(true);
     });
 
+    it('Should advertise that instances are not observable', function () {
```
brianmhunt: Thanks for the PR;  Perhaps this would be better labelled `instances` should be `constructors`?

src/subscribables/observable.js
```diff
@@ -63,8 +63,10 @@ ko.hasPrototype = function(instance, prototype) {
 };
 
 ko.isObservable = function (instance) {
-    return ko.hasPrototype(instance, ko.observable);
-}
+    //test if not constructor function;
+    return  ko.hasPrototype(instance, ko.observable) && (instance['fn'] === undefined);
```
brianmhunt: For older browsers, it's better to use `void 0` instead of `undefined`, or to use the test `typeof instance['fn'] === 'undefined'`, just in case `undefined` is redefined (as can happen prior to ES5, rare though it would be).

MrFrankel:
Good points, fixed:) thanks for the review!

src/subscribables/observable.js
mbest: Knockout defines an `undefined` itself (and so doesn't use the global value). We can safely compare to `undefined`.

MrFrankel:
OK, that seems to be the consist method to test undefined throughout. Reverted back. thanks

mbest:
See #1958. Fixed with cdd2741d0f196005a9688cda3f0daa0462171852

Issue #1973, opened 2016-01-13
Knockout observable arrays leak memory when resubscribing to arrayChange events


ThomasMichon:
When debugging the causes of memory leaks in my application, I observed that objects were being kept by the `previousContents` variable from the `trackArrayChanges` extender for an observable array. What was odd was that it was not only the most-recent contents which were being captured here; in fact there were multiple closures captured, each with a distinct value for `previousContents`.

The closures in question were those of the `notifySubscribers` functions added by calls to `trackChanges()`. Each time a call to `trackChanges()` is made, it captures the current value of `target.notifySubscribers` and writes it to a local variable, `underlyingNotifySubscribersFunction`.

What I observed was that over time, `trackChanges()` was being called repeatedly on the same observable array, which was creating a new `notifySubscribers` method on each call. The old method turns out to be the previous `arrayChange` subscription, which has a `previousContents` variable with whatever the array had at the time. It never got garbage-collected.

The issue is that the `trackArrayChanges` extender does not properly restore the previous `notifySubscribers` function when the last subscription is removed. The function supplied for `afterSubscriptionRemove` sets `trackingChanges = false`, but it does not use the value for `underlyingNotifySubscribersFunction` to clean up. Thus, if you repeatedly subscribe and unsubscribe from `arrayChange` events on the same observable array, allowing some changes to happen in-between subscriptions, you will leak memory.

I have a JSFiddle showing the root cause in action:
https://jsfiddle.net/ThomasMichon/p0dgdm6y/

Although the `trackArrayChanges` extender is permanent, the changes made by `trackChanges()` are not expected to permanent (since `trackingChanges` may be set to `false`). Thus I should expect `notifySubscribers` to be the original function once I have disposed all subscriptions to the `arrayChange` event.

mbest:
Fixed with #1874.

PR #1974, opened 2016-01-13
Restore original notifySubscribers method when unsubscribing from arrayChange events

ThomasMichon:
This fixes issue #1973 by restoring the `underlyingNotifySubscribersFunction` in the `afterSubscriptionRemove` override imposed by the `trackArrayChanges` extender.

The impact of this change is that it eliminates memory leaks encountered when repeatedly subscribing and unsubscribing to the `arrayChange` events of observable arrays.


mbest:
This looks good. Thanks for the fix.

PR #1988, opened 2016-01-29
Enable composition in component binding

Manuel-S:
Enables the component binding to use a property on the parent viewmodel as the viewmodel of the component.

usage example:
```
<div data-bind="component: { name: 'myComponent', viewModel: childViewModel }" ></div>
```

```
ko.components.register('myComponent', { viewModel: MyComponent, template: '<b>Hello World</b>' } );
ko.applyBindings({childViewModel: new MyComponent()})
```


mbest:
You can already do this either with a custom loader or by defining a `createViewModel` function for the component.

Issue #1917, opened 2015-10-29
Performance issue - splice()


vamp:
Hello, masters! :massage: 

- ``` continuousNodeArray.splice(0, 1)``` should be ```continuousNodeArray.shift()```
- ```underlyingArray.splice(0, underlyingArray.length)``` should be ```underlyingArray.length = 0;```

P.S. ```splice(0,1)``` up to 99% slower then ```shift()```
based on: http://jsperf.com/remove-first-element-from-array/14

simonmurdock:
Ha, nice!

brianmhunt:
@vamp cool!  Just to be sure, are you looking at:

[continuousNodeArray.splice(0, 1)](https://github.com/knockout/knockout/blob/a03cd3f413850cc5401324452fec76763c404cac/src/utils.js#L271)

and 

[underlyingArray.splice(0, underlyingArray.length)](https://github.com/knockout/knockout/blob/8decc433942d7413b47768e0f45c304e8f15aa09/src/subscribables/observableArray.js#L40)

vamp:
@brianmhunt, yep, You're right

EricSmekens:
Voting for closing this issue.
See PR #1991.

brianmhunt:
Ah, it's jQuery compat ... #1513.

@vamp Are you okay if we close this issue (at least for now)?

mbest:
I think the second one is still fine to change.

puzrin:
> P.S. splice(0,1) up to 99% slower then shift()

Even more, JIT stores spliced arrays as "array with holes". Those have much slower read access. When you have rare splices & hot reads, it's better to create new array and copy elements there one-by-one.

jlaustill:
Have you tested this with different sized arrays?  In my testing, this is only true for small arrays, and the larger they get the less true this is.  My team tested this up to a million splices to come to this conclusion.  I'd be interested to see a fiddle that showed otherwise :)

puzrin:
@jlaustill we had issue with read speed from spliced arrays in [hike-js](https://github.com/nodeca/hike-js). Don't remember details, probably < 100 elements. Also lodash does rewrite in method, responsible for elements remove.

I don't know what happens with bigger sizes, but IMHO splice is canditade for benchmarking, if one does speed opts.

PR #1991, opened 2016-01-29
Changed splice(0,1) to shift.

EricSmekens:
See issue #1917 

Very minor performance improvemen, don't know if it's really worth it, but I don't see cons.


brianmhunt:
While we're at it there's also the `underlyingArray.length = 0;`  :)

mbest:
As far as I remember, the `splice` here is purposeful.

EricSmekens:
Ah ok, wasn't aware of that. Closing this PR!

PR #1995, opened 2016-02-04
Purpose readme contribution

stefanfrede:
To make Knockout more approachable to a larger number of users—from informed developers to curious passersby and new and interested visitors—I have updated the README.

The sections I added are entitled What is Knockout, What to Use Knockout for, and When to Use Knockout. As the name suggests, the new sections, three brief paragraphs, explain in plain and straightforward language what Knockout is, what it is used for and when one should use it.

Because this is part of a study assignment there is a good chance that similar requests will follow.


stefanfrede:
I need to squash my commits. Will request a commit afterwards again.

PR #2004, opened 2016-02-21
Evaluate the valueAccessor in a separate dependentObservable

simonwalz:
On a binding update the javascript expression defined in the data-bind attribute is re-evaluated each time. This is not necessary. Actually, this makes expressions returning a dynamically new created dependentObservable nearly impossible.

When evaluation the expression in a dependentObservable, the result is saved in this expression and only re-evaluated when necessary.

It would be kind, if you take my suggestion into account. Thank you.

### DIFF
--- a/src/binding/bindingAttributeSyntax.js
+++ b/src/binding/bindingAttributeSyntax.js
@@ -332,7 +332,15 @@
             var getValueAccessor = bindingsUpdater
                 ? function(bindingKey) {
                     return function() {
-                        return evaluateValueAccessor(bindingsUpdater()[bindingKey]);
+                        var bs = bindingsUpdater();
+                        if (bs[bindingKey].cache) {
+                            return bs[bindingKey].cache;
+                        }
+                        var value = evaluateValueAccessor(bs[bindingKey]);
+                        if (ko.isSubscribable(value)) {
+                            bs[bindingKey].cache = value;
+                        }
+                        return value;
                     };
                 } : function(bindingKey) {
                     return bindings[bindingKey];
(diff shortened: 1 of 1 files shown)


mbest:
Of course, for the case you've presented, this change could make sense. For most cases, though, it seems like overkill.

simonwalz:
Thank you, for your kind answer. You are right. I added another implementation in a third commit.

In this implementation I just cache a resulting observable object, when evaluating the valueAccessor. To store the object, I use a new property (named “cache”) in the valueAccessor function object. If the bindingsUpdater re-evaluates the bindings, a new valueAccessor function object is created and the cache property is deleted along with the old object.

Do you like this implementation, or do you have another idea how I could prevent complex expressions from being re-evaluated over and over again?

IanYates:
In my client-side code I might have a wrapped up an API that gets information on library books.  I could have a function like
```
function getBooksFor(author) {

}
```
 that returns an observable containing an array of books for the author.

As new books come in I'd like that observable to be updated automatically so the rest of the app can react to the changes.

I'd also like to be able to call that function multiple times.

So I write that function to effectively memoize its results by using a dictionary where author is the key and a pureComputed is stored as the value.  The nice thing is that I get back a consistent observable which avoids the issue you raised.  Because the value returned is a pureComputed I also avoid the expense of updating it if no one is subscribing to it currently.

I just tend to follow that kind of pattern if I have expensive work.  Would that work for you?

mbest:
This is a big change in how binding are handled that would require lots of performance and sanity testing. I don't really see that it would provide much benefit in most situations, but thanks again for the suggestions.

PR #2011, opened 2016-03-07
3.4.1

kostyamy:
I would like suggest to extend signature of the afterRender callback method to make a bit more informative while rendering foreach: {} constructs. Long story short on line 4988 i propose to add two additional arguments to the call so instead of current:
>>>   options["afterRender"] (addedNodesArray, arrayValue);
it would say something like following:
>>>   afterRender.call((arrayItemContext.$root || this), addedNodesArray, arrayValue, arrayItemContext, index);
in this case actual afterRender function is aware of the data-binding context


simonmurdock:
?

PR #2016, opened 2016-03-15
If a part of a Binding is "undefined" knockout should catch the excep…

jogibear9988:
…tion

I've fixed this in my fork like this! Maybe it will be possible to create a parameter where the template of the value function could be modified?


mbest:
Thanks for the suggestion. Perhaps we could add a `bindingOptions` setting for it.

You might also be able to do this through a preprocessor.

jogibear9988:
any news to this?

jogibear9988:
hello? should i upload a new request?

jogibear9988:
new patch : #2177

PR #2036, opened 2016-04-15
fall back to last page when items removed

hejiantao5:
if last few items are removed, we should switch to show the last page items instead of staying in no items page


rniemeyer:
Thanks @hejiantao5. Looks like maybe you want #2040 to replace this one? Just need to either update code here or submit new PR against `gh-pages`. Thanks.

hejiantao5:
yes, please help merge 2040

PR #2040, opened 2016-04-20
avoid negative index and fall back to last page

hejiantao5:
if there is no data, we should return maxPageIndex as 0;
if last item in last page removed, fall back to previous page


rniemeyer:
Thanks @hejiantao5 ! You will want to submit this against `gh-pages` to get it into the latest docs.

PR #2043, opened 2016-04-27
avoid zero max index and fall back to last page

hejiantao5:
if there is no data, we should return maxPageIndex as 0;
if last item in last page removed, fall back to previous page


hejiantao5:
@rniemeyer, re-commit the change accordingly #2040

Issue #2001, opened 2016-02-11
3.4.1 release


mbest:
So far:
- `pureComputed` sometimes doesn't pick up a change in a dependency (#1975 and #1992)
- Knockout observable arrays leak memory when resubscribing to `arrayChange` events (#1973 and #1974)
- `$rawData` within `with` bindings and rendered templates (#1415)
- Export `computed._latestValue` in debug mode (b3f24b1e247c1b)

https://github.com/knockout/knockout/tree/3.4.1

IanYates:
:100: from me
I hadn't brought over that $rawData one into my own copy but it looks like it'd be nice to have just in case.

dealproc:
is this possibly what I'm seeing when i have a computed observable that is bound to the checked binding, and i'm only getting writes on the uncheck (false) and not on the check (true)?

mbest:
@dealproc If you think it might be related, try building the 3.4.1-pre code and see if it helps.

codymullins:
Hi guys - this might not be the best place to ask, but something I've been trying to figure out. Is there some kind of long term road map for knockout or is it currently just being maintained? I use Angular at work, but have a side project (startup app) that uses knockout - it's one of the few that support back to IE8, which I currently require. Just wanted to know if there was a plan in terms of features or product evolution that I could expect or potentially contribute to making happen?

IanYates:
@codymullins - not sure.  In terms of what KO does, it's fairly complete.  Components, etc are well-defined.  There are some bits and pieces being discussed around refining component loading some people can do transitions and "know" when every component is ready (a tricky term to define since a new component could be loaded at any time really)

The 3.x series of releases has been pretty huge in terms of new functionality.  The deferred update capabilities that @mbest introduced look awesome but I haven't had a chance to introduce them into our codebase yet as it would require a good deal of testing of some UI/Promise/Computed/etc interactions.

A lot of new functionality comes via way of the many third party bindings.  Having @mbest's punches library integrated into KO would be neat but adding it on as a second script isn't really a big deal.  I do wish it was more discoverable to newcomers though - maybe having it exposed on the home page as a featured plugin.  Same goes for other bindings I use like switch/case (love that one!)

What sort of things are you coming across that you wish were in core KO?

JD-Robbs:
Just wondered if it'd make sense to have a release with just the pureComputed fix for now, seeing it's already been fixed. 

@IanYates having knockout.mapping supported in the core would be nice as well.

[...]

PR #2064, opened 2016-06-03
Add template literal parsing to expressionRewriting

cmbankester:
Currently, Knockout does not parse ES2015 [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) correctly. E.g., this does not work:

``` html
<div data-bind="a: `Hello, world!`"></div>
```

The reason it fails to parse should be obvious: the backtick is considered as part of an un-quoted string instead of string literal syntax, and so the comma in the middle gets picked up by the expression rewriter as an indication of a new binding attribute. Thus, the expression rewriter would parse the above as:

``` javascript
[
  { key: 'a', value: '`Hello' },
  { unknown: 'world!`' }
]
```

While it _should_ parse it as:

``` javascript
[
  { key: 'c', value: '`Hello, world!`' }
]
```

This PR augments the `expressionRewritingBehaviors#Should be able to parse object literals containing string literals` spec with a check for template literals, and augments the expressionRewriter with the ability to parse template literals correctly.

Let me know if I need to do anything else to get this merged in!

### DIFF
--- a/spec/expressionRewritingBehaviors.js
+++ b/spec/expressionRewritingBehaviors.js
@@ -22,12 +22,24 @@ describe('Expression Rewriting', function() {
     });
 
     it('Should be able to parse object literals containing string literals', function() {
-        var result = ko.expressionRewriting.parseObjectLiteral("a: \"comma, colon: brace{ bracket[ apos' escapedQuot\\\" end\", b: 'escapedApos\\\' brace} bracket] quot\"'");
-        expect(result.length).toEqual(2);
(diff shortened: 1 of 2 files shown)


brianmhunt:
This is simple but important; I'm ok with this for #2001 / Release 3.4.1.

cmbankester:
Just checking to see if anything needs to be done with this to get it merged in?

mbest:
I'm adding this for 3.5.0

brianmhunt:
Related:  tko supports template literals out of the box.

PR #2065, opened 2016-06-06
Update isComputed evaluation

robinchow:
When knockout.mapping is used, it replaces ko.computed with its own definition of ko.computed, and ends up causing isComputed to return incorrect results (ie true for computed observables). This patch introduces a new ko.originalComputed variable that should not be overwritten/reassigned, and is used in the evaluation of isComputed.


robinchow:
I understand this is technically knockout.mapping's "fault", but I've looked into that code and I don't have a good enough understanding to fix it on that side. And given that knockout itself is more actively maintained than knockout.mapping, and that this fix (adding an extra variable ko.originalComputed that tracks the original computed prototype) doesn't really affect anything else, I think it would be useful to have it added to the next release of knockout.

PR #2068, opened 2016-06-07
Add typescript as the first source transpiler

SergioMorchon:
Add typescript as the first source transpiler.
It will keep the .js files, but also will transpile .ts ones. This allows the hybrid scenario of having typescript and javascript files, both at the same time, and write newer code in typescript.


mbest:
see #2069

PR #2069, opened 2016-06-07
Update build process

SergioMorchon:


mbest:
Why?

SergioMorchon:
Sorry, I intended to make this pull request to my own fork.

By the way, my goal is to include typescript as a task before the current steps.
With that, you can take advantage of static typing and modern es6 syntax.

And in a future we could code using es6 modules with typings, generate vanilla es6 javascript and later use some modern transpilers (babel, corejs, ...) and bundlers (webpack, rollup, ...).

What do you think @mbest?

SergioMorchon:
With this changes we can also have both current js and new ts files.

I have a branch where I will migrate the files.

First, using namespaces (and without ko.exportXxxx).
Later, modules.

PR #2091, opened 2016-07-02
Computed observable memory leak fix

coot:
The dispose method of `ko.computed` should release a link to the owner (vel `evaluatorFunctionTarget`).


mbest:
The test updates show that it's currently possible to "write" to a disposed computed observable, and the changes here could cause that to fail.

chrisknoll:
I understand we wouldn't want to break existing functionality, but shouldn't the case where you access a disposed observable be considered invalid?  Shouldn't disposed mean it's gone?

mbest:
Reading a disposed computed is definitely supported and used internally in Knockout. Writing to a disposed computed, though, doesn't seem to have much point. So I'm okay with saying it's not supported. Perhaps, then it should throw an error rather than failing in some unknown way.

coot:
Closing because of inactivity.

PR #2071, opened 2016-06-10
`as:` binding can be used with `with:`

mblarsen:
Add the option to use `as:` with `with:`. This is useful for deeply nested hierarchies where code like `$parents[3]` or `$parents[$parents.length-2]` becomes hard to read.

PR #2093, opened 2016-07-11
Added predicate object match for arrayFirst like lodash find()

mblarsen:
E.g. 

```
ko.utils.arrayFirst(this.categories(), { id: 4 });
```

This is useful for determining of new data should be added to an array.

Specs, updated.

### DIFF
--- a/spec/utilsBehaviors.js
+++ b/spec/utilsBehaviors.js
@@ -148,6 +148,48 @@ describe('arrayFirst', function () {
         expect(matchD.calls[1].args).toEqual(["b", 1]);
         expect(matchD.calls[2].args).toEqual(["c", 2]);
     });
+
+    it('Should match no non-objects with object match', function () {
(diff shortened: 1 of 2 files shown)


mblarsen:
@rniemeyer I think I managed to branch it correctly #2071

Thanks for the help 👍

mbest:
In general, the `ko.utils` functions are meant to support Knockout itself. Thus I don't think this addition is important to include in `ko.utils`.

mblarsen:
I kind of got that feeling reading the code. 👍

miellaby:
@mblarsen you could transform your code in a tool returning a arrayFirst compatible closure. usage: ` ko.utils.arrayFirst(this.categories(), matcher({ id: 4 })); `

mblarsen:
@miellaby good point. This seem to work https://gist.github.com/mblarsen/b9cc9ae0086e8eb07c78f5745afc783f

PR #2100, opened 2016-07-25
Support afterRender for if, ifnot and with bindings, fixes #2099

### LINKED ISSUE #2099: afterRender is not available for with, if and ifnot bindings
I made this issue to clean up up my earlier pull request and have an issue to reference.

The `with`, `if` and `ifnot` bindings currently do not support the `afterRender` callback.

cervengoc:
Adds support for `afterRender` callback to `if`, `ifnot` and `with`
binding.

This pull request is a cleaned and updated version of my earlier one,.

### DIFF
--- a/spec/defaultBindings/ifBehaviors.js
+++ b/spec/defaultBindings/ifBehaviors.js
@@ -94,4 +94,83 @@ describe('Binding: If', function() {
         condition2(true);
         expect(testNode).toContainHtml("hello <!-- ko if: condition1 -->first is true<!-- ko if: condition2 -->both are true<!-- /ko --><!-- /ko -->");
     });
+
+    it('Should call an afterRender callback function and not cause updates if an observable accessed in the callback is changed', function () {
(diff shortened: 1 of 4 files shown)


mbest:
This looks nice. Can you update the indent spacing in ifIfnotWith.js to match the rest of the code (4 spaces)? Thanks,

mbest:
Looks like this is missing some changes from #1951.

IanYates:
I could see this being useful in a few cases.  Thanks for making the PR

cervengoc:
@mbest it only misses changes which were incorrectly there in the previous PR, so it should be fine. Actually I'm still a beginner in using git and only have time for "in-time learning". I will fix the indentation.

mbest:
I've merged this with changes to make it work with virtual elements (000339731fa22395c01948800f7b8f83d325ea1e).

mbest:
I just thought that this could be made generic to work with any binding (or not) that binds its descendants. In other words, I could use `<div data-bind="afterRender: myCallback">...</div>` to call `myCallback` after the `<div>`'s descendant nodes were bound.

PR #2103, opened 2016-07-27
Implementing the hidden binding

cervengoc:
- Added the `hidden` binding similarly to `disable`.
- Added the corresponding tests based on tests for `visible` binding
- Renamed corresponding files to reflect the changes


cervengoc:
See also #968, this should close that one as well.

mbest:
Github shows way more changes than it should. Can you update this pull request without the whitespace changes? You can replace the commit and then do a forced push to Github.

cervengoc:
I'll try to do my best, sorry. It seems like Visual Studio messes up the formatting automatically in some cases.

cervengoc:
To me it doesn't look like it's related to whitespaces in this case. Probably the issue was that I first made the code changes, and then ran the `git mv` command. I'll try to replace the commit and see the results.

spec/runner.html
```diff
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+﻿<!DOCTYPE html>
```
cervengoc: This is because the added UTF-8 header, I don't know wether this is problem or not. It was placed when Visual Studio saved the file, and I have no idea how I could avoid it.

cervengoc:
@mbest I couldn't improve the diff on renamed files, but the `source-references.js` and `runner.html` have improved. Please check it if it's alright now.

mbest:
Okay. Thanks for trying.

mbest:
Merged with 79d3f5f6b

PR #2105, opened 2016-07-27
Check actual value of nodes proeprty within template binding

### LINKED ISSUE #2104: Using nodes within template binding does not take returned value into account
Consider a situation where the viewmodel itself has control over the template contents.:

```
var model = {
  dataTemplate: ko.observable(), // can be an ID string, HTML string or Node array, or empty to use the child nodes as contents
  data: ko.observable()
}

function isTemplateName(template) {
  return typeof template === "string" && template.length && template.charAt(0) == "#";
}

function getTemplateNodes(template) {
  if (template && template.length) return template;
  if ("template is HTML string")
    return ko.utils.parseHtml(template);
  return null;
}

<div data-bind="template: { name: isTemplateName(dataTemplate()) ? dataTemplate() : null, nodes: getTemplateNodes(dataTemplate()) }"></div>
```

With the current implementation, if `dataTemplate` is empty, it will still use the `nodes` value because the value is actually not checked, only the existence of the `nodes` proeprty is checked. The correct behavior would be to fall back to child nodes, like there were no `nodes`, nor `name`.

cervengoc:
Fixes #2104

### DIFF
--- a/src/templating/templating.js
+++ b/src/templating/templating.js
@@ -224,7 +224,7 @@
             if (typeof bindingValue == "string" || bindingValue['name']) {
                 // It's a named template - clear the element
                 ko.virtualElements.emptyNode(element);
-            } else if ('nodes' in bindingValue) {
+            } else if (bindingValue['nodes']) {
                 // We've been given an array of DOM nodes. Save them as the template source.
                 // There is no known use case for the node array being an observable array (if the output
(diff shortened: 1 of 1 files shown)


cervengoc:
Actually I've just checked the failed test. Don't know why it didn't fail locally at me.

So it seems like this logic was un purpose. Could anyone explain why?

PR #2123, opened 2016-09-16
Added bindingInfo to binding exception

cbastos:
Added DOM element node and binding context info to 'unable to process binding' exception.
With this change we can retrieve more info in exception management:

For example: viewmodel constructor and failing dom element node.
ko.onError = function(e){
	console.log({ viewModel : e.bindingInfo.context["$component"].constructor , node :   e.bindingInfo.node });
}

### DIFF
--- a/src/binding/bindingAttributeSyntax.js
+++ b/src/binding/bindingAttributeSyntax.js
@@ -391,6 +391,7 @@
                         );
                     }
                 } catch (ex) {
+                    ex.bindingInfo = { context: bindingContext, node: node };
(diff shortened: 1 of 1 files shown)


brianmhunt:
This was one of the first issues addressed and will definitely be covered in tko. ( when it builds ... Soon! )

this is a pretty innocuous add to ko 3.4 but probably not API compatible going forward.

Want to leave it open or wait for tko?

cbastos:
I would like to be included in both versions :)

PR #2129, opened 2016-09-22
Documents parseHtmlFragment and setHtml

ManrajGrover:
This PR documents `ko.utils.parseHtmlFragment` and `ko.utils.setHtml`.

### DIFF
--- a/src/utils.domManipulation.js
+++ b/src/utils.domManipulation.js
@@ -94,12 +94,20 @@
         }
     }
 
+    /**
+     * parseHtmlFragment uses its jQuery's parseHTML function,
+     * otherwise it uses simple HTML parse
+     */
     ko.utils.parseHtmlFragment = function(html, documentContext) {
         return jQueryInstance ?
             jQueryHtmlParse(html, documentContext) :   // As below, benefit from jQuery's optimisations where possible
             simpleHtmlParse(html, documentContext);  // ... otherwise, this simple logic will do in most common cases.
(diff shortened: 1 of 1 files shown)


ManrajGrover:
@brianmhunt Looks like Travis Build didn't pass. I feel this is because of the warning. Some insights and help would be required.

brianmhunt:
@ManrajGrover tko.utils (where this was merged) passes: https://circleci.com/gh/knockout/tko-policy

There are a long list of build & testing issues with knockout. See e.g. #1360 .

mbest:
The build failed with "The following files have trailing spaces that need to be cleaned up". Why does this fail the build? It's to prevent files with trailing spaces from being checked in so that later on, editors that automatically remove trailing spaces don't cause superfluous changes.

ManrajGrover:
@brianmhunt I'll have a look at the PR. So, contributions should now be make to `tko.utils`?

@mbest I'll take care of that in future contributions. 😅

chrisknoll:
This probably would have failed on the TKO side if the TKO required those trailing spaces to be fixed (if I'm reading that correctly).  I wouldn't base where to place pull requests based on those guidelines.  

@brianmhunt: if it's _technically_ knockout, shouldn't that apply to these sorts of test considerations as well?

brianmhunt:
Yes, we will have an eslint test on checking for the tko family of packages too at some point. :)

PR #2170, opened 2016-12-09
Improve performance of named templates by caching the parsed nodes

mbest:
… so they can be used for future rendering by using `cloneNode`. This will only help if a template is used more than once, especially with `foreach`.

See #1201 for the motivation.

### DIFF
--- a/spec/nativeTemplateEngineBehaviors.js
+++ b/spec/nativeTemplateEngineBehaviors.js
@@ -50,7 +50,7 @@ describe('Native template engine', function() {
 
         it('can fetch template from <textarea> elements and data-bind on results', function () {
             var testTextAreaTemplate = ensureNodeExistsAndIsEmpty("testTextAreaTemplate", "textarea"),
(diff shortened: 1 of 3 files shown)


spec/nativeTemplateEngineBehaviors.js
```diff
@@ -50,7 +50,7 @@ describe('Native template engine', function() {
 
         it('can fetch template from <textarea> elements and data-bind on results', function () {
             var testTextAreaTemplate = ensureNodeExistsAndIsEmpty("testTextAreaTemplate", "textarea"),
-                prop = (typeof testTextAreaTemplate.innerText !== "undefined") ? "innerText" : "textContent";
+                prop = "value";
             testRenderTemplate(testTextAreaTemplate, "testTextAreaTemplate", prop);
```
mbest: This is a possible concern. Because of the changes here, you won't be able to change a `<textarea>`-based template by changing its `textContent` after its been used. That's because the template code uses `value` to read and set the content, and once `value` is set, the browser ignores changes to `textContent`.

PR #1835, opened 2016-12-09
ko.computed gets stuck (rate-limited dependency)

mbest:
Something seems to be wrong with the current implementation of rateLimit extender. Having a hard time nailing down what's going on in the code, but it looks like a rateLimited observable at some point fails to notify its dependents. IOTW, it seems to "settle" on a value w/o notifying its observers of the change.

I've distilled the problem down to this fuzzy example https://gist.github.com/hnovikov/ac311d6b923fc18774f9 (viewable at https://rawgit.com/hnovikov/ac311d6b923fc18774f9/raw/5b44eb99d4405b6b5934d210cc6332a94d89810e/koextender.html). It simply flaps two observables (one's rate limited) until a computed based on both of them jams on a previous value (!!)

Did a git bisect which identified https://github.com/knockout/knockout/commit/e7a95aae8915a38ce0f72e3ffcbdf8e4f65eb3a3 as the commit in which the issue was introduced.

### DIFF
--- a/spec/asyncBehaviors.js
+++ b/spec/asyncBehaviors.js
@@ -426,6 +426,36 @@ describe('Rate-limited', function() {
             expect(evalSpy).not.toHaveBeenCalled();
         });
 
+
+        it('Should not cause loss of updates when an intermediate value is read by a dependent computed observable', function() {
+            // From https://github.com/knockout/knockout/issues/1835
+            var one = ko.observable(false).extend({rateLimit: 100}),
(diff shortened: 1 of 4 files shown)


mbest:
I've attached a fix here so that if a computed observable accesses an observable with a pending notification, it will get notified of the actual final value (bypassing equality checks).

mbest:
Hmm. The changes here cause the new spec in #1943 to fail. I'll have to take another look at this.

PR #2171, opened 2016-12-10
observables notify a "spectate" event whenever their value changes.

mbest:
This is like the "change" event but is not delayed by rateLimit/deferred and will notify when a sleeping pure computed observable gets a new value (by being accessed). Subscribing to "spectate" will not awaken a sleeping pure computed either.

See #1735 and #1511

### DIFF
--- a/spec/asyncBehaviors.js
+++ b/spec/asyncBehaviors.js
@@ -245,6 +245,31 @@ describe('Rate-limited', function() {
             expect(beforeChangeSpy.calls.length).toBe(1);   // Only one beforeChange notification
         });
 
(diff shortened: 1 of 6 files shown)


mbest:
`spectate` occurs before `change`. For a rate-limited or deferred computed, the `spectate` event doesn't cause the computed value to be calculated early. Instead it will be fired right before the delayed `change` event, or if the computed is accessed earlier, it will be fired then.

brianmhunt:
@mbest - Glad you added 4e55d4b – makes the changes to `updateVersion` easier to understand.

PR #2148, opened 2016-12-17
dataFor / contextFor not honoring controlsDescendantBindings

mbest:
so in my initial research, the common check to see if an element was bound that people do is something along the lines of ```!!ko.dataFor(someElement)```... this at the surface seems like exactly what i need for my usecase, however there seems to be one issue: this does not take in to account ```controlsDescendantBindings```. I have areas of my application that i need to separately bind and ideally i'd like to be able to ask ko if that has happened... i utilize ```controlsDescendantBindings``` to power this separation of concerns and am looking for a solution to check if ```applyBindings``` has occured and ```dataFor``` seems to traverse up the DOM and look for some parent binding. i'm expecting ```null``` and getting data for an unbound div.

thoughts? this seems like a deficiency in that ```dataFor``` / ```contextFor``` logic.

Thanks.

### DIFF
--- a/spec/bindingAttributeBehaviors.js
+++ b/spec/bindingAttributeBehaviors.js
@@ -2,13 +2,13 @@ describe('Binding attribute syntax', function() {
     beforeEach(jasmine.prepareTestNode);
 
     it('applyBindings should accept no parameters and then act on document.body with undefined model', function() {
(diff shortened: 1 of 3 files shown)


mbest:
This topic has been around a while. See #905. I've attached the fix.

PR #2174, opened 2016-12-29
deferedUpdate causes depended observable chains to reevaluate every time

mbest:
Lets consider the following example:
```
let source = ko.observable({ key: 'value' });

let c1 = ko.computed(
    () => {
        let val = source()['key'];
        console.debug('EVALUATING c1:', val);
        return val;
    }
);

let c2 = ko.computed(
    () => {
        let val  = c1();
        console.debug('EVALUATING c2:', val);
        return val;
    }
);

ko.computed(
    () => c2() === 'value' && console.debug('SIDE EFFECT')
);

source({ key: 'value' });
```

We would expect the following debug log:
```
EVALUATING c1: value
EVALUATING c2: value
SIDE EFFECT
EVALUATING c1: value
```

But when using:
```ko.options.deferUpdates = true;```

The debug log shows:
```
EVALUATING c1: value
EVALUATING c2: value
SIDE EFFECT
EVALUATING c1: value
EVALUATING c2: value
SIDE EFFECT
```
where the side effect is evaluated twice.

In our case the side effect is an async knockout validator that sends an HTTP request and the source is reloaded periodically using another HTTP request.

A further check shows that no "change" events fire for c1 or c2 when changing the source.

Searching the source code for v3.4.0 reveals that a depended observable (computed) decides 
to reevaluate its value whenever a "dirty" event (not a "change" event) is broadcast from one or more of its dependencies.

I narrowed down the problem to a section of code from  src/subscribables/extenders.js:
```
47:    if (!target._deferUpdates) {
48:        target._deferUpdates = true;
49:        target.limit(function (callback) {
50:        var handle;
51:        return function () {
52:            ko.tasks.cancel(handle);
53:            handle = ko.tasks.schedule(callback);
54:            target['notifySubscribers'](undefined, 'dirty');
55:        };
56:    });
```

Here we can see that the when using deferUpdates the code notify all subscribers with a drity event which cause depended observables (computeds) reevaluation.

### DIFF
--- a/spec/asyncBehaviors.js
+++ b/spec/asyncBehaviors.js
@@ -975,6 +975,26 @@ describe('Deferred', function() {
             expect(notifySpy.argsForCall).toEqual([['i(x,h(cx,g(ex,fx),d(bx,cx)),bx,fx)']]);    // only one evaluation and notification
         });
 
+        it('Should minimize evaluation when dependent computed doesn\'t actually change', function() {
+            // From https://github.com/knockout/knockout/issues/2174
+            this.restoreAfter(ko.options, 'deferUpdates');
+            ko.options.deferUpdates = true;
+
+            var source = ko.observable({ key: 'value' }),
+                c1 = ko.computed(function () {
+                    return source()['key'];
+                }),
+                countEval = 0,
+                c2 = ko.computed(function () {
+                    countEval++;
+                    return c1();
+                });
+
+            source({ key: 'value' });
+            jasmine.Clock.tick(1);
+            expect(countEval).toEqual(1);
+        });
+
         it('Should ignore recursive dirty events', function() {
             // From https://github.com/knockout/knockout/issues/1943
             this.restoreAfter(ko.options, 'deferUpdates');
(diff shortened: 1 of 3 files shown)


mbest:
I've attached the code changes for this.

nb-ohad:
Is this fix compatible with v3.4.1 codebase ?

mbest:
I can check later, but it should be.

nb-ohad:
Do you have any intentions to release this fix as/in a v3.4.2 release or should i fork v3.4.1 myself?

Just to give a bit of context. Our product installs itself on customer sites using "npm install" so in order for us to be able to use this fix we have to have a git repository serving a version of knockout that includes this fix. Currently the product was tested with v3.4.1 and updating to v3.5 (which is not even released yet) is a risk we cannot take at the moment (We are going to have a major release in a few days). 

If the only merge path for this fix is the master brunch (v3.5) please notify my quickly so i could fork v3.4.1 myself.

nb-ohad:
Was closed by mistake

PR #2163, opened 2016-12-31
Dynamically displayed component being rendered more than once

mbest:
I've been battling this one for a few weeks, and just now got confirmation that someone else was able to reproduce this and I wasn't just going crazy.

The issue arises when I am dynamically displaying a component based on some logic - in this case, I'm parsing the route and grabbing a parameter. When the component is looking at an observable to determine the name, it's being loaded twice in the `getComponent` call and thus my constructor is called twice. If I bind the same form with a literal string, the component only renders once.

Here's an example (it has the same issue even without the outer `if` binding):
```
<!-- WORKS, renders only once -->
<div data-bind='if: currentFormComponent() != null'>
    <div data-bind='component: {name: "my-component", params: { route: pageParams.route }}'></div>
</div>

<!-- Renders twice (currentFormComponent is an observable and is set to "my-component" as above) -->
<div data-bind='if: currentFormComponent() != null'>
    <div data-bind='component: {name: currentFormComponent, params: { route: pageParams.route }}'></div>
</div>
```

### DIFF
--- a/spec/asyncBehaviors.js
+++ b/spec/asyncBehaviors.js
@@ -347,6 +347,28 @@ describe('Rate-limited', function() {
             expect(notifySpy).not.toHaveBeenCalled();
         });
 
+        it('Should not notify future subscribers', function() {
+            var observable = ko.observable('a').extend({rateLimit:500}),
+                notifySpy1 = jasmine.createSpy('notifySpy1'),
+                notifySpy2 = jasmine.createSpy('notifySpy2'),
+                notifySpy3 = jasmine.createSpy('notifySpy3');
+
+            observable.subscribe(notifySpy1);
+            observable('b');
+            observable.subscribe(notifySpy2);
+            observable('c');
+            observable.subscribe(notifySpy3);
+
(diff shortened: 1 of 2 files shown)


mbest:
I've attached the fix here.

codymullins:
I can confirm this fixes the issue. I don't see any other evident side effects either.

codymullins:
@mbest since this is merged in, are you planning the 3.4.2 release anytime soon?

mbest:
I only just found this relates SO post: http://stackoverflow.com/questions/36281676/knockout-3-4-0-deferupdates-and-reevaluation-count

mbest:
> since this is merged in, are you planning the 3.4.2 release anytime soon?

Hopefully. I just added another issue to consider for 3.4.2, plus there's a [possible regression](http://stackoverflow.com/questions/42134449/knockout-js-3-4-1-throws-multiple-bindings-error) in 3.4.1 that I want to look into.

mbest:
@codymullins 3.4.2 is now released.

PR #2177, opened 2017-01-04
Exception friendly Binding

jogibear9988:


jogibear9988:
Any News if this could be merged?

brianmhunt:
This duplicates functionality that's in tko (KO4) in a way that is not forward compatible, so it's unlikely this will be merged.

There are also a number of other PR requests and issues about this re. KO 3.x.

See: https://github.com/knockout/tko.bind/blob/master/src/applyBindings.js#L332

PR #2183, opened 2017-01-11
Remove hidden dev dependency on Java

Iiridayn:
Since Java wasn't listed as something I should need to install to build knockout, I've tweaked the build file and dev dependencies to remove the hidden Java dependency. It might be nice to improve the errors/warnings to know which file they originally came from, but right now it's just feature equivalent.


Iiridayn:
Looks like you'd need to update your minimum developer version of node to support this. Possibly not a deal breaker.

Issue #2182, opened 2017-01-11
ES6-style params shorthand (feature request)


meltuhamy:
It would be cool if we could use shorthand property names/values in bindings.

e.g. 

```<my-component params="thisIsSomethingOnTheVM, somethingElse, asBefore: something"></my-component>```

would be equivalent to 

```<my-component params="thisIsSomethingOnTheVM: thisIsSomethingOnTheVM, somethingElse: somethingElse, asBefore: something"></my-component>```

Thoughts?

brianmhunt:
I like it

codymullins:
This is something I've wanted to explore too. A similar item is on the data-bind, seems very verbose.

meltuhamy:
@brianmhunt I've created a pull request with the change. I have added and modified tests to allow this syntax. I'm not sure if maybe a switch is required to allow backwards compatibility? If so, where is the best place to put it? Does it maybe make more sense to treat this kind of modification as an extension?

PR #2185, opened 2017-01-17
#2182 allow es2015 style params and bindings

meltuhamy:


meltuhamy:
hmm, cant seem to reproduce this locally.. I tried running `grunt`, `npm run test` and `npm install` locally and they all pass... 

```
λ npm run test

> knockout@3.5.0-pre test D:\dev\knockout
> node spec/runner.node.js

Running Knockout tests in Node.js
Total:1015 Passed:1015 Failed:0
```

Am I missing something?

mbest:
This would be a breaking change in `data-bind` bindings. However, this might be okay for components. It would require passing an option when parsing a component's bindings.

PR #2188, opened 2017-01-23
Fixing multiple anonymous template bindings with same nodes, fixes #2187

### LINKED ISSUE #2187: Anonymous templates with same nodes cause problems
If I have template bindings which refer to the same `nodes` set, then when reevaluating at first time, all of the nodes are lost, except for the "last" one.

I've been digging into the source for an hour now, and I've found that this is due to a behavior that ko stores the `nodes` in a detached `div`, which is created on a per-binding basis. So when ever it meets a template binding which refers to the same node set, it will replace them into a new detached `div`. 

So when the first template binding would like to reevaluate itself, it will meet an empty `div` because it has lost it's nodes. The last `template` binding with the same `nodes` steals the nodes from every previous binding.

A relatively simple fix could be if ko could behave like *"hey, these nodes are attached to my technical `div`, so I shouldn't remove them, but clone them".*. Or if it's safe, it would be even better to do like *hey, these nodes are attached to my technical `div`, so I should just use that `div` and not create another*.

cervengoc:
Fixes #2187

### DIFF
--- a/spec/templatingBehaviors.js
+++ b/spec/templatingBehaviors.js
@@ -518,6 +518,36 @@ describe('Templating', function() {
         expect(testNode.childNodes[0]).toContainHtml("begin<span>the name is beta</span>end");
     });
 
+    it('Should accept a "nodes" option that gives the template nodes, and able to use the same nodes for multiple bindings', function () {
+      testNode.innerHTML = "<div data-bind='template: { nodes: testNodes, data: testData1, bypassDomNodeWrap: true }'></div><div data-bind='template: { nodes: testNodes, data: testData2, bypassDomNodeWrap: true }'></div>";
+      var model = {
+        testNodes: [
+            document.createTextNode("begin"),
+            document.createElement("span"),
+            document.createTextNode("end")
+        ],
+        testData1: ko.observable({ name: ko.observable("alpha1") }),
+        testData2: ko.observable({ name: ko.observable("alpha2") })
+      };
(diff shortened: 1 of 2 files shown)


spec/templatingBehaviors.js
```diff
@@ -518,6 +518,36 @@ describe('Templating', function() {
         expect(testNode.childNodes[0]).toContainHtml("begin<span>the name is beta</span>end");
     });
 
+    it('Should accept a "nodes" option that gives the template nodes, and able to use the same nodes for multiple bindings', function () {
+      testNode.innerHTML = "<div data-bind='template: { nodes: testNodes, data: testData1, bypassDomNodeWrap: true }'></div><div data-bind='template: { nodes: testNodes, data: testData2, bypassDomNodeWrap: true }'></div>";
```
mbest: What is bypassDomNodeWrap?

mbest: Oh I see it's used elsewhere. It might be leftover from something else.

cervengoc: Yes I've actually copied part of the code. I've checked that setting too and haven't found any other usages so I guess it was used in some previous versions and it's left there accidentally. I didn't remove it to keep the focus of this PR.

cervengoc:
@mbest could you please give some feedback on this?

mbest:
This looks good. Thanks.

mbest:
I decided to make the logic change in templating instead utils: https://github.com/knockout/knockout/commit/10ec231

PR #2191, opened 2017-01-26
Fixes #2190 Implement a util.wrap function

### LINKED ISSUE #2190: Implement a util.wrapObservable
Me and my colleagues often find ourselves having to test if a value is an observable and if not wrap it in one. E.g. when building components that should support both primitive/obj and Observable parameters. This functionality would basically be the reversed ko.util.unwrapObservable.

bjornalm:
First stab, without wrapping functions in computables.

### DIFF
--- a/spec/utilsBehaviors.js
+++ b/spec/utilsBehaviors.js
@@ -29,6 +29,55 @@ describe('unwrapObservable', function() {
     });
 });
 
+describe('wrap', function() {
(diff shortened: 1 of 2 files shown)


brianmhunt:
Feel free to submit a similar PR to [tko.computed](https://github.com/knockout/tko.computed), and it'll make it into the next tko alpha.

I think `ko.wrap` should stick to primitives, since otherwise it'd be wrapping `ko.computed`.

PR #2199, opened 2017-02-07
Change 'textInput:' binding to 'value:' in Ex. 3

james1293:
IMO, in this situation, the immediate reaction caused by the textInput binding is irritating from the user's perspective. If you try typing 42 in the live example < http://knockoutjs.com/documentation/computed-writable.html > you will end up with $4.002, because it immediately formats the 4 as "$4.00". If you use value binding instead, it lets the user finish typing. Example: https://jsbin.com/megiredusu/edit?html,js,output

Note that I also removed "immediately" from the explanation of the example.

### DIFF
--- a/documentation/computed-writable.md
+++ b/documentation/computed-writable.md
@@ -116,12 +116,12 @@ Sometimes you might want to represent a data point on the screen in a different 
     ko.applyBindings(new MyViewModel());
 {% endcapture %}
 {% capture live_example_view %}
-    <div>Enter bid price: <input data-bind="textInput: formattedPrice"/></div>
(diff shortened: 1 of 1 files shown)


chrisknoll:
I also have to admit that I've been burned in cases where I want to input a numeric formatted number, but it immediately reverts the invalid value of - when all i want to write is a negative number, or . when I want to write a decimal.

Doesn't work too good with validated input.

mbest:
Another option might be to use `rateLimit` with `notifyWhenChangesStop` on the computed.

james1293:
Thanks! I hope it wasn't bad github etiquette to leave the issue sitting open for so long... I just never could think of an alternative that would simply and effectively handle all scenarios, but without a delay.

PR #2197, opened 2017-02-13
Pure computed briefly serves stale value, when there are no subscriptions to it and when deferUpdates is on.

mbest:
Steps:
1. `git clone https://github.com/d-ph/bug-reproduction-knockout-pure-computed.git`
2. `cd bug-reproduction-knockout-pure-computed`
3. `bower install`
4. Open the `index.html` in Chrome
5. Open dev tools. Go to the Console tab.
6. Run: `knockoutBugReproduction.pageNumber$(3)`

Result:
```
Computed: pageNumber: 3 listener1LocalVarPageNumberPlusTwenty: 0
Listener 1: pageNumber: 3 pageNumberPlusTen: 13 listener1LocalVarPageNumberPlusTwenty: 0
Listener 2: pageNumber: 5 pageNumberPlusTen: 13 listener1LocalVarPageNumberPlusTwenty: 23
Computed: pageNumber: 5 listener1LocalVarPageNumberPlusTwenty: 23
Listener 3: pageNumber: 5 pageNumberPlusTen: 15 listener1LocalVarPageNumberPlusTwenty: 23
Computed: pageNumber: 5 listener1LocalVarPageNumberPlusTwenty: 23
Listener 1: pageNumber: 5 pageNumberPlusTen: 15 listener1LocalVarPageNumberPlusTwenty: 23
Listener 2: pageNumber: 5 pageNumberPlusTen: 15 listener1LocalVarPageNumberPlusTwenty: 25
Computed: pageNumber: 5 listener1LocalVarPageNumberPlusTwenty: 25
Listener 3: pageNumber: 5 pageNumberPlusTen: 15 listener1LocalVarPageNumberPlusTwenty: 25
```
Expected result:
```
Computed: pageNumber: 3 listener1LocalVarPageNumberPlusTwenty: 0
Listener 1: pageNumber: 3 pageNumberPlusTen: 13 listener1LocalVarPageNumberPlusTwenty: 0
Listener 2: pageNumber: 5 pageNumberPlusTen: 15 listener1LocalVarPageNumberPlusTwenty: 23
Computed: pageNumber: 5 listener1LocalVarPageNumberPlusTwenty: 23
Listener 3: pageNumber: 5 pageNumberPlusTen: 15 listener1LocalVarPageNumberPlusTwenty: 23
Computed: pageNumber: 5 listener1LocalVarPageNumberPlusTwenty: 23
Listener 1: pageNumber: 5 pageNumberPlusTen: 15 listener1LocalVarPageNumberPlusTwenty: 23
Listener 2: pageNumber: 5 pageNumberPlusTen: 15 listener1LocalVarPageNumberPlusTwenty: 25
Computed: pageNumber: 5 listener1LocalVarPageNumberPlusTwenty: 25
Listener 3: pageNumber: 5 pageNumberPlusTen: 15 listener1LocalVarPageNumberPlusTwenty: 25
```
More info:
The third line of the output shows, that for a brief moment, the `pageNumberPlusTen` observable yielded 13 instead of 15. This is not the case, when the `ko.options.deferUpdates` is `false`, or when there is at least one subscriber to that observable (you can test it, by uncommenting `knockout-bug-reproduction.js:44-46`). There is also no issue, if `ko.computed` is used instead of `ko.pureComputed`.

I think, that computed observables should yield the latest values no matter what. I believe the update is scheduled due to `deferUpdates` being on, and that's why the value is correct later.

For now, I'm using `ko.computed`s as a workaround.

### DIFF
--- a/spec/asyncBehaviors.js
+++ b/spec/asyncBehaviors.js
@@ -989,7 +989,7 @@ describe('Deferred', function() {
         });
 
         it('Should *not* delay update of dependent deferred computed observable', function () {
-            var data = ko.observable('A'),
+            var data = ko.observable('A').extend({deferred:true}),
                 timesEvaluated = 0,
                 computed1 = ko.computed(function () { return data() + 'X'; }).extend({deferred:true}),
                 computed2 = ko.computed(function () { timesEvaluated++; return computed1() + 'Y'; }).extend({deferred:true}),
@@ -1008,6 +1008,24 @@ describe('Deferred', function() {
             expect(computed2()).toEqual('BXY');
             expect(timesEvaluated).toEqual(2);      // Verify that the computed wasn't evaluated again unnecessarily
             expect(notifySpy.argsForCall).toEqual([ ['BXY'] ]);
+        });
+
+        it('Should *not* delay update of dependent deferred pure computed observable', function () {
+            var data = ko.observable('A').extend({deferred:true}),
+                timesEvaluated = 0,
+                computed1 = ko.pureComputed(function () { return data() + 'X'; }).extend({deferred:true}),
+                computed2 = ko.pureComputed(function () { timesEvaluated++; return computed1() + 'Y'; }).extend({deferred:true});
+
+            expect(computed2()).toEqual('AXY');
+            expect(timesEvaluated).toEqual(1);
+
+            data('B');
+            expect(computed2()).toEqual('BXY');
+            expect(timesEvaluated).toEqual(2);
+
+            jasmine.Clock.tick(1);
+            expect(computed2()).toEqual('BXY');
+            expect(timesEvaluated).toEqual(2);      // Verify that the computed wasn't evaluated again unnecessarily
         });
 
         it('Should *not* delay update of dependent rate-limited computed observable', function() {
--- a/src/subscribables/dependentObservable.js
+++ b/src/subscribables/dependentObservable.js
@@ -162,7 +162,7 @@ var computedFn = {
         for (id in dependencyTracking) {
             if (Object.prototype.hasOwnProperty.call(dependencyTracking, id)) {
                 dependency = dependencyTracking[id];
-                if (dependency._target.hasChanged(dependency._version)) {
+                if ((this._evalDelayed && dependency._target._notificationIsPending) || dependency._target.hasChanged(dependency._version)) {
                     return true;
                 }
             }


mbest:
I've attached the fix here.

d-ph:
Thanks. I tested my use case against branch `2197-deferred-pure-computed-update` and it works fine now.
```
Computed: pageNumber: 3 listener1LocalVarPageNumberPlusTwenty: 0
Listener 1: pageNumber: 3 pageNumberPlusTen: 13 listener1LocalVarPageNumberPlusTwenty: 0
Computed: pageNumber: 5 listener1LocalVarPageNumberPlusTwenty: 23
Listener 2: pageNumber: 5 pageNumberPlusTen: 15 listener1LocalVarPageNumberPlusTwenty: 23
Computed: pageNumber: 5 listener1LocalVarPageNumberPlusTwenty: 23
Listener 3: pageNumber: 5 pageNumberPlusTen: 15 listener1LocalVarPageNumberPlusTwenty: 23
Computed: pageNumber: 5 listener1LocalVarPageNumberPlusTwenty: 23
Listener 1: pageNumber: 5 pageNumberPlusTen: 15 listener1LocalVarPageNumberPlusTwenty: 23
Computed: pageNumber: 5 listener1LocalVarPageNumberPlusTwenty: 25
Listener 2: pageNumber: 5 pageNumberPlusTen: 15 listener1LocalVarPageNumberPlusTwenty: 25
Computed: pageNumber: 5 listener1LocalVarPageNumberPlusTwenty: 25
Listener 3: pageNumber: 5 pageNumberPlusTen: 15 listener1LocalVarPageNumberPlusTwenty: 25
```
Looking at the output, don't you think, that the computed's evaluator function is run too often (the occurrences of the `Computed: ...` lines)? I mean, the computed's value stabilizes very soon, so I'd expect, that a cached value would be returned after the value becomes 15 (i.e. the evaluator function wouldn't be run after the value becomes 15).

Or maybe it's just me not completely understanding how computeds work behind the scenes. Anyway, if you don't think this is a problem, then I'm happy if you merge PR and close this ticket 👍

abram27:
The added check should potentially include a check on dependency._target.hasChanged(dependency._version) assuming that will prevent the multiple computes @d-ph pointed out.

mbest:
> Looking at the output, don't you think, that the computed's evaluator function is run too often (the occurrences of the Computed: ... lines)?

That's a valid concern. I'm not sure if there's a good way to solve that without breaking something else, though.

singularity-zero:
Would it be much work to push this to npm, at least as a rc? It's a small bug, but it's produced some very confusing issues for my team lately. Either way, thanks so much for the hard work on ko! This is the first bug we've ever experienced, which is a testament to the quality of the code.

d-ph:
I think @mbest will release 3.4.2 very soon now.

In the meantime, you can always install specific commit hash from github using npm.

E.g. instead of:
```
"dependencies": {
  "knockout": "^3.4.1"
}
```
Use:
```
"dependencies": {
  "knockout": "https://github.com/knockout/knockout.git#d57baad844d3ee629e93cb56ce40f643bf94649c"
}
```
Not ideal, but can get you going without waiting for anyone.

singularity-zero:
Thanks for the thought @d-ph 👍 . I didn't realise you could point to commits on github. Given the build files aren't committed, I guess the only alternative is to build it locally and commit to our source control for now.

codymullins:
@d-ph @singularity-zero it's because you need to run the grunt build to get the output. I don't think @mbest checked in the build/output folder....I've also had issues because some npm dependencies try to check the version too.

mbest:
@singularity-zero 3.4.2 is now released.

PR #2210, opened 2017-02-27
Performance enhancement and better CommonJS support

jhorowitz-firedrum:


mbest:
We don't want to make a hard dependency on jQuery.

brianmhunt:
I think it's safe to say this will not be merged because of the jquery deoendency - not to dissuade conversation on any points that might bear reopening this!

PR #2215, opened 2017-03-24
Array Remove and Observable Cascading

mbest:
This may not be a bug per se, but it's an edge case that I'd like to submit for consideration.  The essential parts of our case were:

 - Stored objects in a KO array
 - Displayed them in a specific order
 - Occasionally, removed some objects in response to server-side events

... and the essential parts of our (typescript) implementation were:

```
    public regularTasks:KnockoutObservableArray<TaskComponent>;

    public taskUpdateEvent = (data:any):void => {
        if (data['type'] == 'DELETE') {
            this.regularTasks.remove((item) => {
                // This function manages many-to-many relationships.
                // Its return value indicates that the object has no more relations.
                return item.checkForDeletion(data)
            });
    };

    // referenced by HTML code
    public _regularTasksSorted = () => {
        return this.regularTasks().sort(
            function(a:any,b:any):number{
                // The key is that this is computed from the many-to-many relationships.
                // A simple example would be a count of related objects.
                return a.priority()-b.priority()
            }
        );
    };
```

I was troubleshooting a bizarre issue where the wrong items were being removed, exemplified by debug sequences like:

```
initial:
 - generic_commit
 - assess
analyzing for removal: generic_commit
... not eligible for removal
analyzing for removal: assess
... removing
removed:
 - assess
retained:
 - assess
```

After much pain, I figured out that the sort function was to blame.  If I sort and return a **copy** of the array (i.e. `concat` or `slice`), the problem disappeared.  This leads me to believe that the `remove` function is using indexes that are corrupted by the sort.  Presumably this could be reproduced by a trivial example where the remove function deliberately changes indexes.

Certainly a "safer" (if less performant) option would be to track these relationships (in `remove`) using an object reference.

### DIFF
--- a/spec/observableArrayBehaviors.js
+++ b/spec/observableArrayBehaviors.js
@@ -193,6 +193,20 @@ describe('Observable Array', function() {
         expect(notifiedValues).toEqual([[x]]);
     });
 
+    it ('Should throw an exception if matching array item moved or removed during "remove"', function () {
+        testObservableArray(["Alpha", "Beta", "Gamma"]);
+        notifiedValues = [];
+        expect(function () {
+            testObservableArray.remove(function (value) {
+                if (value == "Beta") {
+                    testObservableArray.splice(0, 1);
+                    return true;
+                }
+            });
+        }).toThrow();
+        expect(testObservableArray()).toEqual(["Beta", "Gamma"]);
+    });
+
     it('Should notify subscribers on replace', function () {
         testObservableArray(["Alpha", "Beta", "Gamma"]);
         notifiedValues = [];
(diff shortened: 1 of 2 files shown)


mbest:
I've attached the code changes here.

ambsw-technology:
> On another note, this also demonstrates an advantage of asynchronous updates. Adding ko.options.deferUpdates = true; to the above example will make it work properly.

This is actually how Python's Twisted works -- the async framework I'm most familiar with.  You're absolutely right that this is better because it ensures that `remove` (and similar operations) is atomic.  Is there a reason that this is not the default?

codymullins:
@ambsw-technology backwards compatibility for people who built apps prior to deferred updates.

PR #2217, opened 2017-03-29
Fixed memory leak of previousContents when tracking array changes and disposing the subscription

### LINKED ISSUE #2216: Observable array leak last value when subscribing with 'arrayChange' and calling dispose.
Observable arrays seem to leak the last known value after calling dispose. I should add, that this only happens if a reference to the observableArray is kept. I would actually expect that the last known value is disposed when calling dispose() and not when the observableArray is garbage collected.

To reproduce the issue you can use the following snippet (tested with knockout 3.4.2). After running it, run garbage collection, take a memory snapshot, you will see that one GCLeakMarker object has been leaked.
```javascript
    function GCLeakMarker() {};

    window.test = ko.observableArray();
    var subscription1 = test.subscribe(function () {}, null, 'arrayChange');
    test([new GCLeakMarker()])
    subscription1.dispose();
    subscription1 = null;
    test([]);

    // After running GC, a CGLeakMarker() is still alive...
```

decden:
The observableArray did leak the last value held by the observable when
subscribing for array change events, and then disposing the subscription.

The vlaue was being held by the closure held by the
arrayChangeSubscription. Setting it to null allows the GC to collect the
closure and the previousContents variable.

This commit fixes #2216

### DIFF
--- a/src/subscribables/observableArray.changeTracking.js
+++ b/src/subscribables/observableArray.changeTracking.js
@@ -36,7 +36,8 @@ ko.extenders['trackArrayChanges'] = function(target, options) {
                 target['notifySubscribers'] = underlyingNotifySubscribersFunction;
                 underlyingNotifySubscribersFunction = undefined;
             }
-            arrayChangeSubscription.dispose();
+            if (arrayChangeSubscription)
+	        arrayChangeSubscription.dispose();
             arrayChangeSubscription = null;
             trackingChanges = false;
         }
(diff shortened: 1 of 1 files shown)


mbest:
Looks good. Thanks.

codymullins:
I must be one of the few that shudder at if statements without brackets ;-)

mbest:
@codymullins Knockout is inconsistent regarding braces, but I've been trying to make sure all new code uses them. @decden Can you add the braces?

PR #2221, opened 2017-04-08
hasFocus alias do not work with the expressionRewrite for property bindings

mbest:
Took me some time to figure this out, but I have always used hasFocus binding instead of hasfocus. I now discovered that hasFocus is an alias. This makes the following code not work.

https://github.com/knockout/knockout/blob/master/src/binding/defaultBindings/hasfocus.js#L25

Because the binding is hasFocus for those using the alias. 

Not sure what the best fix would be at this stage.


navels:
Until the next release, can we get a prominently-displayed note in the docs that you need to use "hasfocus" if you want 2-way databinding?  This cost us some time trying to figure out.

PR #2218, opened 2017-04-09
Select element doesn't update properly when combined with if binding

mbest:
Given the following code:

```html
<div data-bind="ifnot: hidden">
  <select data-bind="options: letters, value: letter, optionsCaption: 'Select a letter'"></select>
  <button data-bind="click: hide">hide</button>
</div>

<button data-bind="click: reset">reset</button>
```

```javascript
  this.hidden = ko.observable(false);
  this.letters = ['a', 'b', 'c'];
  this.letter = ko.observable();
  
  this.reset = function() {
    this.hidden(false);
    this.letter(null);
  };
  
  this.hide = function() {
    this.hidden(true);
  };
```

If you select an option, click hide, then click reset, the select element still shows the option you had selected. Well, it actually will show the placeholder text for a fraction of a second, and then go back to the value you had previously selected. The actual value of the observable will correctly be cleared out, but the select element will show that option as being selected.

Using the `visible` binding instead of `if` will avoid the issue, as will adding `allowValueUnset: true`.

Here is an example: http://codepen.io/caseyjacobson/pen/wJQBpZ

### DIFF
--- a/spec/defaultBindings/valueBehaviors.js
+++ b/spec/defaultBindings/valueBehaviors.js
@@ -360,6 +360,11 @@ describe('Binding: Value', function() {
             observable("");
             expect(testNode.childNodes[0].selectedIndex).toEqual(0);
 
+            // Also check that the selection doesn't change later (see https://github.com/knockout/knockout/issues/2218)
+            waits(10);
+            runs(function() {
+                expect(testNode.childNodes[0].selectedIndex).toEqual(0);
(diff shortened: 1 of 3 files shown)


mbest:
I've attached a fix. Besides fixing this problem, I also made the `setTimeout` only run on IE 6, since it's meant only to work around a problem in that browser.

PR #2024, opened 2017-04-15
A way to extend default binding context

mbest:
Say, for instance, I have binding expressions depending on a library other then `ko` (e.g. `lodash`).

I wish I could just do this:

``` js
ko.applyBindings(viewModel, element, {
    _: require('lodash')
});
```

to make `_` available in `data-bind` expressions.

Currently available options are:
1. Export dependency as a property on view model then use `let` binding to extend binding context

``` js
ko.applyBindings({
    lodash: require('lodash'),
    actualModel: model
}, element);
```

``` html
<!-- ko let: { _: lodash }, with: actualModel -->
<!-- view... -->
<!-- /ko -->
```

This just bloats the code.
2. Export via browser globals `window._ = require('lodash');`. This is just terrible.
3. Use `ko.bindingContext`:

``` js
var ctx = new ko.bindingContext(viewModel).extend({ _ : require('lodash') });
ko.applyBindings(ctx, element);
```

This is bad too, since we now depend on `ko` internals (can we make `bindingContext` a part of public API?)

### DIFF
--- a/spec/bindingAttributeBehaviors.js
+++ b/spec/bindingAttributeBehaviors.js
@@ -29,7 +29,7 @@ describe('Binding attribute syntax', function() {
                 didInit = true;
             }
         };
-        testNode.innerHTML = "<div id='testElement' data-bind='test:123'></div>";
+        testNode.innerHTML = "<div id='testElement' data-bind='test'></div>";
         ko.applyBindings(suppliedViewModel);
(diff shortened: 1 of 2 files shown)


mbest:
I've attached the code here for a change to `ko.applyBindings` where the third argument can be a callback function to modify the binding context. What do you guys think?

Issue #2236, opened 2017-05-03
IE "event" binding for checkbox input


sharpcoder7:
Hello!
1) Bind to event like:
`<input type="checkbox" data-bind="checked: value, event: { change: watcher }" />`

2) After check we go to watcher func

2) In watcher func if checkbox was changed to **unchecked** in chrome and firefox **model.value** value is set as **false** but in IE (including edge) **model.value** is **true**.

So as I understood in IE the event fired **before** model was changed but in other browers - **after**.

sharpcoder7:
The problem is KO binds **changed** event to **onclick** DOM event.
IE has different order while firing the events.
You can check this using specified code in differen browsers:

`<input type="checkbox" onchange="console.log('changed')" onclick="console.log('clicked')">`

sharpcoder7:
I have create new pull requres what fix this issue

PR #2237, opened 2017-05-03
issue #2236 fix

sharpcoder7:
see issue #2236


sharpcoder7:
This will not work in IE < 9. So I have to close this request.

PR #2240, opened 2017-05-29
deferred updates: only notify computed changed if it was evaluated

mbest:
Currently, if a computed has a non-primitive value, it will always notify a change if it was marked as dirty. It should only notify a change if it was actually recalculated. Thus it's not sufficient to base the notify decision solely on the value, as is currently done.


mbest:
I've attached the changes here.

PR #2248, opened 2017-06-02
Export ko.utils.cloneNodes

dregre:
`ko.utils.cloneNodes()` is a pretty useful method, but is unexported; let's export it.

### DIFF
--- a/spec/utilsDomBehaviors.js
+++ b/spec/utilsDomBehaviors.js
@@ -93,3 +93,25 @@ describe('registerEventHandler', function() {
         expect(eventFired && !jQueryModified).toBe(true);
     });
 });
+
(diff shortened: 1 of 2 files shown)


mbest:
If you add a spec to spec/utilsDomBehaviors.js, I'll merge this.

dregre:
Thanks @mbest for the quick reply!  I've added the spec.
I've rebased so be sure to reset locally.

codymullins:
@dregre typically you should not rebase branches that have been made "public"

dregre:
My bad, force of habit @codymullins
The other commit was a one-liner, should be ok if it makes its way back to the commit history.

PR #2265, opened 2017-07-07
Add CopelexValue binding for select element.

alireza-salemian:
This is very useful in most scenarios. For example, when creating a data grid with editable rows or inline editable form element.


IanYates:
I suspect the maintainers will ask for this to be an add-in rather than built into KO core.

With that said, what is this trying to do?  Is it that I can pass in some custom functions to do the text & value setting for my options?

chrisknoll:
Maybe the .vs/* files should not be part of this PR...

Issue #2196, opened 2017-02-03
observableArray().length is returning 0 in ko.Computed()


michael-letcher:
Wondering if this is intentional?

Calling for the length if an observableArray inside of a ko.Computed will always return 0.

chrisknoll:
Sounds odd, but without some code to see what you're talking about, it's hard to judge.  

But at the sound of it, if you create a new observableArray() it defaults to an empty array so the length of it will be 0.  But I'm not sure if that's exactly the case you're talking about.  So please elaborate.

michael-letcher:
@chrisknoll Ah, after a play around we have found the cause is the use of `observableArray.push(item)`

calls the `ko.Computed` twice in which the length will be 0 then the actual number, which doesn't work for us.

If we change the code as per the docs (http://knockoutjs.com/documentation/observableArrays.html) to `observableArray().push(item)` the  `ko.Computed` is called but it breaks the .length completely.

http://jsfiddle.net/bxfXd/8399/

In this fiddle on line 9 `self.items().push("hello");`

Adding elements doesn't work, but if you change it back to `self.items.push("hello");`.
It works.

mbest:
Almost always you should use `observableArray.push` and not `observableArray().push`.

chrisknoll:
When you called items().push, you're unwrapping the underlying array from the observable, and ten pushing something into it. The computed won't be notified of the change because the observable wasn't modified.

calling items.push() uses the observable's interface into the array, so notifications will be aware.

I am wary of observableArray.push() because it causes excessive signals of updates to listeners for each push() (ie: you are inserting a large set of elements into an array).

I think the proper patter in this case is:
var myArray = obsArray();
// add/delete.modify myArray.
obsArray(myArray); // assigns the updated arrray into the observable.

pure computeds that pulled the array out of the observable should be notified once.

michael-letcher:
@mbest I suggest changing the DOCs to not say 

> The syntax is more convenient. To call KO’s push method, just write myObservableArray.push(...). This is slightly nicer than calling the underlying array’s push method by writing myObservableArray().push(...).

When theses will actually do different things.

If it is recommend to `observableArray.push()` this should also be said and why.

mbest:
@michael-letcher Good point. I'll reopen this as a documentation issue.

arturmiz:
@mbest Is the documentation maintained somewhere as an open source repository?

mbest:
[...]

PR #2267, opened 2017-07-10
Documentation update for observableArray push method

arturmiz:
Documentation fix regarding use of `observableArray.push(...)` vs `observableArray().push(...)` which was misleading. PR related to #2196 issue.


mbest:
It seems that this information is mostly given by the line above it.

> For functions that modify the contents of the array, such as `push` and `splice`, KO's methods automatically trigger the dependency tracking mechanism so that all registered listeners are notified of the change, and your UI is automatically updated.

So I think this can just be extended a little to specify that this won't happen if using `...().push`.

arturmiz:
You're right, I haven't noticed that, hope the fix will do the work.

mbest:
Thanks.

PR #2268, opened 2017-07-11
Add observableArray.replace()

james1293:
People [seem to like this function](https://stackoverflow.com/a/15440969/1630632), so I figured it should be in the docs! :-)
Feedback is welcome.


mbest:
Thanks.

PR #2261, opened 2017-07-19
checked binding (3.5.0-pre)

mbest:
Note - this report is about 3.5.0-pre.

The checked binding appears to be broken in 3.5.0-pre, see http://jsfiddle.net/fastfasterfastest/qyrb146o/1/
It seems so elementary so I don't know if I am having a mind block, it's possible, but the behavior is as expected with 3.4.2.  I am seeing the behavior in a private build from the current state of affairs, but the fiddle is using one of @mbest's builds he made available in another issue a while ago.

The input element does not get checked when the checked binding's value is set to `true`.  It does get checked if the checked binding's value is set to same as value binding's value.

If you change to http://knockoutjs.com/downloads/knockout-3.4.2.js the input element gets checked when the checked binding's value is set to `true`, as expected.

Added: https://github.com/knockout/knockout/pull/1899 seems to be the reason/cause.  Should the discussion be carried on there?

### DIFF
--- a/spec/defaultBindings/checkedBehaviors.js
+++ b/spec/defaultBindings/checkedBehaviors.js
@@ -392,69 +392,74 @@ describe('Binding: Checked', function() {
                 expect(testNode.childNodes[0]).toHaveCheckedStates([false, true]);
             });
 
-            it('Should use that value as the checkbox\'s value when not bound to an array', function () {
-                var myobservable = ko.observable('random value');
-                testNode.innerHTML = "<input type='checkbox' data-bind='checked:someProp, " + binding + ":true' />" +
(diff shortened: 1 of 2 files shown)


fastfasterfastest:
I am not a GitHub expert.  That said, I find it confusing that in this repository issues seem to be converted to pull requests and in doing so they disappear from the Issues section.  Many other repositories don't seem to operate like that.  I regularly look at what new and/or closed issues may have arisen and in this repository one also have to look in Pull Requests to make sure one is not "missing" anything.

mbest:
I like being able to link the code directly to the issue, but I can see that it can create confusion. If you just want to look at issues and pull requests together, here is a link for that: https://github.com/knockout/knockout/issues?q=is%3Aopen

Issue #1695, opened 2015-02-13
Knockout DOM data leaks, if element removed by jQuery


San4es:
See sample: http://jsfiddle.net/6e52o541/

If a DOM node is removed by ko.removeNode, jQuery data for this node is cleaned, but if it removed by $.remove() - it is not.

brianmhunt:
@San4es – Thanks for the report.

Is it possible that the data remains attached after `$.remove` pending garbage collection? That would be my first thought; whereas `ko.removeNode` deletes the data explicitly, jQuery waits for gc. That's just the first theory that came to mind to explain the problem.

Is this issue exhibiting as some sort of problem you are experiencing?

Cheers.

San4es:
@brianmhunt,

These data will not be cleared automatically by GC, while it has alive reference on it.

rniemeyer:
@San4es - I am not sure that I understand the issue clearly. Are you seeing an issue in a certain browser? I see that called `.remove` clears the jQuery specific data in both Chrome and Firefox.  jQuery's `.remove` would have no way to know to remove the data that Knockout adds (`ko.utils.domData`). Unless something alive referenced something in the domData for the element, it would not prevent the element from being garbage collected though. Thanks!

San4es:
@rniemeyer,

"jQuery's .remove would have no way to know to remove the data that Knockout adds (ko.utils.domData)."

Yes, until Knockout does not tell about it to jQuery ;)

I created the ticket to jQuery bug trackerhttps://github.com/jquery/jquery/issues/2088 about lacking 'remove' event and got an answer, that the way to know when node is removed is wrap native $.cleanData method as jQuery UI does.

So, I think, that jQuery is  enough popular library, that Knockout can clean data on jQuery.remove(), if user use jQuery an Knockout in the same time.

rniemeyer:
I am not sure how far down that road that Knockout core would want to go (replacing/wrapping a jQuery method). Is there a reason in your application that you need to use `$.remove` vs. using KO's normal templating / control-flow (which does the clean up for you) or using the `ko.removeNode` API directly? Just want to understanding your scenario fully.

IanYates:
Also, usually in the MVVM pattern it's best/easiest to let the binding engine in use (knockout) control the updating of the DOM, or at least have the updates happen in coordination via binding handlers.  Is it a third party control that's using $.remove?  If so, could you incorporate that cleanup into your binding handler?

SteveSanderson:
[...]

PR #2141, opened 2017-07-21
ko.utils.domData storage scheme requires explicit cleanup

mbest:
This problem has been previously brought up in the rejected Issue #1695. Our case idifferent, however: we have no way of ensuring that the explicit ko.cleanNode() is always invoked for the removed element. 
Also note that jQuery has switched to storing data on the expando - https://github.com/jquery/jquery/issues/1734. And they had long been advocating for the same data storage solution as Knockout's...

Or use case is as follows:
1. We rely on custom element support (part of the web component spec) to activate our composite component (see https://developer.mozilla.org/en-US/docs/Web/API/Document/registerElement)
2. We are using Knockout to manage the composite component's template. 
3. The template will only reference observables stored with the element's properties. No external observables will be used. This means that the custom element should be garbage-collectable when it is removed. 
4. Custom elements are designed to rely on garbage collection for their cleanup, and we cannot require our users to call the explicit cleanup method. 

Unfortunately, Knockout's domData maintains references to to all DOM nodes where bindings were applied, so all DOM nodes within the Knockout template remains in memory heap.

Here is a jsFiddle that demonstrates the issue: https://jsfiddle.net/mstarets/4nfsebec/5/. It needs to be run in Chrome (otherwise you will need to include the Custom Element polyfill). After you run the page, click on 'Create Custom Element button, then on 'Remove Custom Element' button. Repeat this sequence 2-3 times. Open Developer Tools, go to the 'Profiles' tab and take a heap snapshot. Note that there are several hundred of HTMLButtonElements that have been leaked. Now uncomment the line calling clearDataForButtons() and repeat the same test. Note that the button elements are no longer leaked.

### DIFF
--- a/src/utils.domData.js
+++ b/src/utils.domData.js
@@ -4,33 +4,37 @@ ko.utils.domData = new (function () {
     var dataStoreKeyExpandoPropertyName = "__ko__" + (new Date).getTime();
     var dataStore = {};
 
-    function getAll(node, createIfNotFound) {
-        var dataStoreKey = node[dataStoreKeyExpandoPropertyName];
-        var hasExistingDataStore = dataStoreKey && (dataStoreKey !== "null") && dataStore[dataStoreKey];
-        if (!hasExistingDataStore) {
-            if (!createIfNotFound)
-                return undefined;
-            dataStoreKey = node[dataStoreKeyExpandoPropertyName] = "ko" + uniqueId++;
-            dataStore[dataStoreKey] = {};
-        }
-        return dataStore[dataStoreKey];
-    }
-
-    return {
-        get: function (node, key) {
-            var allDataForNode = getAll(node, false);
-            return allDataForNode === undefined ? undefined : allDataForNode[key];
-        },
-        set: function (node, key, value) {
-            if (value === undefined) {
-                // Make sure we don't actually create a new domData key if we are actually deleting a value
-                if (getAll(node, false) === undefined)
-                    return;
+    var getDataForNode, clear;
+    if (window['WeakMap']) {
+        getDataForNode = function (node, createIfNotFound) {
+            var ownerDoc = node.ownerDocument,
+                dataStore = ownerDoc[dataStoreKeyExpandoPropertyName] || (ownerDoc[dataStoreKeyExpandoPropertyName] = new ownerDoc.defaultView['WeakMap']());
+            if (dataStore['has'](node)) {
+                return dataStore.get(node);
             }
-            var allDataForNode = getAll(node, true);
-            allDataForNode[key] = value;
-        },
-        clear: function (node) {
+            if (createIfNotFound) {
+                var dataForNode = {};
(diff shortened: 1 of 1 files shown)


brianmhunt:
Thanks @mbest — I've opened an issue in TKO and will mull & merge a commit.

src/utils.domData.js
```diff
@@ -4,41 +4,58 @@ ko.utils.domData = new (function () {
...
-        set: function (node, key, value) {
-            if (value === undefined) {
-                // Make sure we don't actually create a new domData key if we are actually deleting a value
-                if (getAll(node, false) === undefined)
-                    return;
+    var getDataForNode, clear;
+    if (window['WeakMap']) {
+        getDataForNode = function (node, createIfNotFound) {
+            var ownerDoc = node.ownerDocument,
+                dataStore = ownerDoc[dataStoreKeyExpandoPropertyName] || (ownerDoc[dataStoreKeyExpandoPropertyName] = new ownerDoc.defaultView['WeakMap']());
```
brianmhunt: Good job finding this _**edge**_ case (pardon the pun)

mbest:
Maybe it's just simper to use @mstarets's technique of storing the data on the node itself. The old method would only need to be used in old-IE.

brianmhunt:
Yeah an IE test is simpler than this ownerdox logic.  Does latest Edge fail too? Is it a known issue?

mbest:
> Is it a known issue?

There doesn't seem to be a report of it.

https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/?page=1&q=weakmap

mbest:
This version doesn't use `WeakMap`. What do you think?

brianmhunt:
I think this is the most practical approach.

I think we should be explicit about what happens with Edge.  I wonder if there's a bug report / reporting facility / this would be considered a bug.

src/utils.domData.js
```diff
@@ -4,41 +4,56 @@ ko.utils.domData = new (function () {
...
-            var allDataForNode = getAll(node, false);
-            return allDataForNode === undefined ? undefined : allDataForNode[key];
-        },
-        set: function (node, key, value) {
-            if (value === undefined) {
-                // Make sure we don't actually create a new domData key if we are actually deleting a value
-                if (getAll(node, false) === undefined)
-                    return;
+    var getDataForNode, clear;
+    if (!ko.utils.ieVersion) {
```
brianmhunt: Perhaps this should be 

```javascript
   if (!ko.utils.ieVersion && 'WeakMap' in window) {
```

brianmhunt: We should also probably note the bug report here or explain why IE's `WeakMap` is broken.

mbest:
I've reported the `WeakMap` issue: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/12891461/

brianmhunt:
Noting Microsoft/ChakraCore#2983

Issue #1269, opened 2014-01-03
Issue with select element binding using options, optionsValue, and value


loesak:
Knockout 3.0.0

Assume the following:

I have a complex object (lets say Person) who has a reference variable (favoritePet) to another complex object (Pet). The value that can be held for the 'favoritePet' references comes from an asynchronous HTTP request. Meaning, if a value is populated in the 'favoritePet' reference, then the object selected is equivalent to another object in the fetched Pet instances, but is not the same object in memory. 

According to the documentation, I need to provide the binding for 'optionsValue' and I assumed this would work:

```
<select class="form-control" data-bind="{
        'options' : availablePets,
        'optionsText' : function(item) { 
            return item.name; 
        },
        'optionsValue' : function(item) { 
            return item.id; 
        },
        'value' : favoritePet,
        'optionsCaption' : 'Choose ...'
    }"></select>
```

However, I found that it does the following:
- when selecting a value, tries to assign the the selected pets 'id' to the reference variable instead of the pet instance itself
- when a pet instance is already assigned to the reference variable, does not show that value initially selected when the view is bound

I dug through the code and I believe this is because the code upon section, tries to determine the data object that was selected by comparing the option elements value with the list of available data objects and upon setting selection, tries to compare the elements option text with the observable value. So in both instances trying to compare the option elements value (a string) with the object (an object), which will never be equal.

I can resolve the second problem by changing the data-binding to: 

```
<select class="form-control" data-bind="{
        'options' : availablePets,
        'optionsText' : function(item) { 
            return item.name; 
        },
        'optionsValue' : function(item) { 
            return item.id; 
        },
        'value' : favoritePet().id,
        'optionsCaption' : 'Choose ...'
    }"></select>
```

Now when a value is assigned to the reference variable, then it can properly resolve which option to select by default because the assigned value for the binding is the observable values 'id' field, which eventually is equivalent to an the 'id' field of one of the instances of available pets.

However, this causes a different problem when there is no value assigned to the 'favoritePet' reference variable as 'favoritePet()' resolves to null, throwing an error.

I believe all this is again due to the code trying to determine the appropriate model to both select and assign by checking the equivalence of the value and the 'optionsValue' of each available data object.

Instead i think the code should be checking for equivalency between the values after begin passed through the 'optionsValue' method. Meaning this binding is valid:

```
<select class="form-control" data-bind="{
        'options' : availablePets,
        'optionsText' : function(item) { 
            return item.name; 
        },
[...]

PR #2273, opened 2017-07-24
Made it possible to use list from two different sources.

JesperKoldHansen:
This commit, makes it possible to use checked and checkedValue, where the items in checked comes from a different source then checkedValue, but are inherently the same.


mbest:
Similar to #1269. I don't think the bindings should be enforcing or coercing this kind of "equality". Instead, you should do so in your model initialization code.

Issue #2252, opened 2017-06-06
toString invocation on subscribable instance throws Exception


Lucasus:
**To reproduce try to invoke**
`new ko.subscribable().toString()`

**Expected**
Method should return `"[object Object]"` (this is the behavior at least in Knockout 3.0.0, any other return value will actually work as well)  

**Actual**
Following exception is thrown:
`
Uncaught TypeError: Function.prototype.toString is not generic
    at v.subscribable.toString (<anonymous>)
    at <anonymous>:1:23
`

**Context of this bug**
I'm in the process of upgrading Knockout from version 3.0.0 to 3.4.2 for large code base and I stumbled on this issue. We're using knockout.mapping plugin and it's `ko.mapping.toJS` method breaks with exception above every time we're passing some viewmodel with `ko.subscribable` in it (and we had such cases)

**Workaround**
I didn't want to modify Knockout library itself, so I've added in our project following code after Knockout.js is loaded. As you see, implementation is far from ideal:

> ko.subscribable['fn'].toString = function () {
    return "[object Object]";
}

krnlde:
What happens when you do `Object.prototype.toString.call(new ko.subscribable());`?

Lucasus:
@krnlde  `[object Object]` is returned, as it should

krnlde:
Does this solve the issue for you?

Lucasus:
@krnlde, I've already solved this issue in my code base by using workaround I've described. Your code statement will work as well if used as replacement for my hard coded `"[object Object]"` return value . Nevertheless, it solves symptoms, not the root cause (which is the `toString` exception)

orbweaver-:
PR #2279

mbest:
I've merged #2279

PR #2279, opened 2017-08-09
Add toString method to subscribable so it doesn't error out

orbweaver-:
from issue #2252 
Simply returns the expected string "[object Object]"


mbest:
Thanks.

PR #2283, opened 2017-08-19
Fix side effects on awake callback

dancras:
We are starting to use pure computed asleep / awake events a lot more to implicitly optimise resource consumption and have stumbled across this bug:

> If a pure computed in a chain of pure computeds has a subscription to its awake event which causes side effects that update its value, it is possible for listeners to get into a bad state. 
> 
> This only happens if the pure computed in question is invoked before being subscribed to.

Added specs to capture this case, and a possible solution by checking if dependencies have changed one more time after setting up subscriptions to them.

### DIFF
--- a/spec/pureComputedBehaviors.js
+++ b/spec/pureComputedBehaviors.js
@@ -382,6 +382,27 @@ describe('Pure Computed', function() {
         expect(timesEvaluated).toEqual(2);
     });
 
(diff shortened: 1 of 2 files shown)


dancras:
Took me a moment to notice I had made the changes from a very old master base. Fixed that now.

mbest:
I've merged this with some tweaks (3949c3e067636f90fcf3abfc1dd9e9b0aae9b308). Thanks!

PR #2290, opened 2017-09-06
Checking if state exists when disposing

andersnygaard:
I've encountered cases where we call
`bindingContext.createChildContext(viewModel);`
and when dispose function is called internally, state is not found

This probably doesn't fix the problem, but it fixes the symptoms. Feel free to investigate more, because my knowledge of the internal workings of Knockout is limited. The error seems to be introduced in version 3.4.0.

### DIFF
--- a/src/subscribables/dependentObservable.js
+++ b/src/subscribables/dependentObservable.js
@@ -384,7 +384,7 @@ var computedFn = {
     },
     dispose: function () {
         var state = this[computedState];
-		if (typeof(state) === "undefined") {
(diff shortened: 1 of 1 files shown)


mbest:
I don't see why `state` would ever be undefined.

brianmhunt:
I think that the issue here should be better identified, i.e. in a jsFiddle during a legitimate use case.

I'm closing this just for cleanup, pending an example that illustrates that this can occur.  If there is one @andersnygaard, please post and we'll re-open and can merge!

Issue #2294, opened 2017-09-22
Binding context exports unnecessary properties which could shadow globals.


bennieswart:
In the file bindingAttributeSyntax.js a `_subscribable` property is created on a binding context under certain conditions. During build, this property is minified and in the 3.4.2 release it is exported as `Q`. See this jsfiddle for an example: https://jsfiddle.net/s0wf1hga/1/. Note the errors in console.

This becomes a problem when a global property with the same name exists and referenced from inside bindings. Granted, the chance of a name conflict is rare. I did, however, encounter a problem with a custom build I was making where `_subscribable` was minified to `$`, which shadowed jquery in all bindings and broke a lot of things.

The workarounds for this problem are many, I know, but rather than a workaround I feel it is more right and rather easy to remove the problem. I'm also not saying that all unnecessary properties everywhere should be hidden, but the binding context is a rather unique situation with more ramifications than usual.

mbest:
Those are good points. Perhaps we shouldn't minify `_subscribable`.

bennieswart:
Not minifying `_subscribable` is the workaround I ended up using.
Ideally I suppose one would want to 'hide' the property on the instance, for example by using Symbols. Is there a way to do this across all the currently supported browser versions?
If not I'd be happy to submit a PR which prevents `_subscribable` from being minified, as well as a test case if you think it useful.

mbest:
We could use a `Symbol` in browsers that support it and a non-minified property on the others. We have that already for `ko.computed`: https://github.com/knockout/knockout/blob/master/src/subscribables/dependentObservable.js#L1

PR #2296, opened 2017-09-23
Hide or don't minify extra internal properties on binding context

bennieswart:
It is possible for properties such as `_subscribable` to be created on a
binding context, which, after minification, could be renamed to any
arbitrary character. In most cases this is not an issue, except when a
global variable with the same name exists, however rare the case may be.
This causes the global variable to be shadowed in all bindings, which is
especially problematic in the case of, say, `$`, which breaks jquery.

To fix this, either hide such properties with a Symbol, or simply don't
minify them so that the name is constant and predictable.

Fixes https://github.com/knockout/knockout/issues/2294

### DIFF
--- a/spec/bindingAttributeBehaviors.js
+++ b/spec/bindingAttributeBehaviors.js
@@ -229,6 +229,19 @@ describe('Binding attribute syntax', function() {
         expect(testNode).toContainText("my value");
     });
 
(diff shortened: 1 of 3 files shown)


mbest:
Thanks the contribution. This looks good!

Issue #1533, opened 2014-09-10
feature request: add afterRender event for knockout components


SuperCuke:
Hello,

I propose to add an afterRender event to components binding - the event to be fired after viewModel and template are loaded and bound. Event will be useful for example to make area visible after component is loaded.

Best regards

yohanliyanage:
+1

Having this event would be useful IMO.

SteveSanderson:
Yes, I see this would be useful. To make the behaviour more precisely self-descriptive, I suggest the callback would be called something like `componentWasDisplayed`, an optional method on your component viewmodel that KO invokes only once when the template is first bound.

It's important to note that this would not have anything to do with child components, as it's much less clear what it means to say the child components are loaded or displayed (the set of child components can change over time, and in general considering that component loading is asynchronous, there's no way [even theoretically](http://en.wikipedia.org/wiki/Halting_problem) to know when you've finished adding child components, unless someone can give a more useful definition for "child components have finished loading" than I can think of presently).

brianmhunt:
It would be possible, though arguably not practical, to have a queue (or `Set`) of components, with a component being added upon construction, removed just after its bindings are applied (or some hook that defaults to binding, but may be extended to cover other user-land asynchronous elements).

If components are the only asynchronous part of the process, then watching for the queue being empty *should* cover the component binding process.

Not only is this probably not performant, but it is also very "rough", in the sense that there is no way to easily track the component hierarchy completion – one can only discern whether some component, somewhere, is between starting and binding. Building a hierarchy would likely add substantial performance overhead.

In the ordinary course, this may differ from the halting problem because the component lifecycle has a definite start (construction) and end (binding). It is correct to say that where there is an aberration like a browser error, there may be no end – which is the halting problem. (For a Javascript error one could set up a timeout to check after a reasonable period for the components that did not complete– but that is way outside what the library ought to do, IMHO.)

Given how "rough" the apparent solution is, and the potential issues with performance, my sense is that `afterRender` is not a feasible built-in for components. However for an actionable item, perhaps hooks that would permit the user to implement something along the lines of the above – e.g. hooks such as `ko.components.beforeViewConstruction` and `ko.components.afterComponentBinding` (which I hope are self-descriptive).

bjornalm:
+1

matthewma:
[...]

PR #2303, opened 2017-10-06
Add component afterRender callback

mbest:
See discussion at #1533. We can discuss this specific implementation here, though.

### DIFF
--- a/src/components/componentBinding.js
+++ b/src/components/componentBinding.js
@@ -1,33 +1,67 @@
 (function(undefined) {
-
     var componentLoadingOperationUniqueId = 0;
+
+    function ComponentDisplayDeferred(element, parentComponentDeferred, replacedDeferred) {
+        var subscribable = new ko.subscribable();
+        this.subscribable = subscribable;
+
+        this._componentsToComplete = 1;
+
+        this.componentComplete = function () {
+            if (subscribable && !--this._componentsToComplete) {
+                subscribable['notifySubscribers'](element);
+                subscribable = undefined;
+                if (parentComponentDeferred) {
+                    parentComponentDeferred.componentComplete();
+                }
+            }
+        };
+        this.dispose = function (shouldReject) {
+            if (subscribable) {
+                this._componentsToComplete = 0;
+                subscribable = undefined;
+                if (parentComponentDeferred) {
+                    parentComponentDeferred.componentComplete();
+                }
+            }
+        };
+
+        if (parentComponentDeferred) {
+            ++parentComponentDeferred._componentsToComplete;
+        }
+
+        if (replacedDeferred) {
+            replacedDeferred.dispose();
+        }
+    }
 
     ko.bindingHandlers['component'] = {
         'init': function(element, valueAccessor, ignored1, ignored2, bindingContext) {
             var currentViewModel,
                 currentLoadingOperationId,
-                currentAfterRenderWatcher,
+                displayedDeferred,
                 disposeAssociatedComponentViewModel = function () {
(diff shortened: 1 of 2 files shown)


brianmhunt:
This implementation will be completely different in tko, but I think the API surface is perfectly replicable.

Is the intention to have `afterRender` passed alongside `params`?  Or have `afterRender` in the `viewModel`?  Or in the `ko.components.register`?  Or all?

src/components/componentBinding.js
```diff
@@ -49,13 +91,27 @@
...
+                    var componentInfo = {
+                        'element': element,
+                        'templateNodes': originalChildNodes
+                    };
+
+                    var componentViewModel = createViewModel(componentDefinition, componentParams, componentInfo),
                         childBindingContext = bindingContext['createChildContext'](componentViewModel, /* dataItemAlias */ undefined, function(ctx) {
                             ctx['$component'] = componentViewModel;
                             ctx['$componentTemplateNodes'] = originalChildNodes;
+                            ctx._componentDisplayDeferred = displayedDeferred;
```
mbest: Should make this a Symbol (see #2296)

mbest:
It's just a method in the viewmodel for now. I'm not attached to calling it `afterRender`. Perhaps we can use this opportunity to define a reserved set of method names. Examples: `ng...` (Angular), `handle...` (JET), `...Callback` (Web Components).

chrisknoll:
If it's another component parameter that sits along side 'param' we can do things like warn about invalid property assignment (I think the only valid things to the component binding is name and params). Making it a function int he view model means you have to do things like reserve function names.  I know that's the pattern we have today with init, update, dispose, etc...but couldn't we consider making a dedicated, optional param to store the callback function that should be invoked after rendering? 

ie:
```
<my-component params="param list", afterRender="theFunction"/>
```
and we can warn about this sort of typo:
```
<my-component params="param list", afterRendered="theFunction"/>
<!-- afterRendered not a valid attribute of component binding -->
```

Also, you don't need to be attached to a function name now, the consumer of the component can decide what they want to call their afterRender handler.

mbest:
@chrisknoll I agree that having access to the callback in the binding is useful, but it seems that most people want to have access from within the component rather than from outside. We could have both, though, like Oracle JET does.

chrisknoll:
I see now.  Up to this point I thought the objective was so that external parties could do something after the component rendered, but I see now that it's meant to allow the component to do something for itself once it is rendered (which could be to make a callback on its own to external observers).  Thanks for the clarification.

codymullins:
Really to me this callback seems like it will be the place to handle most initialization logic. "afterRender" is the "when" it's fired but it probably wouldn't hurt to call this the init like angular $onInit, right?

IanYates:
Having available both to the component, and to its host/parent, sounds like a good approach.  I've got parts of my large codebase where I've had to do one or the other on occasion.   Do you think you'll go that way with this PR?

🎉 
I hacked some afterRender for components into KO enough for my needs but will happily switch to this when it's released.

mbest:
I've created a separate pull request that add an `afterRender` callback for any binding that updates and binds its descendant elements (including `component`): #2310. The "afterRender" defined here is different in that it waits for all child components to finish as well. I think we should call this something different to distinguish it, maybe `afterComplete`. Any other ideas?

mbest:
I've been working on re-implementing this based on the work in #2310. The implementation will be more generic, and you'll be able to attach an event handler from anywhere. There will still be an automatic subscription if the component viewmodel contains a specific function. I'm thinking to call it `onDescendentsComplete` rather than `afterRender`.

mbest:
Replaced by #2319

PR #2310, opened 2017-10-12
add generic afterRender binding

mbest:
that works with any binding that calls `ko.applyBindingsToDescendants` or with just normal binding processing (more tests to come).

### DIFF
--- a/spec/bindingAttributeBehaviors.js
+++ b/spec/bindingAttributeBehaviors.js
@@ -583,4 +583,42 @@ describe('Binding attribute syntax', function() {
             expect(testNode).toContainHtml('<p>replaced</p><template>test</template><p>replaced</p>');
         });
     });
+
+    it('Should call an afterRender callback function after descendent elements are bound', function () {
+        var callbacks = 0,
+            callback = function (nodes, data) {
+                expect(nodes.length).toEqual(1);
+                expect(nodes[0]).toEqual(testNode.childNodes[0].childNodes[0]);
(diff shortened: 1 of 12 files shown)


mbest:
@brianmhunt What do you think of this?

mbest:
For some reason, this pull request hasn't updated with the last commit. Please see https://github.com/knockout/knockout/compare/afterRender-generic

brianmhunt:
I think this is a great idea, and - with one caveat - I think this is the right implementation for knockout 3.x.

The caveat is that `afterRender` should probably in the future also handle asynchronous peer bindings.  I don't think that's feasible in KO 3.x but it is in tko since `applyBindings*` return promises, so we just want to be future-proof.

Because of the way `tko` works, it may be possible for `afterRender` to be a proper binding (i.e. no changes to core), so the core code will be different but the tests should be pretty much the same and I don't think there'd be any issues with backwards compat.

@mbest do you think `afterRender` should always be async?  Or only async when at least one child-binding is async?

Other than that, it looks good.

codymullins:
Am I understanding this correctly that this notifies the parent when the child is rendered? Can the child's "callback" also get called when it is rendered?

mbest:
@codymullins Any binding on the "child" will get updated when it's rendered. It's not difficult to run something on the element itself.

mbest:
After making this work generically, I felt that the `afterRender` term was no longer appropriate. After mulling it over, I've changed it to `childrenComplete`.

Issue #1475, opened 2014-07-31
How to know when a component has finished updating DOM?


PrinceAli321:
Hi Guys,

I need to reflow my page after some KO components have finished rendering in the DOM to support this Foundation plugin: http://foundation.zurb.com/docs/components/equalizer.html

I know we have afterRender for the template and foreach bindings but i can't find anything similar for the new components functionality.

How can i go about executing a callback when my components are loaded (either individually or after all have been loaded)? Is there something i've missed?

W3Max:
+1 for this issue and for @richotaylor comments about child components

jods4:
I solved this problem using an early build of 3.3 and turning all my components synchronous. Other benefits: better perf and less flashing content.

This does not prevent you from dynamically loading component templates, I just do it myself and pass a string to the component declaration. It's only a different way to organize the codebase (and it also makes me think that including `require` logic inside KO components was a bit of a mistake).

W3Max:
@jods4 Could you share a code example of how you do it? I would really appreciate it!

jods4:
Well, in my project I'm using Typescript with requirejs as AMD loader, but in essence it works like this (Sorry if my js syntax for require is not 100% correct, TS makes this so much easier):

Step 1: be sure to have a KO 3.3 build, otherwise this won't work.

Step 2: define your component.
Key points here: I use requirejs to fetch the template as a string and use the option `synchronous: true`.
```javascript
define(['knockout', 'text!template'], function(ko, template) {
  ko.components.register('my-component', {
    viewModel: MyVM,
    template: template,
    synchronous: true
  });
});
```

Step 3: make sure your component is loaded before any view tries to use it. 
You can do as you like, I do so by adding a fake require dependency on any viewmodel that needs a component in its view.
```javascript
define(['MyComponent'], function() {
  /* Define whatever you want that needs MyComponent to be loaded */
});
```

W3Max:
@jods4 Thanks! I will look into ```synchronous: true```

jods4:
@W3Max Note that as of today 3.3 is officially released! Look at @rniemeyer  blog post for more info: http://www.knockmeout.net/2015/02/knockout-3-3-released.html

lzl124631x:
@jods4 @W3Max I want to note that the first time load still might be aync even if you mark the component with `synchronous: true`.

[...]

PR #2319, opened 2017-11-14
component "afterrender" using binding events

mbest:
Adds `descendantsComplete` event that is used by components to notify of completion, waiting until descendant components are also complete.

This replaces #2303.


mbest:
This supports a special callback method on components called `koDescendantsComplete` which is called once all descendant bindings (including async components) are complete.

@brianmhunt I hope you can take some time to look over these changes and provide feedback. Thanks.

brianmhunt:
Excellent 😀

siimnelis:
This is excellent feature that I have been missing.
I have used it now in one of the projects we use knockout and it works great.

There is one issue I'm still having and maybe there is a solution for that also. In our project when we bind our component, everything in the viewmodels markup is wrapped inside a ko if statement.



business logic here
<span data-bind="text:example"></span>



show loading mask


The viewmodels data is loaded async and when the data is received, the viewmodel is initialized in a callback and the isLoaded observable is marked true.

self.getViewModel(function(viewModel){

self.example=ko.observable(viewModel.exaple);

});

Now the viewModel has example obervable and it's safe to bind the markup.

The problem with the koDescendatsComplete binding event is that it fires the first time around and it doesn't know about the nested components inside if loaded(that's expeted). 



business logic here
<span data-bind="text:example"></span>
<nested-async-component-here-which-loads-the-same-way>


What I'm trying to achieve is to mark the parent loaded only when all the nested children are loaded as well and then show the finished markup.

I tried to use the childrenComplete binding handler in this case but it doesn't work the way I hoped.


<div data-bind="childrenComplete:childrenComplete"> 
business logic here
<span data-bind="text:example"></span>
<nested-async-component-here-which-loads-the-same-way>


childrenComplete function runs before the nested-async-component-here-which-loads-the-same-way koDescendatsComplete binding event.

I also tried to subscribe ko.bindingEvent(div, "koDescendantsComplete",function(){}); It throws an error that it can't bind to that nodetype(this was expected as well, was just trying).

Would it be possible to achieve such a binding handler/event which notifies when a binding handlers such  as if, nested components have completed. Maybe there already a solution to achieve similar results already today, that I'm just not aware right now.

Thank you is advance

mbest:
@siimnelis Components are designed so that the viewmodel code and/or template can be loaded asynchronously. The [component loader documentation](http://knockoutjs.com/documentation/component-loaders.html) describes how to do this. Would that work for you?

siimnelis:
The viewmodel code and template both are loaded asynchronously using requirejs as amd. I have looked into [component loader documentation](http://knockoutjs.com/documentation/component-loaders.html). 
The problem comes from that the viewmodels data itself also is loaded asynchronously.

This example is simplification.

```
define(['knockout', 'text!templates/partialViews/example', 'partialViewPage'],
    function (ko, htmlString, partialViewPage) {
        function viewModel(params) {

            var self = partialViewPage("UrlToLoadDataFrom", params);

            self.getViewModel(function (viewModel) {
                self.example= ko.observable(viewModel.example);
            }, {Id:1}); //can pass query string arguments to server

            return self;
        }

        return {
            viewModel: {
                createViewModel: function (params, componentInfo) {
                    params.element = componentInfo.element;
                    return viewModel(params);
                }
            }, template: htmlString
        };
    });
```

The partialViewPage.js adds basic functionality to all asynchronously loaded viewModels.

The partialViewPage wraps components markup inside ko if statement.

` params.element.innerHTML = "<!-- ko if:isLoaded() && !loadingFailed() -->\r<div data-bind='visible:ready, childrenComplete:childrenCompleteCallback'>" + `params.element.innerHTML` + "</div>\r<!-- /ko -->";`

There is the method getViewModel and isLoaded observable and other functionality.

```
self.isLoaded = ko.observable(false);

self.getViewModel = function (callback, data) {
         
//loading data with jquery ajax call on success callback if everthing is okey callback will be invoked with the viewmodel loaded from server and isLoaded is marked true, which renders the actual html

            $.ajax(viewmodelUrl,
                {
                    dataType:'json',
                    data:data,
                    success:function(viewModel) {
                        aetService.resetSessionTimer();
                        
                       callback(viewModel);

                        self.isLoaded(true);

                    }
                });

}
```

I thought about maybe I could load data from server inside createViewModel function and maybe make a custom loader. But the createViewModel must return viewmodel instance synchronously.

```
viewModel: {
                createViewModel: function (params, componentInfo) {
                    params.element = componentInfo.element;
//cant load data async here and pass it to there viewModel instance as a parameter

                    return viewModel(params);
                }
```

I look more into maybe I can somehow still achieve this with custom loader but from the documentation I understand for now that the job of a component loader is to asynchronously supply a template/viewmodel pair for any given component name. Viewmodel being the javascript class, not the data itself.

I hoped I could write custom loading for loadViewModel(name, templateConfig, callback), but the documentation states it must return the viewModel synchronously and thats the problem.

mbest:
This is an interesting approach. I think that rather then the `if` binding, which can be either true or false at any time, we need a binding that encapsulates the idea of async loading. I'm thinking to call it `when`, based on #2113.

brianmhunt:
@mbest We probably also want the contrapositive, an e.g. `until` binding that hides after a promise resolves.

siimnelis:
It's a good idea to have a binding that encapsulates the idea of async loading. 

From the [#2113 ](https://github.com/knockout/knockout/pull/2113) I understand the idea of when is to only run it once. With async loading there is  also a usecase for reloading the view. I can call

`self.getViewModel(function(viewModel){}); //marks isLoading(false) and the loading mask shows up until loaded`

any number of times, which reloads the view.

It's comes handy when a POST request changes server state and I don't have the information how it affected the UI state(some webservice requests). 

Normally I update the UI from POST requests response.

src/binding/bindingAttributeSyntax.js
```diff
@@ -255,9 +328,7 @@
                 bindingApplied = true;
             }
 
-            if (bindingApplied) {
```
brianmhunt: It looks like `bindingApplied` is no longer used and can be removed from this context.

mbest: Thanks.

empress95:
I am new to KO. I need to know when a component has finished updating DOM as stated in #1475. The issue is stated to be resolved by the descendantsComplete event. There is currently no documentation on how to implement this. Please can I be provided with a JSFiddle example on how the descendantsComplete event works?

I currently get alot of flicker as each value gets rendered one by one on the webpage after applying bindings. I am looking for a way only render after all the values in the array has been computed.

Totati:
> I am new to KO. I need to know when a component has finished updating DOM as stated in #1475. The issue is stated to be resolved by the descendantsComplete event. There is currently no documentation on how to implement this. Please can I be provided with a JSFiddle example on how the descendantsComplete event works?

Hi, @ShallomAkporuku search for those in the repository. They have tests, so you can **[find out](https://github.com/knockout/knockout/blob/2db9f7f79939ed289621de72340ab048362ed76b/spec/defaultBindings/withBehaviors.js#L280)** how to use them.

phil-warner:
Additional example on implementing koDescendantsComplete also shown in [this test](https://github.com/knockout/knockout/blob/2bec689a9a7fcaaed37e6abb9fdec648f86c4f81/spec/components/componentBindingBehaviors.js).

PR #907, opened 2017-11-25
When using 'as' option with 'foreach' don't create child context

mbest:
We added the `as` option to `foreach` in version 2.2.0 (#606), which allows you to give a custom name to the data item. Currently this still creates a child context, with `$data` also pointing the the data item, and requires the need to use `$parent` to access the outer context.

We can provide more flexibility by maintaining the same context when `as` is given. With this change, you would then have to use the name given with `as` to access the array items. Since this is a breaking change, I'm suggesting we include it in the upcoming 3.0.0 version.

### DIFF
--- a/spec/defaultBindings/foreachBehaviors.js
+++ b/spec/defaultBindings/foreachBehaviors.js
@@ -540,7 +540,7 @@ describe('Binding: Foreach', function() {
 
     it('Should be able to give an alias to $data using \"as\", and use it within a nested loop', function() {
         testNode.innerHTML = "<div data-bind='foreach: { data: someItems, as: \"item\" }'>"
-                           +    "<span data-bind='foreach: sub'>"
+                           +    "<span data-bind='foreach: item.sub'>"
                            +        "<span data-bind='text: item.name+\":\"+$data'></span>,"
                            +    "</span>"
                            + "</div>";
@@ -551,7 +551,7 @@ describe('Binding: Foreach', function() {
 
     it('Should be able to set up multiple nested levels of aliases using \"as\"', function() {
         testNode.innerHTML = "<div data-bind='foreach: { data: someItems, as: \"item\" }'>"
-                           +    "<span data-bind='foreach: { data: sub, as: \"subvalue\" }'>"
+                           +    "<span data-bind='foreach: { data: item.sub, as: \"subvalue\" }'>"
                            +        "<span data-bind='text: item.name+\":\"+subvalue'></span>,"
                            +    "</span>"
                            + "</div>";
@@ -600,7 +600,7 @@ describe('Binding: Foreach', function() {
         }
     });
 
-    it('Should provide access to observable array items through $rawData', function() {
+    it('Should provide access to observable items through $rawData', function() {
         testNode.innerHTML = "<div data-bind='foreach: someItems'><input data-bind='value: $rawData'/></div>";
         var x = ko.observable('first'), y = ko.observable('second'), someItems = ko.observableArray([ x, y ]);
         ko.applyBindings({ someItems: someItems }, testNode);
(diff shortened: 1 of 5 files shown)


mbest:
I've attached a change that adds a `noChildContextWithAs` option for this purpose.

mbest:
@brianmhunt I think in TKO, this is already the default. What do you think of this approach for 3.5.0?

brianmhunt:
@mbest  I think this is the better approach conceptually, for 3.5.

I'd argue it's not API breaking because it's more a clarification of an ambiguity, and this feels like the more natural outcome.

mbest:
@brianmhunt I've made the new behavior optional and default to off. That makes it not a breaking change. However, I've already decided to include another "breaking" change to `foreach` (#2324), also optional, but turned on by default. Would you be okay with having this turned on by default for 3.5.0? (considering that the existing behavior is counter intuitive and the documentation already implies the new behavior)

brianmhunt:
👍 @mbest I think #2324 is a good choice, especially if featured prominently in the changelog for 3.5.

mbest:
What about the change here? Should I make the new behavior the default?

brianmhunt:
I think the new behaviour here should be the default too.  The number of people upgrading, who don't read the release notes, who will be affected should be pretty marginal.

fastfasterfastest:
> We can provide more flexibility by maintaining the same context when as is given.

@mbest - what flexibility is provided by _not_ creating a child context?

fastfasterfastest:
> I'd argue it's not API breaking because it's more a clarification of an ambiguity

@brianmhunt - what ambiguity are you referring to?
IMO, http://knockoutjs.com/documentation/foreach-binding.html is clear that
- the `foreach` binding duplicates a section of markup for each entry in an array, and binds each copy of that markup to the corresponding array item
- the `as` option specifies an alias to `$data`

I don't see any ambiguity there - the context of the markup `foreach` is the array item and `as` simply introduces an alias to `$data`.

> The number of people upgrading, who don't read the release notes, who will be affected should be pretty marginal.

I disagree.  I think this change will affect lots of people.  I think most usages of `foreach` with the `as` option will be affected.

As http://knockoutjs.com/documentation/foreach-binding.html states, using the `as` option is especially useful with nested `foreach` bindings to refer to the "outer", or parent, item more easily.
I can imagine many instances of `foreach`, including my own, started out without an `as` option.  And as such properties of the array item were refered to "unqualified", simply by using property names.
Later on, perhaps as the markup got more elaborate or perhaps when a nested `foreach` was needed, it was nice, clearer and more explicit to use the `as` option to refer to the array item.
But, there was no need to go through and change existing markup - the context stayed the same (the array item) and there was no need to change property references and thus it is very likely that there are lots of references that implicitly uses $data.  With this change this will all break.

fastfasterfastest:
> I'd argue it's not API breaking because it's more a clarification of an ambiguity

I disagree.  I think it is definitively a breaking change.  And a bad one.

Per http://knockoutjs.com/documentation/foreach-binding.html:

> The `foreach` binding duplicates a section of markup for each entry in an array, and binds each copy of that markup to the corresponding array item.

With this change, this will no longer be true if the `as` option is used.  The `foreach` binding will still duplicate the section of markup for each entry in an array, but will bind each copy of the markup to the _parent_, not to the array item.  That is a big change and changes how `foreach` works in a drastic way.

What is the rationale to introduce this functionality as a _breaking_ change?
What is the benfit, what does it add?

I urge you to change `ko.options.createChildContextWithAs` default value to `true`.

I also would urge you to introduce another option, perhaps named `createChildContext` (default `true`) or `noChildContext` (default `false`), to allow specifying the behavior on an individual basis, for each usage of `foreach`.  That way users can elect to enroll in this new behavior incrementally.

I would even argue for there is no need for a global `ko.options.createChildContextWithAs` - simply having an `noChildContext` option should be sufficient.  The normal and common way for `foreach` is to create a child context, just as the documentation describes.  Only in rare(?) instances might it be useful to not have a child context created and the `noChildContext` option can then be used.

Having a global `ko.options.createChildContextWithAs` makes it hard to reliably share and reuse templates, especially without an option that allows you to override the global behavior.  A template may have been authored with an expectation of `foreach` to create a child context (as usual).  If a user/caller of that template sets `ko.options.createChildContextWithAs` to another value in her application then that template will now break.

mbest:
I've finally gotten some time to work on this. Please take a look at #2403.

PR #2324, opened 2017-12-03
Improve foreach performance by using arrayChange event

mbest:
Currently, the `foreach` binding has to compare the old and new arrays on each update. Using the `arrayChange` event will provide a performance improvement when using a known operation like `push` with large arrays.

Unfortunately, in order to gain this benefit, we'll need to change the default behavior of `foreach` to consider `includeDestroyed` as `true`.

My question is: How many are using the `destroy`/`destroyAll` methods?

### DIFF
--- a/spec/defaultBindings/foreachBehaviors.js
+++ b/spec/defaultBindings/foreachBehaviors.js
@@ -61,7 +61,7 @@ describe('Binding: Foreach', function() {
 
 
     it('Should add and remove nodes to match changes in the bound array', function() {
-        testNode.innerHTML = "<div data-bind='foreach: someItems'><span data-bind='text: childProp'></span></div>";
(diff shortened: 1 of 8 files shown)


codymullins:
Never used it.

codymullins:
I use the `onDestroy` method for the component but I don't believe I've used it for arrays.

miellaby:
I guess this optimisation won't work with "foreach: { data: array }" syntax, as it will be neutralized by the re-template-everything-when-any-binding-definition-dependency-coughs undesired behavior of  "template/foreach: { someThing: someExpression, ... }" syntax, as in #2332.

mbest:
The syntax doesn't matter with `foreach`. In both cases, this optimization will work (or not, e.g, `foreach: array.sorted()`).

Issue #1990, opened 2016-01-29
CSS Binding does not remove classname on observable update


gooeyideas:
When you are setting CSS Binding's class names with a JavaScript object, you sometimes run into an issue where you also want to include a dynamic class name alongside your static ones. This is simple to do with the css binding using the following example:

JavaScript
````js
function MyViewModel() {
  this.status = ko.observable("incomplete");
  this.isRequired = ko.observable(true);
  this.message = ko.observable("Please Complete the Following Task");
}
````
HTML
````html
<div data-bind="css:{'alert--required': isRequired, ['alert--status-' + status()]:true}" class="alert">
    <span class="alert-message" data-bind="text: message"></span>
</div>
````

The only issue with this method is that the class name set via `['alert--status-' + status()]` does not get removed when the `status` observable changes. Each time the observable changes, it adds a new class to the element.

To work around this I have created a custom extender that clears the class name before it updates:
JavaScript
````js
ko.extenders.css = function(target, value) {
  var beforeChange;
  var onChange;

  //add sub-observables to our observable
  target.show = ko.observable(true);

  beforeChange = function(oldValue){
    target.show(false);
  }
  onChange = function(newValue){
    target.show(true);
  }
  target.subscribe(beforeChange, null, "beforeChange");
  target.subscribe(onChange);
  return target;
};
function MyViewModel() {
  this.status = ko.observable("incomplete").extend({ css: true });
  this.isRequired = ko.observable(true);
  this.message = ko.observable("Please Complete the Following Task");
}
````
HTML
````HTML
<div data-bind="css:{'alert--required': isRequired, ['alert--status-' + status()]:status.show()}" class="alert" >
    <span class="alert-message" data-bind="text: message"></span>
</div>
````
Here is a link to a working CodePen for reference: http://codepen.io/anon/pen/obyYXR

**My question is: Why doesn't the CSS binding clear the class names in this situation? Would it be terribly difficult to make knockout handle this type of functionality without a custom extender?**

mbest:
We are already working on this. See #1710.

gooeyideas:
Will all due respect, this is not a pull request.

The CSS binding lets you set class names with a JavaScript object. JavaScript lets you set dynamic properties via: objectName[varPropName]. 

Therefore the CSS Binding should allow the same functionality however there is a bug when using an observable for the dynamic property name that causes class names to persist after the observable that generated them has changed.

[...]

PR #2327, opened 2017-12-06
1990 remove css class on observable change

setagana:
This pull request is meant to address [Issue 1990](https://github.com/knockout/knockout/issues/1990).

## New functionality for the CSS binding 
With this pull request, the CSS binding will accept an array of strings which forms the definitive list of classes for the bound element. Below is my reasoning for this additional behaviour which can (in time) I believe replace the current use of class objects.

This functionality has been added to existing functionality for the sake of backwards compatibility.

### The CSS binding is inconsistent with other data bindings in the framework.

In its current form, I believe the **CSS** binding to be the only binding that requires both knowledge of the existing state and the necessary delta in order to reach a desired state. If you start with an element like:

`<div data-bind="css: { myFirstClass: true, mySecondClass: true }"></div>`

which generates

`<div class="myFirstClass mySecondClass"></div>`

and you desire to achieve the state:

`<div class="myFirstClass myThirdClass"></div>`

you are required to emit a change which contains the full delta from initial state to desired state:

`<div data-bind="css: { myFirstClass: true, mySecondClass: false, myThirdClass: true }"></div>`

Compare this to the **text** data binding where it is sufficient to emit only the desired state in order to make a change:

`<p data-bind="text: 'This text will overwrite anything which was previously bound'"></p>`

Presumably this is because the **CSS** binding was written to work in tandem with directly applied classes like so:

`<div class="directly-applied-class" data-bind="css: { binding-applied-class: true }"></div>`

However this is itself an inconsistency with other bindings, where directly applied content is overwritten by the data-binding:

`<p data-bind="text: 'Text which will be shown'">Text which will be overwritten</p>`

### Knockout observableArrays are easier to manipulate than objects

Using arrays as the input for the **CSS** binding allows users to leverage the many helpers for array manipulation contained in knockout. For more on this, please refer to the [documentation for working with knockout arrays](http://knockoutjs.com/documentation/observableArrays.html).

### Having the bound array be the definitive list of classes makes working with dynamic classes trivial

As raised in issue 1990 and evident from the [discussion of pull 1710](https://github.com/knockout/knockout/pull/1710), working with dynamic classes in the current version of knockout isn't self explanatory. By making the **CSS** data-binding definitive for the list of classes on an element (as the **text** binding is for its contents) it becomes trivial to work with static and dynamic classes as shown by [this pen](https://codepen.io/setagana/pen/ooRNJq) using the changes proposed in this PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,359 @@
+{
+  "name": "knockout",
+  "version": "3.5.0-pre",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "abbrev": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
+      "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==",
+      "dev": true
+    },
+    "argparse": {
+      "version": "0.1.16",
+      "resolved": "https://registry.npmjs.org/argparse/-/argparse-0.1.16.tgz",
+      "integrity": "sha1-z9AeD7uj1srtBJ+9dY1A9lGW9Xw=",
+      "dev": true,
+      "requires": {
+        "underscore": "1.7.0",
+        "underscore.string": "2.4.0"
+      },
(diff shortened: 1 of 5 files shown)


mbest:
Related: #930

mbest:
You can already use the `class` binding with string value to achieve this functionality (minus removing fixed classes). If it's important to support arrays in addition to strings, that should be added to the `class` binding.

PR #2335, opened 2018-01-10
ko.when returns predicate truthy value as callback 1st parameter

miellaby:
see discussion in https://github.com/knockout/knockout/pull/2113 comments

### DIFF
--- a/spec/observableUtilsBehaviors.js
+++ b/spec/observableUtilsBehaviors.js
@@ -94,5 +94,5 @@ describe('ko.when', function() {
             expect(value).toEqual(2);
         }
         ko.when(ko.observable(2), myCallback);
-    });    
+    });
 });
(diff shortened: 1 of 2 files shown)


mbest:
Thanks.

miellaby:
next improvement: not wrapping an already subscribeable into a computed?

mbest:
> not wrapping an already subscribeable into a computed?

I don't think it's necessary, and it would make the code a bit more complex.

miellaby:
Now, I'm thinking... ko.when should accept a second `owner` argument for consistency with the remaining JS API. It would be passed to the computed.

mbest:
> Is it OK to pass an observable as a computed read function?

Yes,

mbest:
> ko.when should accept a second owner argument

It probably makes sense to use the existing `context` parameter for the computed. I'll work on that.

PR #2342, opened 2018-01-29
Add when binding; use context for predicate in ko.when

mbest:
See https://github.com/knockout/knockout/pull/2335#issuecomment-358889127 and https://github.com/knockout/knockout/pull/2319#issuecomment-361317523


brianmhunt:
I think the `when` should accept the canonical representation of a future, a `Promise` or `thenable`.

The binding becomes (mostly) trivial if observables can be easily convertible to promises with e.g.

```javascript
promise = obs.when(true)
```

with e.g. [`when`] (https://github.com/knockout/tko/blob/4f302f55d075bacad8ece3b440f0803aa95e6574/packages/tko.observable/src/subscribable.js#L126), etc.

mbest:
If we include support for `Promise` in Knockout, we'll need to polyfill it for older browsers. I started working on that (https://github.com/knockout/knockout/issues/1533#issuecomment-325525471) for the component "after-render" but later decided against that approach.

src/subscribables/observableUtils.js
```diff
@@ -1,15 +1,22 @@
...
-    var subscription = observable.subscribe(function(value) {
-        if (value) {
-            subscription.dispose();
-            callback.call(context, value);
-        }
-    });
-    // In case the initial value is true, process it right away
-    observable['notifySubscribers'](observable.peek());
+    function kowhen (resolve) {
+        var observable = ko.pureComputed(predicate, context).extend({notify:'always'});
```
brianmhunt: For performance, would it make sense if this was always a deferred observable?

We may have to change the initial test so it resolves synchronously (unless `ko.when` is to be always async), but otherwise that *feels* like a good internal boost. :)

mbest:
From https://github.com/knockout/knockout/pull/2319#issuecomment-361515887

> With async loading there is also a usecase for reloading the view. 

I've updated the `when` binding to support this. This makes the `when` and `if` bindings almost identical. The only difference is that `when` delays the `descendantsComplete` event until its value is true.

mbest:
@brianmhunt Thanks for all of your advice. I've been thinking about how to use promises in Knockout, but observables generally provide more flexibility. In this case, I've decided to support multiple "reloads" and so a promise would be insufficient.

codymullins:
My personal opinion would be to not avoid promises and perhaps embrace the poly fill, or perhaps only return a promise or add this supported feature if the promise is available.

Could be a modern library or promise-polyfill requirement to enable this feature....

chrisknoll:
I don't think adopting promises isn't the problem, just that promises are useful as an async process completion, not as a callback handler. I believe what @mbest is describing is that the situation he's addressing is the situation when something may change it's value over time, and he needs to respond to each change (like what observable do). If I understand the pattern around promises, something gets set in a sort of 'pending' state, and multiple sources can 'wait' on it and will fire once the promise is resolved (and if the promise was resolved and someone later waits on it, it immediately completes). But once resolved, I don't believe promises go back to 'unset' so that they can re-notify their callers? The value that is resolved by the promise completion is always going to result in the same value (the value that the asynchronous method yielded).  Much different then a pub-sub or callback mechanism.

brianmhunt:
I think users expect a function called `when`  to return a Promise.  Plus, that's what TKO does for its `when` functions, so this could be a point of regression.

We should be cognizant of [TC39](https://github.com/tc39/proposal-observable), which creates filterable, mappable streams - which is what this PR's `when` seems like it might be doing if it can "reload".  A reloadable "when" would seem to be closer to this, and might better be named `listen` e.g. `ko.listen(...)`.

brianmhunt:
... which is really just a concern about making sure we've got the right name to match the behaviour that users expect. 😄

chrisknoll:
I'm beginning to see why the if: and when: bindings are similar.... the if binding is used to conditionally present a template/dom based on the value of an observable....the when binding seems to do the same thing (except for the handling of the descendantsComplete event being delayed in the case of when).

Makes me wonder when I should be using the if: binding vs when: binding.  Usually it's helpful for me to see the usecase to understand the subtle differences.  

@brianmhunt : I get your point about whens and promises:  I refreshed my memory on the JQuery page on when() and it's about accepting deferreds and exposing the promise for it (a single deferred yields it's promise vs. an array of deferred results in a 'wrapper' promise once all deferred are done vs passing a non-deferred yields a resolved promise).  So I think of when() as an API to deferreds (in the context of promises).

For knockout bindings, tho, seems like introducing the notion of when but having in the context of observables is logically consistent.  I think this makes sense to me, and considering the deep connection to observables in knockout, it follows that this would be the focus.  It seems that the main difference between observables and promises: promises go from an unknown state to a resolved state (and keep this resolved state forever) while observables are built to continuously notify changes. 

I use promises all over the place in my own apps, but usually in the context of responding to some user action that results in some asynchronous process (button click -> ajax call -> then( update observable).  I'm having trouble seeing where promises fit in the knockout observable space, but if TKO is using them then I'm interested in seeing what's in store in the future of knockout!  Hopefully we'll see a full fledged 4.0!

codymullins:
Yeah, you're right @brianmhunt. I was confused since Angular has `$q.when(value, [successCallback], [errorCallback], [progressCallback]);`.

fastfasterfastest:
Is there a description or spec of what the new `ko:when` binding would do?

I looked and found https://github.com/knockout/knockout/pull/2113 that describes it as:

> The first parameter is an observable (or a function that will be wrapped in a computed), and the second is a function that's run once the former becomes truthy.
...
So it'd only fire once and never again
...
Yes, once only.

However, above @mbest states:

> I've decided to support multiple "reloads"

I haven't really studied how the new `ko:when` binding behaves, but am a little concerned about the similarity with the `ko:if` binding.
If the `ko:when` binding can "fire" multiple times, then what is the difference between `ko:if` and `ko:when`?  I think it would be very unfortunate to have two bindings that are very similar with only some minor, esoteric differences.  Perhaps it is better to stick with `ko:if` and add some parameter(s) to it to get the behavior of the new `ko:when`.
But again, I am voicing my concern with limited knowledge and don't really know what the differences are - perhaps the differences are major, perhaps not?

fastfasterfastest:
This is how http://knockoutjs.com/documentation/if-binding.html describes the `if` binding:
> The `if` binding causes a section of markup to appear in your document (and to have its data-bind attributes applied), only if a specified expression evaluates to true (or a true-ish value such as a non-null object or nonempty string).

@mbest: How is the `when` binding described?  What is the difference between the `if` and `when` bindings?

mbest:
All of the above applies to the `when` binding as well. The difference is that the `when` binding considers a false value to be temporary and delays an ancestor's `descendantsComplete` event until the value becomes true. Considering this subtle difference, it may be better represented as an option to the `if` binding. What do you guys think?

chrisknoll:
well, one one hand, maybe it's redundant.  On the other hand, I'm not sure how if + options would look?   If you go with the if binding with options, does it look something like:

```
<div data-bind="if: condObs, ifOptions:{some params}"/>

```
If so, i'd leave the when binding, alter the IF documentation to be 'If and When' and just describe the subtle difference between the two.

fastfasterfastest:
I think it would be better with one `if` binding, as opposed to an `if` binding and a `when` binding.  And allow the `if` binding to take either an expression or an object literal with `condition` and `delayDescendantsComplete` properties.

`<div data-bind="if: condObs" />`
`<div data-bind="if: { condition: condObs, delayDescendantsComplete: true }" />`

(I may also have some questions about how the `descendantsComplete` event works but that will be for another issue/pull request thread.)

brianmhunt:
I agree with @fastfasterfastest that a single binding makes sense - perhaps a better name might just be `wait` so e.g. `if: { condition: x, wait: true }`.  It could also be exposed on `allBindings` as e.g. `if: x, wait`.  If we want to be verbose, perhaps `waitForChildren`.

Somewhat aside, I think a `when` binding feels like something different.  Given `when: x` I expect KO to 1. unwrap; 2. wait for any thenables/promises  i.e. (with ES7-asyncs) e.g.

```javascript
ko.bindingHandlers.when = {
   init (node, valueAccessor) {
      ko.when(valueAccessor()).then(() => ko.applyBindingsToDescendants(...))
      return { controlsDescendantBindings: true }
   }
}
```

Where `ko.when` will unwrap/resolve e.g.

```javascript
ko.when = (predicate) => {
   return Promise.resolve(ko.unwrap(predicate)).then(v => ko.unwrap(v))
}
```

As all of the following should resolve to `x`

- `ko.when('x')`
- `ko.when(ko.observable('x'))`
- `ko.when(Promise.resolve('x'))`
- `ko.when(Promise.resolve(ko.observable('x')))`
- `ko.when(ko.observable(Promise.resolve('x')))`
- `ko.when(Promise.resolve(ko.observable(Promise.resolve('x'))))`

(And equivalently for thenables in place of Promises).

At least, that sort of strikes my expectations on the head when I see `when`.

chrisknoll:
@brianmhunt : it sounds like you're describing knockout extending a 'when' capability that handles any resolvable that the default Promise framework supports (literals, thenables) plus handling the case when an observable wrapping a literal/thenable is passed in (by doing a ko.unwrap on the target of the binding).

What I do not understand is your example implementation of ko.when:  Isn't that stuff supposed to be context-specific?  Meaning: you're saying if you do ko.when, it has a resolve handler to unwrap the value.  Is that to set up dependcies on the value of the observable (if it's an actual observable)?  I just need to see how it would be used in practice:

Model:
```
function myViewModel() {
  var self = this;
  self.currentPromise = ko.observable();

  self.searchClick = searchText => { self.currentPromise(ajax.get(...url, params, etc})};
  
}
```

View:
```
<div data-bind="when: currentPromise">
  <searchResults params="results: $data {?}"> // $data stores the resolved value?
</div>
```

The View just wraps a fictional "searchResults" component that renders the given results. What I'm missing is how to get to the value of the promise stored in the currentPromise observable...does the value get set into the $data context of the binding context?

In my own applications, when I need something that works off promises, I don't need to rely on anything inside the core of knockout to handle this for me, I just do this:

Model:
```
  var self = this;
  self.results= ko.observable();

  self.searchClick = searchText => { 
    // I could cancel a pending ajax call here if necessary
    ajax.get(...url, params).then(results => self.results(results)
  };
```
View
```
<div data-bind="if: results">
  <searchResults params="results: results">
</div>
```

So I think the thing you're adding is a knockout enhancement to handle when() in the case that the target of the when() is an observable wrapping a resolvable.  So a ko.when() makes sense when you do a ko.when(obs).then(val=>{ ...work with resolved result ...}.  I also don't quite get the default resolve handler then(v=> unwrap(v)) unless that's just to set up a dependency on v in the event the promise resolves to an observable?  I don't get the when: databind tho.

-Chris

mbest:
> Then i'd say drop the When, document the descendantsComplete binding as a callback handler, and have it apply to anything (like components, templates, etc) that has descendant bindings.

@chrisknoll I really like this approach. I'll work on an update to do that.

fastfasterfastest:
>> Then i'd say drop the When, document the descendantsComplete binding as a callback handler, and have it apply to anything (like components, templates, etc) that has descendant bindings.
>
>@chrisknoll I really like this approach. I'll work on an update to do that.

I think I agree.
"callback handler" - does that mean for notification of the `descendantsComplete` event only?  Or does that mean for "controlling" the firing of the `descendantsComplete` event?

The idea with the (maybe-scrapped) `when` binding was to be able to "control" **when** the `descendantsComplete` event was percolated upwards.  I am not sure if that is really needed - perhaps things can/should be handled in the view model instead - but think a use case would be helpful here.

mbest:
See #2364.

> does that mean for notification of the descendantsComplete event only? Or does that mean for "controlling" the firing of the descendantsComplete event?

It will do both.

Issue #2353, opened 2018-02-20
Include Typescript types inside repository


SomaticIT:
Hi wonderful knockout team !

I wrote precise TypeScript types definitions for knockout.
Definitions can be found here: https://github.com/typed-contrib/knockout/blob/master/global/index.d.ts

This project was written for the `typings` which is shutting down leaving me with two solutions:
1. Merge my definitions with `DefinitelyTyped` types.
2. Merge my definitions directly inside the source project.

I prefer the second solution because it avoids having multiple dependencies in NPM.
Indeed, TypeScript compiler automatically finds types which are included in NPM packages.

Do you agree to include these types?
If you do, I can open a PR to make the appropriate configuration for you.

Thank you very much for this library which I really make huge use!
You can find most of my knockout related open-source projects [here](https://github.com/spatools?q=ko)

brianmhunt:
@SomaticIT  — Great work, thanks for posting this.

There's a related issue on [tko](/knockout/tko): knockout/tko#6 — which is already transpiled with Typescript.

@mbest Is there any intention of / ability to include Typescript definitions in ko 3.x?

SomaticIT:
Thank you for your quick response!
I think I can help if there is any TypeScript migration project for `tko`.
@mbest Let me know if you want me to open a PR to apply the config and merge my types.

mbest:
Thanks for sharing your work on this! I'd like to get this into Knockout as well.

SomaticIT:
Hi @mbest,
I just forked the repo and start working on some improvements to match new typescript capabilities.
I will open a PR soon.
Thanks

brianmhunt:
@SomaticIT Thanks.  What's a Typescript migration project? :)

IanYates:
As a big Typescript and KO user who long ago took the DefinitelyTyped definitions and updated them (probably around the time components landed) I'd be keen to see your PR and put in our own improvements & tweaks too.  I'll keep watch :)

SomaticIT:
I just submitted a PR with an improved version of typed-contrib types.

mbest:
I've made some updates to the code from @SomaticIT. See 70b7d7c. Please look over them and let me know if you find any problems. I'm still pretty new to TypeScript, so I may have missed something.

Sebazzz:
I left some notes 👍

mbest:
@Sebazzz No release yet. That should be coming soon. Thanks for your notes. I'll make some changes based on them.

[...]

PR #2354, opened 2018-02-22
Add typescript types

SomaticIT:
This Pull Request is created to include `TypeScript` types into `knockout` repository. This is related to #2353. 

This allows the `TypeScript` compiler to automatically find the `knockout` types when using option `moduleResolution: "node"`.

This includes:
 * Types definitions taken from [typed-contrib](https://github.com/typed-contrib/knockout) and revised for `TypeScript` 2+
 * Types tests taken from [typed-contrib](https://github.com/typed-contrib/knockout)
* `Grunt` task to test types automatically
* `package.json` configuration to help `tsc` finding the types.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -2,3 +2,4 @@
 *
 !build/output/knockout-latest.js
 !build/output/knockout-latest.debug.js
+!build/types/knockout.d.ts
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,1 +1,5 @@
 language: node_js
+
+before_install:
+  - npm install -g typescript
+  
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -34,6 +34,10 @@ module.exports = function(grunt) {
         test: {
             phantomjs: 'spec/runner.phantom.js',
             node: 'spec/runner.node.js'
+        },
+        testtypes: {
+            global: "spec/types/global",
+            module: "spec/types/module"
         }
     });
 
@@ -151,6 +155,24 @@ module.exports = function(grunt) {
         );
     });
 
+    grunt.registerMultiTask('testtypes', 'Run types tests', function () {
+        var done = this.async(),
+            target = this.target;
+
+        grunt.util.spawn({ cmd: "tsc", args: ["-p", this.data] },
+            function (error, result, code) {
+                grunt.log.writeln(result.stdout);
+
+                if (error)
+                    grunt.log.error(result.stderr);
+                else
+                    grunt.log.ok("Knockout TypeScript " + target + " types validated!");
+
+                done(!error);
+            }
+        );
+    });
+
     grunt.registerTask('dist', function() {
         var version = grunt.config('pkg.version'),
             buildConfig = grunt.config('build'),
@@ -170,5 +192,5 @@ module.exports = function(grunt) {
     });
 
     // Default task.
-    grunt.registerTask('default', ['clean', 'checktrailingspaces', 'build', 'test']);
+    grunt.registerTask('default', ['clean', 'checktrailingspaces', 'build', 'test', 'testtypes']);
 };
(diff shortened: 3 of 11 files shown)


IanYates:
This looks really good! :)

Am I right in understanding that we can add tests to the test-module.ts file to verify things that should be able to compile?

Is there a way to test things that should *not* be able to compile?

SomaticIT:
For now, you need to include your tests in both `test-module.ts` and `test-global.ts`.
I think now that TypeScript support `UMD` module definition, we could only keep `test-module.ts`.
I have to search more on this.

Testing things which should not compile is a really good question! :) 
I do not have a solution right now.

I think we can add a folder `failing` which contains a file containing __X__ typing errors.
We then add a task in Gruntfile which executes `tsc` and count compiler errors.

But it may not be a good solution because it invoke a lot of stuff for few tests and the number of errors could vary depending on the TypeScript compiler.

Let me know if you have another idea.

IanYates:
I don't have another idea yet unfortunately.  I enjoy trying to extend TS definitions to avoid things like any, etc as much as possible.  After a quick browse through what you've posted I think yours as pretty good.

My hope is that, when I get some time, I can try them in our project and see what fails.  That'll lead to a PR to amend them in the future, but that's weeks away.

I've also volunteered to do a task similar to yours for another JS library (fabricJS) which has had a few people try various ways of making definitions in definitely typed.  I've been improving an old copy of those definitions as I've been using the library on and off over a few months.  Unfortunately someone then replaced them completely in definitely typed with something else that was completely incompatible but not any better.  So the project maintainer wants someone to do something official :)

A big part of TS is catching things that might look OK but really are incorrect (as you know already.. pointing this out for the spectators).  That's why I was keen to find out if there was a good method for deliberately failing to compile.  When I get to that project in a week or so I'll see if I find a way of testing for that and let you know.

mbest:
I've gone through this and made some updates. See 70b7d7c9bfd2b9a1201. Please look over them and let me know if you find any problems. I'm still pretty new to TypeScript, so I may have missed something.

michaelaird:
I think having the type definitions in the knockout repository is great. I would suggest reaching out to the people who have contributed to the DefinitelyTyped definitions (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/knockout/index.d.ts) for input.

Also, if the typescript definitions are going to be maintained as part of the repo, they should probably be removed from DefinitelyTyped and replace this package: https://www.npmjs.com/package/@types/knockout

mbest:
Thanks for the input @michaelaird. Would it possible for you to try out the new type definitions?

Also any of the contributors from DefinatelyTyped: @borisyankov, @Igorbek, @moonpyk, @EnableSoftware, @BenjaminEckardt, @ffMathy.

codymullins:
I'll add it to our project and see what happens.

SomaticIT:
I'll review your updates and add it to our projects to see if they match.

codymullins:
I guess I didn't really pay much attention, but this is most definitely not an easy update. We're using the type definitions from DefinitelyTyped, which are significantly different. I suspect there are going to be a lot of people in this same situation who are going to want to know what to change to match, unless we can determine what can remain the same.

mbest:
It looks like the DefinitelyTyped declaration file declares all of the interfaces globally, which is not the recommended approach.

michaelaird:
The one thing I beg of you is please do not replicate this bit of blasphemy

```
interface KnockoutExtensionFunctions {
    [key: string]: any;
}
```

It got added to the DefinitelyType definitions awhile back and has accounted for untold # of errors and bugs. It basically says that any identifier you tack onto an Observable or ComputedObservable is valid and ```any```. This means that any kind of typo or mistake passes the TypeScript compiler and only blows up at runtime. :( :( :(

SomaticIT:
@codymullins 

I worked on new types for `knockout.mapping`: https://github.com/typed-contrib/knockout.mapping
I also have a local version for `knockout.validation`.

I wait this types to be fully validated before submitting the dependent types.

mbest:
I've add #2362 with my updates.

mbest:
Replaced by #2362

PR #2358, opened 2018-03-04
Event binding improvements

cervengoc:
This PR contains several minor improvements related to the `event` and `submit` bindings.

- Refactored the `submit` binding to reuse the `event` binding. This way some concepts can be easily shared, like preventing default behavior or stopping bubbling.
- Added a check in the bound event handler to throw an exception in case of the bound value is not a function. This adds some level of additional usability and makes debugging strange issues easier.
- Added support for stopping event propagation from the handler by returning an explicit `false` value. This behavior is similar to how `jQuery` handles return values from event handlers and it adds support for run-time event propagation handling (in contrast to the existing "design-time" `*Bubble` feature).

The latter point introduces somewhat a breaking change, but I don't think that many code is affected as returning `false` explicitly to prevent default behavior *and* stop propagation is an old convention in my opinion.

Please let me know if we should refine anything here.

### DIFF
--- a/src/binding/defaultBindings/event.js
+++ b/src/binding/defaultBindings/event.js
@@ -21,6 +21,10 @@ ko.bindingHandlers['event'] = {
                 ko.utils.registerEventHandler(element, eventName, function (event) {
                     var handlerReturnValue;
                     var handlerFunction = valueAccessor()[eventName];
+                    if (handlerFunction !== null && handlerFunction !== undefined && typeof handlerFunction !== "function") {
+                        throw new Error("The value for a '" + eventName + "' event binding must be a function");
+                    }
+
                     if (!handlerFunction)
                         return;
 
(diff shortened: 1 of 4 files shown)


cervengoc:
For some reasons phantomjs tests seem to fail, I'll have a deeper look at it in the days.

cervengoc:
@mbest I've double-checked several times the code and cannot see why it doesn't work in phantomjs tests. Could you please have a look at it?

src/binding/defaultBindings/event.js
```diff
@@ -21,6 +21,10 @@ ko.bindingHandlers['event'] = {
                 ko.utils.registerEventHandler(element, eventName, function (event) {
                     var handlerReturnValue;
                     var handlerFunction = valueAccessor()[eventName];
+                    if (handlerFunction !== null && handlerFunction !== undefined && typeof handlerFunction !== "function") {
```
codymullins: you could just do `typeof handlerFunction !== function`, the null and undefined check seems redundant in my opinion, shouldn't those be covered by the `!== function` check?

cervengoc: No, in this case this stricter condition was intentional. I wanted to *allow* `null` or `undefined` values as those can be kind of valid values in some cases.

codymullins: Ah, good point. I was under the assumption we didn't want `null` or `undefined` here. Thanks for the explanation.

cervengoc:
@mbest Would you please share your concerns or thoughts on this one if you have any?

PR #2359, opened 2018-03-09
Make jQuery detection more robust

cervengoc:
A widely applied approach with `webpack` is to use the `provide` plugin for such libraries like `jQuery`. When using that plugin, 99% of developers will ask the plugin to expose `jQuery` in a `jQuery` variable for every imported module in an inner scope (but out from the module itself).

With this enhancement we could support this case out of the box without needing an even more special webpack configuration.


mbest:
Thanks

PR #2360, opened 2018-03-13
Add number input workaround for Microsoft Edge.

Deathicator:
Microsoft Edge doesn't fire 'input' or 'change' events for number inputs when the value is changed via the up / down arrow keys.

https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/14678823/

It has been fixed in an upcoming build, but hasn't yet been released.


mbest:
Thanks

PR #2364, opened 2018-03-21
Remove when binding; instead, the presence of descendantsComplete bin…

### LINKED ISSUE #2343: ko.applyBindingsToNode component binding throws error
In beta version 3.5 line 4333 throws error, when using component binding with ko.applyBindingsToNode

```
var element= document.createElement("div");
                ko.applyBindingsToNode(element { component: { name: "example", params:{} } });
```

childBindingContext = asyncContext.createChildContext(componentViewModel, /* dataItemAlias */ undefined, function(ctx) {

asyncContext is undefined

Problem originates from line 3450 

```
var bindingInfo = ko.utils.domData.get(node, boundElementDomDataKey);

where ko.utils.domData.get(node, boundElementDomDataKey); returns undefined
```

Error code from console:

```
knockout-3.5.0beta.debug.js?cache=53027963:4333 Uncaught TypeError: Cannot read property 'createChildContext' of undefined
    at ko.subscription._callback (knockout-3.5.0beta.debug.js?cache=53027963:4333)
    at ko.subscribable.notifySubscribers (knockout-3.5.0beta.debug.js?cache=53027963:1412)
    at knockout-3.5.0beta.debug.js?cache=53027963:3843
    at knockout-3.5.0beta.debug.js?cache=53027963:3861
    at knockout-3.5.0beta.debug.js?cache=53027963:3890
    at init.js?cache=53027963:35
    at tryIssueCallback (knockout-3.5.0beta.debug.js?cache=53027963:4000)
    at knockout-3.5.0beta.debug.js?cache=53027963:4021
    at knockout-3.5.0beta.debug.js?cache=53027963:3890
    at resolveViewModel (knockout-3.5.0beta.debug.js?cache=53027963:4071)
```

mbest:
…ding modifies how the if binding works

descendantsComplete binding can now be used on any node
Add a note that component contents might be updated while the component itself is being updated
Fix #2343 - support applyBindingsToNode for components

See https://github.com/knockout/knockout/pull/2342#issuecomment-374025989


mbest:
Some other internal changes:

1. `BindingContext` uses a pure computed to update itself and therefore doesn't need to keep track of nodes.
2. When using deferred updates, if a pure computed goes to sleep after a notification is scheduled, don't re-evaluate the computed (because there's actually nobody to notify).

PR #2365, opened 2018-03-26
Fixes simpleHttpParse logic for table-like names

avcs06:
The logic fails to parse table-like tags properly, if there are comments or ko blocks at the beginning of template


mbest:
Thanks.

PR #2369, opened 2018-04-14
Updates to Knockout TypeScript definitions

mbest:
Based on @Sebazzz's comments.

Related #2353

### DIFF
--- a/build/types/knockout.d.ts
+++ b/build/types/knockout.d.ts
@@ -131,8 +131,8 @@ export function isObservableArray<T = any>(instance: any): instance is Observabl
 
 //#region subscribables/dependendObservable.js
 
-export type ComputedReadFunction<T = any> = Subscribable<T> | Observable<T> | Computed<T> | (() => T);
-export type ComputedWriteFunction<T = any> = (val: T) => void;
+export type ComputedReadFunction<T = any, TTarget = void> = Subscribable<T> | Observable<T> | Computed<T> | ((this: TTarget) => T);
+export type ComputedWriteFunction<T = any, TTarget = void> = (this: TTarget, val: T) => void;
 export type MaybeComputed<T = any> = T | Computed<T>;
 
 export interface ComputedFunctions<T = any> extends Subscribable<T> {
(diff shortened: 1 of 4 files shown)


Sebazzz:
Additional thing I noticed:

In case where the `evaluatorTarget` argument is provided, the `this` parameter of the `evaluator` is bound to `evaluatorTarget`. In that case these typings are not correct:

```
 export function computed<T = any>(evaluator: ComputedReadFunction<T>, evaluatorTarget: any): Computed<T>;
 export function computed<T = any>(evaluator: ComputedReadFunction<T>, evaluatorTarget: any, options: ComputedOptions<T>): Computed<T>;
 export function pureComputed<T = any>(options: ComputedOptions<T>): PureComputed<T>;
 export function pureComputed<T = any>(evaluator: ComputedReadFunction<T>, evaluatorTarget: any): PureComputed<T>;
```

You can change:
```
export type ComputedReadFunction<T = any, TTarget=void> = Subscribable<T> | Observable<T> | Computed<T> | ((this: TTarget) => T);
export type ComputedWriteFunction<T = any, TTarget=void> = (this: TTarget, val: T) => void;
```

And then: 
```
 export function computed<T = any, TTarget=any>(evaluator: ComputedReadFunction<T,TTarget>, evaluatorTarget: TTarget): Computed<T>;
 export function computed<T = any, TTarget=any>(evaluator: ComputedReadFunction<T,TTarget>, evaluatorTarget: TTarget, options: ComputedOptions<T>): Computed<T>;
 export function pureComputed<T = any, TTarget=any>(evaluator: ComputedReadFunction<T,TTarget>, evaluatorTarget: TTarget): PureComputed<T>;
```

You could also consider to document this limitation and keep the typings simpler as they are, since I expect that usually in Typescript a fat arrow function will be used as first parameter and the overloads with `evaluatorTarget` are unused.

mbest:
@Sebazzz  Thanks for the notes. I've made those updates.

SomaticIT:
I did a little review, and did not see anything missing.
On the contrary, I think you did great job!

I will make some more tests, but it seems good.

PR #2378, opened 2018-05-09
Reverse some of the ko.utils.array* changes from #1448 and #1200.

mbest:


fastfasterfastest:
I think it would be a good idea to also piggyback on the native implementations of `map` and `filter` for performance reasons.

And to pass along the array itself to the callback function of the `arrayMap` and `arrayFilter` methods (see [note] below) - this will align these methods what the native implementations do.

These additional changes would/should be entirely backward compatible with 3.4.2.

```
arrayMap: Array.prototype.map ?
   function (array, mapping, mappingOwner) {
      return array ? Array.prototype.map.call(array, mapping, mappingOwner) : [];
   } :
   function (array, mapping, mappingOwner) {
      var result = [];
      if (array) {
         for (var i = 0, j = array.length; i < j; i++)
            result.push(mapping.call(mappingOwner, array[i], i, array));  //[note]
      }
      return result;
   },

arrayFilter: Array.prototype.filter ?
   function (array, predicate, predicateOwner) {
      return array ? Array.prototype.filter.call(array, predicate, predicateOwner) : [];
   } :
   function (array, predicate, predicateOwner) {
      var result = [];
      if (array) {
         for (var i = 0, j = array.length; i < j; i++)
            if (predicate.call(predicateOwner, array[i], i, array)) //[note]
               result.push(array[i]);
      }
      return result;
   },
```

mbest:
Give that I want to release 3.5 soon, I'd like to keep this limited to reverting changes since 3.4 rather than adding anything new.

chrisknoll:
That's cool, something to look forward to in 3.6! ;) ;)

Issue #231, opened 2011-12-15
Use jQuery when available in style binding


wallymathieu:
We had a problem in our project with setting background color in IE and firefox. I've tried to make a simple example out of an existing fiddle:
http://jsfiddle.net/ozzymcduff/GSvnh/30/
The fiddle has problem rendering in firefox, but not IE9 and chrome. This is probably a cross browser thing.  

In knockout we have the following in the style bindinghandler
element.style[styleName] = styleValue || "";
We made a separate style bindinghandler and changed the above to 
$(element).css(styleName, styleValue || "");

jrsearles:
Another useful thing jQuery does is check for a vendor prefixed version if the style property does not exist. I ran into this today using the default Knockout style binding.

I agree that the behavior should be consistent when not using jQuery though, so my preference would be for that behavior to be pulled into Knockout.

mbest:
See #2302

derrickb:
Would it be feasible to make this optional?

Using 3.5.0rc, I'm finding that I now have to wrap my url()'s in escaped quotes which, while not the end of the world, is cluttering up my html quite a lot and making VS Code's syntax highlighting unhappy as a side effect.

I've been using the style binding like so:
`style: { backgroundImage: 'url('+getAsset('select_arrow')+')' }`
Where 'getAsset' returns a base64 string.
With the new functionality, I have to use:
`style: { backgroundImage:'url(\"'+getAsset('select_arrow')+'\")' }`

Again, not a severe issue, but I use this binding a LOT in my UI so it's unfortunate to have to go through it all to add the quotes.

chrisknoll:
have you tried tempalted strings (using a ` to wrap your string)?

derrickb:
Do you mean in the binding in the html file? No, because although I see that ko will support parsing these, my app runs in very old browser versions so everything needs to be transpiled, and AFAIK, there isn't a way to preprocess templated strings within data-bind's. Would love to be wrong on that.

Using template strings in the getAsset function hasn't made a difference either. I always have to wrap the result in escaped quotes in the html.

fastfasterfastest:
@derrickb  I think you may have something else going on, or haven't told the whole story.  Are you using jquery, and if so, which version?

Here are two fiddles that uses the `style` binding and sets the background image using a data url
`<div data-bind="style: { backgroundImage: 'url('+getDataUrl(backgroundColor())+')' }"></div>`
and they seem to "work"

[...]

PR #2380, opened 2018-05-11
fix ko.contextFor after applyBindingAccessorsToNode, etc.

mbest:
Make sure you can access a node's context and data when a node has bindings only from `ko.applyBindingsToNode` or `ko.applyBindingAccessorsToNode`. Related: #2148.

From https://github.com/knockout/knockout/issues/231#issuecomment-388210267:

> One bad issue I'm dealing with right now is contextFor() and dataFor() not working on a dynamically added element (that then had bindings applied via applyBindingAccessorsToNode). Both return undefined. If I call either on the parentElement, they return the data or context as expected.


derrickb:
Happy to report that this fixed the issue I described here:
https://github.com/knockout/knockout/issues/231#issuecomment-388210267

Thank you!

PR #2390, opened 2018-06-05
ko.applyBindings requires a valid 2nd parameter if given

### LINKED ISSUE #2388: ko.applyBindings - Second Parameter
**Version**: 3.5 RC & 3.4.2

**Example**: [JSFiddle](https://jsfiddle.net/ypsjdc66/)

Via the documentation the second parameter should _restricts the activation to the element with ID someElementId and its descendants_... but when I pass in an element that doesn't exist, the applyBindings function is ignoring it. Shouldn't it fail if the element doesn't exist instead of searching the entire document?

mbest:
Fixes #2388

### DIFF
--- a/src/binding/bindingAttributeSyntax.js
+++ b/src/binding/bindingAttributeSyntax.js
@@ -529,7 +529,7 @@
     ko.applyBindingAccessorsToNode = function (node, bindings, viewModelOrBindingContext) {
         if (node.nodeType === 1) // If it's an element, workaround IE <= 8 HTML parsing weirdness
             ko.virtualElements.normaliseVirtualElementDomStructure(node);
-        return applyBindingsToNodeInternal(node, bindings, getBindingContext(viewModelOrBindingContext), true);
+        return applyBindingsToNodeInternal(node, bindings, getBindingContext(viewModelOrBindingContext));
     };
 
     ko.applyBindingsToNode = function (node, bindings, viewModelOrBindingContext) {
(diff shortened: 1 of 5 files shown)


src/binding/bindingAttributeSyntax.js
```diff
@@ -539,26 +539,26 @@
...
+    ko.applyBindings = function (viewModelOrBindingContext, rootNodeArg, extendContextCallback) {
         // If jQuery is loaded after Knockout, we won't initially have access to it. So save it here.
         if (!jQueryInstance && window['jQuery']) {
             jQueryInstance = window['jQuery'];
         }
 
-        // rootNode is optional
-        if (!rootNode) {
+        var rootNode = rootNodeArg;
+        if (!rootNode && arguments.length < 2) {
```
fastfasterfastest: change to `if (arguments.length < 2) {`

src/binding/bindingAttributeSyntax.js
```diff
@@ -539,26 +539,26 @@
...
-    ko.applyBindings = function (viewModelOrBindingContext, rootNode, extendContextCallback) {
+    ko.applyBindings = function (viewModelOrBindingContext, rootNodeArg, extendContextCallback) {
         // If jQuery is loaded after Knockout, we won't initially have access to it. So save it here.
         if (!jQueryInstance && window['jQuery']) {
             jQueryInstance = window['jQuery'];
         }
 
-        // rootNode is optional
-        if (!rootNode) {
+        var rootNode = rootNodeArg;
```
fastfasterfastest: Why introduce another variable and not simply use `rootNodeArg`?

PR #2398, opened 2018-07-07
Fix descendantComplete for if, ifnot, with

mbest:
See knockout/tko#65 and #2342

Add `completeOn: "render"` binding option for `if`, `ifnot`, and `with` that makes them treat their un-rendered state as incomplete, instead of using the existence of a `descendantsComplete` callback for this. Also make sure the event works for multiple renders.


mbest:
@brianmhunt What do you think?

brianmhunt:
It looks right; I'll report back any issues when merge it into TKO.

fastfasterfastest:
I have looked for a corresponding issue where `descendantComplete` is described, discussed and the need for it expressed, but cannot find one.  Is there such an issue?
What is a use case for `descendantComplete`?

Issue #2400, opened 2018-07-16
Cannot bind to custom CSS property variables


zero298:
It seems that you are unable to bind to [custom CSS property variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables) using [style bindings](http://knockoutjs.com/documentation/style-binding.html).  I saw this issue on the Stack Overflow question: [Using CSS3 Variables in Knockout Style Binding](https://stackoverflow.com/q/51368941/691711).

The issue seems to stem from [knockout/src/binding/defaultBindings/style.js](https://github.com/knockout/knockout/blob/9eb61061e45e371362aee8d182392c531b7ac2ef/src/binding/defaultBindings/style.js#L15-L24).

Specificially how the style property is set using Array syntax instead of [`CSSStyleDeclaration#setProperty`](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/setProperty).

The custom properties don't seem to be enumerated on the actual Object and instead have to be set through `setProperty`.  I'm not sure if the appropriate action would be to have everything use `setProperty` or if you should check the `styleName` for `"--"`, say it is a variable, and only use `setProperty` for that particular instance.

brianmhunt:
I wrote and expect to include a `style-map` binding for tko that accomplishes this.  Css vars are amazing and it’s be a shame not to support them.  The ordinary binding could be altered to support them too, by -as mentioned- observing if the first two chars are ‘—‘

PR #2401, opened 2018-07-16
Forcing use of setProperty for CSS variables

zero298:
I made a really narrow commit to address #2400.  It seems as though Custom CSS properties can only be set through [`CSSStyleDeclaration#setProperty`](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/setProperty).

I'm not sure why you use the Array syntax for all properties instead of having everything go through `setProperty`, but I wanted to err on the side of caution and only use it when we are using CSS variables.  If you'd prefer this commit to be broader or if you would like some sort of coding style change, let me know.

### DIFF
--- a/src/binding/defaultBindings/style.js
+++ b/src/binding/defaultBindings/style.js
@@ -12,15 +12,30 @@ ko.bindingHandlers['style'] = {
             if (jQueryInstance) {
                 jQueryInstance(element)['css'](styleName, styleValue);
             } else {
-                styleName = styleName.replace(/-(\w)/g, function (all, letter) {
-                    return letter.toUpperCase();
-                });
+                var previousStyle;
+                var postStyle;
+                var requiresGetPropertyValue = (styleName.substring(0, 2) === "--");
 
-                var previousStyle = element.style[styleName];
-                element.style[styleName] = styleValue;
+                if(requiresGetPropertyValue){
+                    previousStyle = element.style.getPropertyValue(styleName);
+                    element.style.setProperty(styleName, styleValue);
+                    postStyle = element.style.getPropertyValue(styleName);
 
-                if (styleValue !== previousStyle && element.style[styleName] == previousStyle && !isNaN(styleValue)) {
-                    element.style[styleName] = styleValue + "px";
+                    if (styleValue !== previousStyle && postStyle == previousStyle && !isNaN(styleValue)) {
+                        element.style.setProperty(styleName, styleValue + "px");
+                    }
+                } else {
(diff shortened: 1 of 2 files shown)


src/binding/defaultBindings/style.js
```diff
@@ -12,15 +12,30 @@ ko.bindingHandlers['style'] = {
...
-                var previousStyle = element.style[styleName];
-                element.style[styleName] = styleValue;
+                if(requiresGetPropertyValue){
+                    previousStyle = element.style.getPropertyValue(styleName);
+                    element.style.setProperty(styleName, styleValue);
+                    postStyle = element.style.getPropertyValue(styleName);
 
-                if (styleValue !== previousStyle && element.style[styleName] == previousStyle && !isNaN(styleValue)) {
-                    element.style[styleName] = styleValue + "px";
+                    if (styleValue !== previousStyle && postStyle == previousStyle && !isNaN(styleValue)) {
```
mbest: This code isn't necessary for CSS variables since they can take any value. In other words, the value that's set won't ever be rejected.

src/binding/defaultBindings/style.js
```diff
@@ -12,15 +12,30 @@ ko.bindingHandlers['style'] = {
             if (jQueryInstance) {
                 jQueryInstance(element)['css'](styleName, styleValue);
             } else {
-                styleName = styleName.replace(/-(\w)/g, function (all, letter) {
-                    return letter.toUpperCase();
-                });
+                var previousStyle;
+                var postStyle;
+                var requiresGetPropertyValue = (styleName.substring(0, 2) === "--");
```
fastfasterfastest: Perhaps instead `var isCustomPropertyName = /^--/.test(styleName);`
And possibly declaring the regexp in some outer scope to allow reuse.

src/binding/defaultBindings/style.js
```diff
@@ -12,6 +12,12 @@ ko.bindingHandlers['style'] = {
             if (jQueryInstance) {
                 jQueryInstance(element)['css'](styleName, styleValue);
             } else {
+                // Is styleName a custom CSS property?
```
mbest: Can you reformat this as `else if (..) {` and then add a new `else` block below?

mbest:
This is looking good, but we should have a test for it as well. Can you add a test to `styleBehaviors.js`?

mbest:
You should be able to enclose the unit test in an `if` block to run it only in browsers that  support custom CSS properties. I found this example:

```javascript
if (window.CSS && window.CSS.supports && window.CSS.supports('--a', 0)) {
    ...
}
```

PR #2410, opened 2018-08-28
Removing the observable unwrapping in ko.toJS function

omar01:
The utility function ko.toJSON unwraps each observable in order to get their value. If an observable is unwrapped elsewhere in other computeds, they will be re-evaluated after calling ko.toJSON. 
 
This behavior is causing us issues and in order to prevent this, mapJsObjectGraph should peek the value of the observable which removes that dependency.

Moreover, we thought of overriding the current mapJsObjectGraph callback but it is not possible while everything is private except ko.toJS and ko.toJSON

### DIFF
--- a/src/subscribables/mappingHelpers.js
+++ b/src/subscribables/mappingHelpers.js
@@ -10,7 +10,7 @@
         return mapJsObjectGraph(rootObject, function(valueToMap) {
             // Loop because an observable's value might in turn be another observable wrapper
             for (var i = 0; ko.isObservable(valueToMap) && (i < maxNestedObservableDepth); i++)
-                valueToMap = valueToMap();
+                valueToMap = valueToMap.peek();
             return valueToMap;
         });
     };


chrisknoll:
I depend on ko.toJS to unwrap observables.  The use case: a dirtyFlag.

When you initialize the dirty flag, it will do a ko.toJS() on the object and invoke the observables on all the properties.  Then, if something inside the object changes, the dirty flag will have it's subscription to the object notified, and the dirty flag can change the state.

Also, I have another case where I show the view model state in a debugging text context that when anything observable changes, the text area gets updated.  In this case, it's to.JSON(), but I'm assuming you would want toJS and toJSON to have similar behavior.

I wouldn't modify this behavior, but if you need something that will 'quietly' de-serialize introduce a different method.

brianmhunt:
This would be API-breaking, and the next major revision will likely be based on TKO; so this is not likely to get merged.

You can replicate this behaviour without modifying the API by wrapping the `toJS` in a `ko.dependencyDetection.ignore()` call i.e. `ko.dependencyDetection(() => ko.toJS(thing))`.

Closing this on the above basis, but please do indicate if there's a reason to keep this open.  Happy to discuss.

@mbest feel free to re-open, of course.

PR #2436, opened 2019-01-25
Enable ko.utils.extend to also extend Symbols

bennieswart:


bennieswart:
@mbest would you mind giving some feedback here? Do you think it is a valid use case?

mbest:
I do see the value of copying symbol properties, but I'm concerned that this would be a breaking change.

bennieswart:
@mbest what would be the protocol for merging this if it is a breaking change, albeit an improvement? In any case, I suspect the number of users that could possibly experience a break due to this would be minuscule.

PS Couldn't we get away with some low-probability breaking changes in 3.5?

chrisknoll:
If all the existing tests work, isn't that enough to assert that it doesn't lead to a breaking change?

uncaught:
This should not be the default behavior. All extending/merging/assigning methods of any library I know (underscore, jquery, lodash) or native code ignore the non-enumerable properties. So while I have actually needed to extend something with symbols before, I still think you should hide this feature behind a flag or use a different method name. It is just not the expected behavior of such a method.

bennieswart:
I think @uncaught raises a valid point. The default behaviour for `ko.utils.extend` should perhaps be to exclude non-enumerable properties.
Still though, I think it's useful to have the option to include symbols when the binding context is extended. Perhaps we should hide this functionality in `ko.utils.extend` behind a flag and set that flag from the call to extend the binding context.

the-djmaze:
I know, old habbits are hard to die in code when the standards use ither namings like Object.assign().
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

Object.getOwnPropertyNames doesn't list Symbol
Object.getOwnPropertySymbols does that

So what about 'extendWithSymbols'?

Issue #2305, opened 2017-10-07
descendant bindings of 'foreach' are evaluated when corresponding item in array is removed


fastfasterfastest:
Documentation states: "When you delete array entries, foreach will simply remove the corresponding DOM elements."  However, it appears that in some cases the bindings in the corresponding DOM elements are re-evaluated _before_ the DOM elements are removed.  And when the bindings are re-evaluated, they are evaluated with the old/previous values of `$index` (and `$data`) even though the array in the parent `foreach` has been updated with new values.

The net effect is that the bindings can be evaluated with an out-of-bound `$index` value - a value that is greater than the length of the array.

Compare this to how the `if` binding behaves (after #https://github.com/knockout/knockout/pull/2226 and https://github.com/knockout/knockout/pull/2234).
The descendant bindings of `if` now have a dependency on the condition of the `if` binding causing the condition of the `if` bindning to be evaluated before the descendants. If the condition is `false` then the descendants are removed and not re-evaluated.

I think the behavior of the `foreach` binding should be something similar.  If the array is replaced with new (fewer) items then the corresponding descendants should be removed and _not_ re-evaluated before being removed.

Normally that happens but I have a situation where the descendants are re-evaluated before being removed, causing an error.

In my case, I have a view model with an observableArray `phrases` (used in the `foreach`).  The view model
also maintains another array, `translatedPhrases` - a one-to-one mapping of `phrases` using computed observable.
A `foreach` binding is used to render the `phrases`.  While rendering each phrase, the corresponding translated phrase is also rendered by accessing it from `translatedPhrases` using `$index`.

```
function ViewModel() {
  this.phrases = ko.observableArray();
  this.translatedPhrases = ko.pureComputed(function(){
    return this.phrases().map(function(phrase){ return { phrase: phrase.toUpperCase() }; });
  }, this);
}

...

<!-- ko if: translatedPhrases --><!-- /ko -->

<!-- ko foreach: phrases -->
    <span data-bind="text: $data"></span>
    <span data-bind="text: $parent.translatedPhrases()[$index()].phrase"></span>
<!-- /ko -->
```

The problem happens when `phrases` is replaced with another array that is shorter than the previous one.
Then, under some circumstances, the descendants of `foreach` gets re-evaluated _before_ they are removed and they are evaluated with values of $index (and $data) of the previous array.

[...]

Issue #2439, opened 2019-01-30
Regression in https://github.com/knockout/knockout/pull/2324


bennieswart:
I've noticed a regression in https://github.com/knockout/knockout/pull/2324 where `foreach` (or `template: { foreach }`) creates two UI elements when there is only one element present in the array, duplicating the element.

I can confirm that this is the case, since inspecting the context of the binding at the time of malfunction shows only one element in the array, yet two elements are being displayed on screen.

I can also confirm that this is indeed the source of the regression, since building the branch prior to the merge works as expected.

Unfortunately I'm having a hard time reproducing this in a fiddle as I cannot seem to recreate the exact circumstances needed for the malfunction. In our code, however, the malfunction occurs with perfect reproducability.

At risk of biasing your diagnosis, I'll add that it seems like this happens when `push`ing into an empty `ko.observableArray` and then replacing the array with a new one. I'll continue investigating, seeing if I can create a simple example, but if the if the issue is apparent to anyone (especially @mbest, since he authored), please don't hesitate to take a shot at what it could be.

brianmhunt:
@bennieswart if you want to try it, my [knockout-fast-foreach](https://github.com/brianmhunt/knockout-fast-foreach) plugin might yield different results that shed light on the issue.

bennieswart:
Ok, I've narrowed it down to a simple example. Looks like this is a legitimate bug.

I've posted a minimal example at https://jsfiddle.net/53bvupxs/1/. You'll see the list contains only one element, but two are visible onscreen.

Unfortunately my expertise of knockout internals falls somewhat short here. Tracking down the exact cause will take me some time. I'm hoping one of the devs can identify this a bit quicker.

chrisknoll:
So, I know nothing about what's going on inside these specific bindings.  I agree there is some issue, but printing a duplicate is not it.  Here's my observation on your test case, however:

This is the template:

```
  <!-- ko if: items().length -->
    <div data-bind="foreach: items">
      <div data-bind="text: $data"></div>
    </div>
  <!-- /ko -->
  <div data-bind="foreach: items">
  </div>
```

[...]

PR #2441, opened 2019-02-01
Fix problem with arrayChange where it might report stale data.

mbest:
See #2439.


fastfasterfastest:
FYI, after an initial check, this appears to have fixed https://github.com/knockout/knockout/issues/2305 as well.

Issue #2297, opened 2017-09-23
ko.contextFor no longer reporting context for dynamically bound elements


bennieswart:
In the v3.4.2 release, `ko.contextFor` correctly reports the context for elements bound with `data-bind` tag attributes as well as elements bound dynamically using `ko.applyBindingsToNode`.

After ccec0fc4ac722170ccc16d2553e49967c5f10bd6, this is no longer the case, and `ko.contextFor` no longer reports the context for elements bound using `ko.applyBindingsToNode`.
See https://jsfiddle.net/cfLyu6je/2/ for an example using ccec0fc4ac722170ccc16d2553e49967c5f10bd6~1, and https://jsfiddle.net/ocm2akpr/3/ for an example using ccec0fc4ac722170ccc16d2553e49967c5f10bd6 (before and after the change).

Is this considered correct behaviour or a bug?

mbest:
I don't have a strong opinion either way. The change was made so that `ko.contextFor` could definitively tell you if Knockout has applied bindings to that element. `ko.applyBindings` can't be called multiple times for an element, whereas `ko.applyBindingsToNode` can. (Also, the the latter can be called before or after the former.)

bennieswart:
If it's not a definite bug, then I see no reason to change anything unless more people experience the issue (we will probably see after the next version is released). Some small changes were required to work around this, but nothing major. For someone else it may be more serious, or maybe not at all. I suppose we can decide later once more information is available.

PR #2447, opened 2019-02-22
Added ko.utils.closestContext

chrisknoll:


fastfasterfastest:
It would return a context for a node that possibly does not have a context.  Thus, I think the name proposed is a misnomer.  If this was to be included, perhaps `findClosestContextFor` or `findNearestContextFor` would be more accurate.

chrisknoll:
Well, the purpose of the method is to find a context, so whether or not the node itself has a context or not, the idea is to find a context for a given starting point in the DOM.  The `'findClosestContextFor` is a bit wordy, and also adds so much description that you might wonder 'is there a furthest one?  a neighboring one? a sibling one? why is there only a closest?' I feel that if there's only one purpose of 'finding' a context is to find the closest one, putting 'closest' in the method is redundant.

This is just an auxiliary function to the `contextFor` which explicitly gets the context for the element (or returns undefined if there is none). So, knowing that `contextFor` is used to get the actual context, and `findContextFor` is used to search for it, I don't really feel like we need to make the name any more descriptive.

fastfasterfastest:
> the purpose of the method is to find a context

Then perhaps `ko.utils.findContext`.

The _for_ in `.findContextFor`, to me, implies the context _for_ that node.  But the node may not have a context, and if a node does not have a context then it is perhaps strange that the method returns a context for that node.  But whatever...

chrisknoll:
Ok, you make a good point. i'll make that change.

San4es:
Let me give you my opinion...

The `ko.contextFor(node)` method says - get a **context for** the current **node** passed as the 1st argument (the "get" verb is omitted). Works with current DOM level.

What about `ko.closestContextFor(node)` - find **closest context for** the current **node** passed as the 1st argument (the "find" verb is ommitted). Starts with current node and traverses up through its ancestors in the DOM tree to find 1st node with model (like [$.closest](https://api.jquery.com/closest/)).

Thus we get 2 symmetrical methods.

chrisknoll:
Ok, seems like there's 2 votes for adding some 'closest' context to the call. I'll change to reflect that.

fastfasterfastest:
If it was to be added (I think my vote would actually be not to include it), I think `findContext` is fine (and `closestContext` is ok too).  I do think `for` should not be part of the name (to avoid suggesting that the returned binding context was the context for that node - it may or may not be that.)

(Depending on why this is needed in the knockout API, perhaps a `ko.utils.findClosestBoundNode` would be more useful?  In the future there may be some knockout API, and perhaps there already is one (well, there is one - `ko.contextFor` :) ), that takes a DOM node that was bound by knockout where that could be handy.)

chrisknoll:
@fastfasterfastest , i'm not quite understand the resistance to this.  The point of the name is that it's similar to conextFor() because it's doing the same thing. Given a node, return a context. One (the core ko function) gets the direct context for the node, the other tries to find it.

The purpose of adding it to ko is that everyone (and I mean EVERYONE) is going to have to write their own implementation of finding the context of a node that may be generated int he context of a custom binding, so why _wouldn't_ you want this in the core? it's literally 5 lines of code.

fastfasterfastest:
I don't know, I haven't had the need for it (yet) myself, so maybe I am not convinced it belongs, maybe it belongs in your app if/when needed.  Maybe there are other, perhaps better or more appropriate, ways to get to the data where `ko.contextFor` is currently used.  But maybe not, and as you say, it's very small so no big deal...

However, IMO, they are not "doing the same thing".  The new method does not find the context for a node - it returns a binding context for _some_ node, but that context is not (necessarily) the context for the node passed in.

chrisknoll:
Maybe I was just confused between:

> (Depending on why this is needed in the knockout API, perhaps a `ko.utils.findClosestBoundNode` would be more useful? In the future there may be some knockout API, and perhaps there already is one (well, there is one - `ko.contextFor` :) ),

And:

> However, IMO, they are not "doing the same thing". The new method does not find the context for a node - it returns a binding context for _some_ node, but that context is not (necessarily) the context for the node passed in.

Anyways, this is getting circular.  Considering the 3.4 release behaved one way (and people developed a dependency on it) the 3.5 way should have a function that gives them the 3.4 behavior.  This PR is for that.  As much as `contextFor` was not behaving as the specs worked, the fact that it gave you 'some' context relative to a node...people found useful. So there should be no harm giving people the prior functionality and not telling them 'sorry guys, just implement it yourself', and hopefully the powers that be recognize that this is just keeping prior functionality that exited since KO 2.0 in the core, so that should be some incentive to keep it.

San4es:
>I do think for should not be part of the name (to avoid suggesting that the returned binding context was the context for that node - it may or may not be that.)
>they are not "doing the same thing"

ko.context**For** of ko v3.5.0 =  ko.context**For** of ko v2.0.0 without traverses up through its ancestors
ko.closestContext**For** of ko v3.5.0 = ko.context**For** of ko v2.0.0

So, I do not see any issues with the "For" suffix.

San4es:
Also, I think it will be helpful to know the author's opinion:
https://github.com/knockout/knockout/commit/5e29175e8cd8d3c870560faf48c79d3a6f234b2d#diff-49b895d65fb975dc8b84e3311720a326R145

fastfasterfastest:
If this is functionality deemed needed in the knockout API, I think it is a better choice to revert `ko.contextFor` to its old behavior (re-open https://github.com/knockout/knockout/issues/2297).

(Or alternatively, provide a `ko.utils.findClosestBoundNode` which will give you the tool and may be useful for other purposes.)

chrisknoll:
Ok, hopefully the latest push makes everyone happy.  

@mbest: I saw that 3.5.0 was tagged, but if it hasn't officially released, I'd suggest incorporating this change, otherwise the result will be a change in 3.5.0 that has been the behavior since 2.0, but with no suitable replacement for that functionality.  If you'd like to do further tweaks on this, please feel free to merge this branch into a local branch for knockout that you can tweak, otherwise I do not foresee me making any further changes here.

chrisknoll:
Closing due to lack of activity.

jeanie77:
Sorry, I don't see this function even in version 3.5.1: has it been renamed  or dismissed at all? Thanks!

chrisknoll:
It never made it, this PR was never merged.  But I think they wanted the function name to be `closestContextFor` but the current PR has it as `closestContext`. I can update the PR with that name if it will be accepted, but otherwise I've moved on.

PR #2475, opened 2019-04-12
Properly use optional parameters in declaration file test

weswigham:
See discussion on https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34699 for details.

### DIFF
--- a/spec/types/module/test-module.ts
+++ b/spec/types/module/test-module.ts
@@ -688,7 +688,7 @@ function test_Components() {
         ko.components.register("name", { template: "string-template", viewModel: { instance: null } });
 
         // viewModel from createViewModel factory method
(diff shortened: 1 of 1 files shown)


DanielRosenwasser:
Should the declaration be the one that's changed? https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#optional-parameters-in-callbacks

weswigham:
I dunno. I operated on the assumption that the types were correct and that the callback could legitimately be called we fewer than all its arguments.

mbest:
The problem is actually in DefinitelyTyped's definition for `createViewModel`. Knockout always passes both parameters to the function.

PR #2488, opened 2019-06-25
enhancement(types): Add generic to extender and extenderoptions

caseyWebb:
Adds generic to `Extenders` and `ExtendersOptions` to pass through boxed observable type. Useful for extenders that accept functions that pass through the observable value.

Consider the following use case:

```typescript
import * as ko from 'knockout'

declare module 'knockout' {
  interface ExtendersOptions<T> {
    validate(v: T): boolean
  }
}

function strValidator(v: string) {
  return true
}

const strObs = ko.observable<string>().extend({
  // types match, all good
  validate: strValidator
})

const numObs = ko.observable<number>().extend({
  // type mismatch, caught by typescript
  validate: strValidator
})

const boolObs = ko.observable<boolean>().extend({
  // type of v is inferred as boolean
  validate: (v) => true
})
```


mbest:
This looks nice. I'll do some additional testing and get it merged.

mbest:
Merged with a6f4b07

Issue #2503, opened 2019-11-14
Use of Reserved Word Breaking IE <= 8 Compatibility


vrad-joshuamckelvey:
**Description**
In the Knockout 3.5.1 release, there was a change to the components/defaultLoader.js that is breaking compatibility issues with IE <= 8 due to the use of reserved word default.

`if (module && typeof module === 'object' && module.__esModule && module.default) {
                        module = module.default;
}`

My suggested fix is to change from dot notation to bracket notation.
`if (module && typeof module === 'object' && module.__esModule && module['default']) {
                        module = module['default'];
}`

I have tested this with IE 8 and it seems to fix the issue.

**Expected Behavior**
Compatibility with IE <= 8

**Actual Behavior**
Unable to load page in IE <= 8

**Steps to Reproduce the Problem**
Open page using templates and knockout 3.5.1 in IE <=8 (or 11 in compatibility mode)

**Specifications**
Version: Knockout 3.5.1
Last Working Version: Knockout 3.5.0
Platform: Windows 10
Browser: IE <= 8

**Additional Info**
[Comment on Commit](https://github.com/knockout/knockout/commit/d7b84f9a5b8169dbed7d6a2284e68cf59d83706e#r35958920)

**Example**
[IE8knockout3.5.1bug.zip](https://github.com/knockout/knockout/files/3848459/IE8knockout3.5.1bug.zip)

mbest:
Thanks for reporting this.

karimayachi:
Ah, this was my PR #2483 .. I had not noticed that it was actually merged in 3.5.1. (Neither had I noticed your comment or this issue until now).

My bad, I had expected the fourth condition and the code block not to be parsed if the first three conditions fail (if they don't, we're in a Typescript transpiled, AMD loader situation, which is unlikely to target IE8). But of course, even though it isn't _evaluated_, it is still parsed! Silly me.

But, to be fair, I just didn't stop to think that `default` is reserved and breaks on IE8.

Sorry for the inconvenience and thanks for reporting and fixing!

PR #2504, opened 2019-11-14
Use of Reserved Word Breaking IE <= 8 Compatibility

vrad-joshuamckelvey:
In the Knockout 3.5.1 release, there was a change to the components/defaultLoader.js that is breaking compatibility issues with IE <= 8 due to the use of reserved word default.

if (module && typeof module === 'object' && module.__esModule && module.default) { module = module.default; }

My suggested fix is to change from dot notation to bracket notation.
if (module && typeof module === 'object' && module.__esModule && module['default']) { module = module['default']; }

I have tested this with IE 8 and it seems to fix the issue.

### DIFF
--- a/src/components/defaultLoader.js
+++ b/src/components/defaultLoader.js
@@ -196,8 +196,8 @@
             // The config is the value of an AMD module
             if (amdRequire || window['require']) {
                 (amdRequire || window['require'])([config['require']], function (module) {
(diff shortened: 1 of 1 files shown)


vrad-joshuamckelvey:
[Related Issue](https://github.com/knockout/knockout/issues/2503)

PR #2507, opened 2019-12-16
fix(typings): utils.toJS return type when observable array, regexp, or date

caseyWebb:
Improves types added in #2494 and added test cases.

- Fixes types on deep/recursive unwrapping (observable array of observables)
- Passes through RegExp and Date w/o mapping prototype ([see here](https://stackoverflow.com/questions/59364768/exclude-builtins-when-in-mapped-conditional-type))

In the added test cases, `observableArr` and `builtins` are the two newly covered conditions (the others were good already, but always nice to have tests).

Also, in finding which built-ins get unwrapped correctly (Date and RegExp), I found [this](https://github.com/knockout/knockout/blob/9893233413e467a40919237e524b545e335c1050/src/subscribables/mappingHelpers.js#L27). Should `Map`, `Set`, `WeakMap`, and `WeakSet` (and potentially [others](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects)) be added here? They currently do not behave as expected with `ko.toJS`. Theoretically this is a breaking change since the output for a given input will change, but as-is the returned map/set is just an empty object so I can't see any way it would be being used currently.

### DIFF
--- a/build/types/knockout.d.ts
+++ b/build/types/knockout.d.ts
@@ -329,8 +329,19 @@ export const extenders: Extenders<any>;
 
 //#region subscribables/mappingHelpers.js
 
-type Unwrapped<T> = T extends ko.Subscribable<infer R> ? R :
-    T extends Record<any, any> ? { [P in keyof T]: Unwrapped<T[P]> } : T;
+type Unwrapped<T> = T extends ko.ObservableArray<infer R>
+    ? Unwrapped<R>[]
+    : T extends ko.Subscribable<infer R>
+    ? (
+        R extends Record<any, any>
+            ? { [P in keyof R]: Unwrapped<R[P]>}
+            : R
+    )
+    : T extends Date | RegExp | Function
+    ? T
+    : T extends Record<any, any>
+    ? { [P in keyof T]: Unwrapped<T[P]> } 
+    : T
 
 export function toJS<T>(rootObject: T): Unwrapped<T>;
 export function toJSON(rootObject: any, replacer?: Function, space?: number): string;
--- a/build/types/knockout.d.ts
+++ b/build/types/knockout.d.ts
@@ -329,7 +329,9 @@ export const extenders: Extenders<any>;
 
 //#region subscribables/mappingHelpers.js
 
-export type Unwrapped<T> = T extends ko.Subscribable<infer R>
+export type Unwrapped<T> = T extends ko.ObservableArray<infer R>
+    ? Unwrapped<R>[]
+    : T extends ko.Subscribable<infer R>
     ? (
         R extends ko.Subscribable
         ? unknown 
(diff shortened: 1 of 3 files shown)


build/types/knockout.d.ts
```diff
@@ -329,8 +329,19 @@ export const extenders: Extenders<any>;
 
 //#region subscribables/mappingHelpers.js
 
-type Unwrapped<T> = T extends ko.Subscribable<infer R> ? R :
-    T extends Record<any, any> ? { [P in keyof T]: Unwrapped<T[P]> } : T;
+type Unwrapped<T> = T extends ko.ObservableArray<infer R>
```
mbest: Rather than treating ObservableArray differently, this should be the pattern for any observable that contains an array.

build/types/knockout.d.ts
```diff
@@ -329,8 +329,19 @@ export const extenders: Extenders<any>;
...
+    ? Unwrapped<R>[]
+    : T extends ko.Subscribable<infer R>
+    ? (
```
mbest: `ko.toJS` actually recursively unwraps observables, so this could be `Unwrapped<R>`.

build/types/knockout.d.ts
caseyWebb: this first condition can actually be removed. `Record<any, any>` also matches arrays (they match the interface `Record<number, any>`).

build/types/knockout.d.ts
caseyWebb: Unfortunately typescript only allows recursive types to be used with array or prop mapping. This type is technically correct, but doesn't compile:

```typescript
export type Unwrapped<T> = T extends ko.Subscribable<infer R>
    ? (
        R extends Record<any, any>
            ? { [P in keyof R]: Unwrapped<R[P]>}
            : Unwrapped<R>
    )
    : T extends Date | RegExp | Function
    ? T
    : T extends Record<any, any>
    ? { [P in keyof T]: Unwrapped<T[P]> } 
    : T
```

Throws `Type alias 'Unwrapped' circularly references itself`

---

The best I'm able to come up with is adding another guard and bailing and assigning an explicit `unknown` type when recursive observables are encountered:

```typescript
export type Unwrapped<T> = T extends ko.Subscribable<infer R>
    ? (
        R extends Record<any, any>
            ? { [P in keyof R]: Unwrapped<R[P]>}
            : R extends ko.Subscribable
                ? unknown
                : R
    )
    : T extends Date | RegExp | Function
    ? T
    : T extends Record<any, any>
    ? { [P in keyof T]: Unwrapped<T[P]> } 
    : T
```

The difference between `any` and `unknown` is that consumers would not get silent failures, they would instead be forced to add a type notation before using the value.

build/types/knockout.d.ts
mbest: Looks like that's something new in 3.7.

PR #2513, opened 2020-02-03
Don't require globally installed grunt or typescript

caseyWebb:


package.json
```diff
@@ -8,7 +8,8 @@
   "main": "build/output/knockout-latest.js",
   "types": "build/types/knockout.d.ts",
   "scripts": {
-    "prepublish": "grunt",
+    "grunt": "grunt",
+    "prepublishOnly": "grunt",
```
mbest: This change seems to have broken the CI build.

PR #2528, opened 2020-05-19
BugFix - Should not fire subscribers if date still the same

vinipickrodt:


mbest:
If you want this type of equality, you can set your own function for equality:

`ko.observable.fn.equalityComparer = ...`

Similarly for `ko.computed`.

vinipickrodt:
> If you want this type of equality, you can set your own function for equality:
> 
> `ko.observable.fn.equalityComparer = ...`
> 
> Similarly for `ko.computed`.

Okay, but the default equalityComparer sees same Date objects as being different. Is this behavior right?

mbest:
The comparison is mainly meant for optimization. And you can customize it as you like.

PR #2561, opened 2021-01-31
Fixes #2560 - also fixes incorrect afterAddCallback unit test

### LINKED ISSUE #2560: In beforeRemove callback, index is always 0
When using the `beforeRemove` callback with the foreach binding, the supplied index parameter is always 0. Bug seems to have been introduced with KO 3.5.0 Beta; it wasn't present in 3.4.2.

See here for reproduction: https://codepen.io/gxzzin/pen/KKNPKeO

marcovdb:
This fixes #2560. It also unearthed an error in the unit test for afterAddCallback, which I then also fixed.


mbest:
I decided to fix this a different way. See 45e86bf0347cf742ca359bab051ea82fdfae7a5d

PR #2566, opened 2021-04-20
Upate npm packages and fix enable binding

manfredwippel:


webketje:
In what prehistoric, by Knockout supported browser is this assumption true?
> //element.disabled may not work in every Browser

manfredwippel:
> In what prehistoric, by Knockout supported browser is this assumption true?
> 
> > //element.disabled may not work in every Browser
> 
> +1 for the devDependencies upgrade though

We had problems in some Firefox versions and changed it - years ago ;).

[https://stackoverflow.com/questions/1710994/firefox-handling-of-disabled-fields](same problem here)

PR #2616, opened 2025-02-19
Fix in the selectedOptions binding

tomasherceg:
See https://github.com/riganti/dotvvm/pull/1910


tomasherceg:
Please accept my apologies - I did not realize GitHub created the PR to the main repo. I intended to do it create it only in my forked repo.

PR #2618, opened 2025-04-02
DOM text reinterpreted as HTML Update utils.domManipulation.js

Shivam7-1:
By using DOMPurify it will avoid the risk of HTML injection, as these properties automatically escape any HTML special characters in the provided text. This helps prevent cross-site scripting (XSS) vulnerabilities.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,5 +41,6 @@
     "grunt": "~0.4.1",
     "grunt-cli": "~0.1.0",
     "typescript": "^3.7.5"
+    "dompurify": "^3.2.4"
   }
 }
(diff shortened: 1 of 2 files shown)


Shivam7-1:
Hii @mbest could you please Review this PR 
Thanks

martin-badin (approved):

phillipc:
That would violate Knockout's zero-dependency principle. Perhaps you could switch to TKO. We have a hook there to use custom implementations like DOMPurify: `options.sanitizeHtmlTemplate`

PR #2629, opened 2026-07-04
fix exponential backtracking in binding expression tokenizer

Arawoof06:
The string and regex-literal tokens in bindingToken share the shape (?:\\.|[^"])*, where the negated class also matches the backslash that the escape branch consumes. On an unterminated string with a run of backslashes the engine can tile those backslashes across the two branches in exponentially many ways before it gives up, so parseObjectLiteral (which runs on every data-bind attribute during applyBindings) stalls: around 40 backslashes takes ~0.8s and ~50 freezes the thread for minutes. Excluding the backslash from each negated class ([^"\\], [^'\\], [^`\\], [^/\\]) makes the escape branch unambiguous so matching stays linear, and every valid literal still tokenizes the same.

### DIFF
--- a/spec/expressionRewritingBehaviors.js
+++ b/spec/expressionRewritingBehaviors.js
@@ -225,4 +225,13 @@ describe('Expression Rewriting', function() {
                 { key: 'd', value: "'empty comment'" }
             ]);
     });
+
+    it('Should not backtrack exponentially on an unterminated string with escape characters', function() {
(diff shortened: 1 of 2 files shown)


Arawoof06:
any update?

mbest:
This is a good change. I'll mark it for the next version.

Arawoof06:
Sounds good, thanks for taking it.

